PR #33, opened 2021-02-04
[RFC 0001] Connection Pooling 

kiwicopple:
We would like to explore connection pooling on Supabase. This RFC is intended to decide:

- Whether we should provide a pooler
- Which connection pooler we should use
- Where in the stack it would be installed - i.e. if should bundle it with the Postgres build

[Rendered](https://github.com/supabase/postgres/blob/rfc/connection_pooling/rfcs/0001-connection-pooling.md)

### DIFF
--- a/rfcs/0001-connection-pooling.md
+++ b/rfcs/0001-connection-pooling.md
@@ -1,0 +1,71 @@
+---
+feature: Connection Pooling
+start-date: 2021-02-04
+author: kiwicopple
+co-authors: steve-chavez, dragarcia
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+We would like to explore connection pooling on Supabase. This RFC is intended to decide:
+
+- Whether we should provide a pooler
+- Which connection pooler we should use
+- Where in the stack it would be installed - i.e. if should bundle it with the Postgres build
+
+
+# Motivation
+[motivation]: #motivation
+
+In Postgres, every connection is a process. Because of this, a lot of connections to the database can be very expensive on memory. 
+
+When connecting to Postgres database from serverless functions, there is no connection pooling, and so the server needs to maintain hundreds/thousands of connections.
+
+
+# Detailed design
+[design]: #detailed-design
+
+This is still in the "Gather Feedback" stage. To start the discussion:
+
+
+### 1. Decide on a PG Pooler
+
+- `pg_bouncer` - https://www.pgbouncer.org/
+- `PG Pool II` - https://www.pgpool.net/mediawiki/index.php/Main_Page
+- `odyssey` - https://github.com/yandex/odyssey
+- others?
+
+### 2. Decide on configuration
+
+Most poolers allow different configurations. We would need to decide on how we would configure the pooler by default
+
+### 3. Decide if the user should be able re-configure the pooler
+
+Should a user be able to change the configuration? If so, how would they do it? 
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Security
+- Not directly relevant to the "supabase" stack, so it's additional non-core support
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Since we already offer [PostgREST](https://github.com/postgrest/postgrest) and [postgres-meta](https://github.com/supabase/pg-api), this isn't entirely necessary for the Supabase stack. Bundling this is only beneficial for connecting external tools. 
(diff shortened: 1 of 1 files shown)


rfcs/0001-connection-pooling.md
```diff
@@ -0,0 +1,71 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Security
+- Not directly relevant to the "supabase" stack, so it's additional non-core support
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Since we already offer [PostgREST](https://github.com/postgrest/postgrest) and [postgres-meta](https://github.com/supabase/pg-api), this isn't entirely necessary for the Supabase stack. Bundling this is only beneficial for connecting external tools.
```
steve-chavez, line 58: I must say I'm surprised of the demand for this because most languages have pools at the library level already(e.g. PostgREST - haskell: hasql-pool, pg-api - Node: pg.Pool?).

From what I've seen it's mostly a Prisma use-case, but seems they do have a [pool](https://www.prisma.io/docs/concepts/components/prisma-client/connection-management#connection-pool).

steve-chavez: I mean adding a pooler won't bring more parallel capacity if the clients are creating sessions.

However, transaction pooling and statement pooling(see [pgbouncer types of pooling](https://www.pgbouncer.org/features.html)), might help here. But I'm not sure how well the client libs support those. Transaction pooling and statement pooling don't allow using prepared statements for example.

rfcs/0001-connection-pooling.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+This is still in the "Gather Feedback" stage. To start the discussion:
+
+
+### 1. Decide on a PG Pooler
+
+- `pg_bouncer` - https://www.pgbouncer.org/
```
steve-chavez, line 35: I'd vote for pgbouncer. Heroku(1) and DO(2) use it.

(1):https://www.digitalocean.com/community/tutorials/managed-databases-connection-pools-and-postgresql-benchmarking-using-pgbench (search for pgbouncer)
(2):https://devcenter.heroku.com/articles/best-practices-pgbouncer-configuration

rfcs/0001-connection-pooling.md
```diff
@@ -0,0 +1,71 @@
...
+- Which connection pooler we should use
+- Where in the stack it would be installed - i.e. if should bundle it with the Postgres build
+
+
+# Motivation
+[motivation]: #motivation
+
+In Postgres, every connection is a process. Because of this, a lot of connections to the database can be very expensive on memory. 
+
+When connecting to Postgres database from serverless functions, there is no connection pooling, and so the server needs to maintain hundreds/thousands of connections.
```
steve-chavez, line 24: Transaction pooling and statement pooling(see [pgbouncer types of pooling](https://www.pgbouncer.org/features.html)) might help there. The disadvantage of those is that they cannot use prepared statements for example, so we'd need to check for support on common client libs(too much work, maybe just document it).

Seems DO also recommends transaction pooling: https://assets.digitalocean.com/articles/managed_db_pools/connection_pool_blank.png

rfcs/0001-connection-pooling.md
```diff
@@ -0,0 +1,71 @@
...
+### 3. Decide if the user should be able re-configure the pooler
+
+Should a user be able to change the configuration? If so, how would they do it? 
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Security
+- Not directly relevant to the "supabase" stack, so it's additional non-core support
```
steve-chavez, line 53: We could add a dedicated GitHub discussion category for "External libraries" and mention it's low prio for us(plus reasoning).

rfcs/0001-connection-pooling.md
kiwicopple: Agreed on this - when I look at the additional tooling around Postgres, pgbouncer seems to be the one supported most frequently (eg, `pgmetrics`, various postgres `operators`).

rfcs/0001-connection-pooling.md
steve-chavez: There's also an AWS pgbouncer fork: https://github.com/awslabs/pgbouncer-rr-patch.

dragarcia:
Will merge this in for future reference.

PR #41, opened 2021-05-11
Add postgres exporter

darora:
Adds postgres exporter, aws cli, protected behind a `supabase_internal` variable.

They're only built into the image when packer is invoked with the aforementioned variable set to true, e.g.:

```
$ packer build -timestamp-ui \
  -var "ansible_arguments=--skip-tags,update-only,-v,-e,supabase_internal='true'" \
  amazon-arm.json
```


ansible/files/queries.yml
```diff
@@ -0,0 +1,30 @@
...
+
+pg_stat_database:
+  query: "SELECT sum(numbackends) as numbackends FROM pg_stat_database"
+  master: true
+  metrics:
+    - numbackends:
+        usage: "GAUGE"
+        description: "The number of active backends"
+
+pg_stat_statements:
```
darora, line 21: This metric is probably the weirdest one here, in that it might be misleading unless you're paying attention to what the data means. Open to suggestions for better ones!

ansible/files/queries.yml
```diff
@@ -0,0 +1,27 @@
+pg_database:
+  query: "SELECT SUM(pg_database_size(pg_database.datname)) / (1024 * 1024) as size_mb FROM pg_database"
```
kiwicopple, line 2: aha - this is great saves us fetching from their DB each time

ansible/files/queries.yml
kiwicopple: This also requires `pg_stat_statements` to be enabled on the database (we should probably do this by default), so we should check that it doesn't fail drastically when it is disabled

ansible/files/queries.yml
darora: Thanks for the pointer--will bring them in here

kiwicopple: I just realised that this is the generic "postgres" image (which anyone can use), so some of the `auth/storage` schema metrics will be irrelevant..

inian:
We might need to separate out the metrics which are specific to Supabase into our worker instead since these images also are pushed to AWS and DO marketplaces and we should figure out a way to keep those images as generic as possible.

inian (approved):

inian:
LGTM!

@dragarcia FYI the [supabase-external variable](https://github.com/supabase/postgres/pull/41/commits/743274e80ea4fe27a495bf64d9b61756487440af#diff-372784febf11c83d9ecf943bec37e9d020784ad32dc405f033cf7552d2923506R1) needs to be set to false when building the marketplace images

PR #59, opened 2021-07-14
feat: add queries for additional metrics

darora:


ansible/files/queries.yml.j2
```diff
@@ -35,6 +173,17 @@ auth_users:
         usage: "GAUGE"
         description: "Number of users in the project db"
 
+replication:
```
darora, line 176: We'll select the series of interest with e.g. `replication_realtime_lag{slot_name="realtime"}`

w3b6x9: @darora Gotcha, would it be preferable to make that part of the query?

i.e. `SELECT slot_name, pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn) as replication_lag FROM pg_replication_slots WHERE slot_name = 'realtime';`

Also, `pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn)` outputs lag in bytes. Would it be preferable to use `pg_size_pretty(pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn))` since its output is more reader friendly (e.g. 92 kB)?

darora: I can divide it if desired, though it needs to be scalar so we wouldn't want to use `pg_size_pretty`. Generally, the norm in this space is to keep the native units (e.g. bytes), and then allow the frontends to convert them as appropriate.

Happy to select for realtime in the query, was leaving it this way in case we wanted info on other replications (now or in the future).

darora: Clarified by renaming metric to `replication_realtime_lag_bytes` (also more idiomatic)

inian (approved):

ansible/files/queries.yml.j2
w3b6x9: > norm in this space is to keep the native units (e.g. bytes), and then allow the frontends to convert them as appropriate

thanks, good to know!

> leaving it this way in case we wanted info on other replications

yep, sounds good!

PR #61, opened 2021-07-15
Boot time optimizations

darora:
Built on top of PG13 refactor, but doesn't need to bundled in as part of the v0 image - let's get that out the door first.

### DIFF
--- a/ansible/tasks/internal/optimizations.yml
+++ b/ansible/tasks/internal/optimizations.yml
@@ -1,0 +1,15 @@
+- name: ensure services are stopped
+  community.general.snap:
+    name: amazon-ssm-agent
+    state: absent
+
+- name: ensure services are stopped and disabled for first boot
+  systemd:
+    enabled: no
+    name: '{{ item }}'
+    state: stopped
+  with_items:
+    - snapd
+    - postgresql
+    - postgrest
+    - fail2ban
--- a/ansible/tasks/internal/optimizations.yml
+++ b/ansible/tasks/internal/optimizations.yml
@@ -12,4 +12,5 @@
     - snapd
     - postgresql
     - postgrest
+    - pgbouncer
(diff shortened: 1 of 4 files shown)


ansible/tasks/internal/optimizations.yml
```diff
@@ -0,0 +1,15 @@
...
+    state: absent
+
+- name: ensure services are stopped and disabled for first boot
+  systemd:
+    enabled: no
+    name: '{{ item }}'
+    state: stopped
+  with_items:
+    - snapd
+    - postgresql
```
dragarcia, line 13: for `postgresql`, correct me if I'm mistaken but I believe we would need it to have it started since we're running the `.sql` scripts at boot (then restarting it)

ansible/tasks/internal/optimizations.yml
```diff
@@ -0,0 +1,15 @@
...
+    - postgrest
```
dragarcia, line 14: should this be `pgbouncer` instead?

ansible/tasks/internal/optimizations.yml
darora: Ah right, yeah

ansible/tasks/internal/optimizations.yml
darora: added pgbouncer; this list is a superset of the things that might be enabled in the image

darora: (pgbouncer isn't enabled)

ansible/tasks/internal/optimizations.yml
darora: Going to leave it stopped per offline discussion; we can start it up once it's been configured, before the init scripts need to be run.

PR #62, opened 2021-07-16
fix: update log rotation location and frequency

darora:


ansible/files/logrotate-postgres
```diff
@@ -1,10 +1,9 @@
-/var/lib/postgresql/12/main/pg_log/postgresql.csv {
-    daily
```
darora, line 2: the `daily` directive is a no-op when used with `size`, so I'm removing it

inian (approved):

ansible/files/logrotate-postgres
```diff
@@ -1,10 +1,9 @@
...
-    size 50M 
+/var/lib/postgresql/data/pg_log/postgresql.log {
```
darora, line 1: @dragarcia fyi it'll be broken on the pg13 ami w/o this

dragarcia: sure thing, let me rebuild and update the images

PR #66, opened 2021-08-12
fix: man-db is a cronjob and not a systemd service

darora:
Additionally, disable a couple of other cronjobs that are not useful
for us.

### DIFF
--- a/ansible/tasks/internal/optimizations.yml
+++ b/ansible/tasks/internal/optimizations.yml
@@ -13,7 +13,6 @@
     - postgresql
     - pgbouncer
     - fail2ban
-    - man-db
     - motd-news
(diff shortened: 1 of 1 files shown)


ansible/tasks/internal/optimizations.yml
```diff
@@ -30,4 +29,13 @@
...
     - lvm2-monitor
+
+- name: disable man-db
+  become: yes
+  file:
+    state: absent
+    path: "/etc/cron.daily/{{ item }}"
+  with_items:
     - man-db
+    - popularity-contest
```
inian, line 40: hahaha

inian (approved):

PR #74, opened 2021-08-26
chore: install a few utilities

darora:


dragarcia (approved):

ansible/tasks/setup-system.yml
```diff
@@ -37,6 +37,14 @@
     update_cache: yes
     cache_valid_time: 3600
 
+- name: Install other useful tools
+  become: yes
+  apt:
+    pkg:
+      - magic-wormhole
```
inian, line 44: What are you planning to use this for Div?

darora: Transferring files/logs as needed

PR #88, opened 2021-10-07
feat: install pg-net extension for internal builds

darora:
I'm not sure what the status of pg_net is; i.e. if it's ready to be supported in non-external environments at this stage. For now, I'd like to move it here to avoid having to do this at project creation time.


darora:
haven't tried out the build yet though!

darora:
Closing in favour of #94

PR #91, opened 2021-10-18
Enable statement logging

darora:
Additionally, enable logrotation on csv logs to cope with increased
log sizes.

### DIFF
--- a/ansible/files/logrotate-postgres-csv
+++ b/ansible/files/logrotate-postgres-csv
@@ -1,0 +1,11 @@
+/var/lib/postgresql/data/pg_log/postgresql.csv {
+    size 50M
+    rotate 9
+    compress
+    delaycompress
+    notifempty
+    missingok
+    postrotate
+        sudo -u postgres pg_ctl -D /var/lib/postgresql/data logrotate
+    endscript
+}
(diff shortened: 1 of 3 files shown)


ansible/files/logrotate-postgres-csv
```diff
@@ -0,0 +1,11 @@
+/var/lib/postgresql/data/pg_log/postgresql.csv {
```
inian, line 1: We will need to roll this out to existing pg13 images too. I see logrotate is currently rotating `/var/lib/postgresql/data/pg_log/postgresql.log` which has no logs going to it

ansible/files/postgresql_config/postgresql.conf.j2
```diff
@@ -570,7 +570,7 @@ log_line_prefix = '%h %m [%p] %q%u@%d '		# special values:
 #log_parameter_max_length_on_error = 0	# when logging an error, limit logged
 					# bind-parameter values to N bytes;
 					# -1 means print in full, 0 disables
-#log_statement = 'none'			# none, ddl, mod, all
+log_statement = 'all'			# none, ddl, mod, all
```
inian, line 573: should we change log_min_messages too? It is only set to warning now

inian (approved):

ansible/files/logrotate-postgres-csv
darora: Yeah there'll be a few more changes as well

ansible/files/postgresql_config/postgresql.conf.j2
darora: Sure, why not. Setting it to info

PR #105, opened 2021-11-19
Add pgsodium extension to supabase/postgres ansible build scripts

michelp:

It installs libsodium and pgsodium into the supbase/postgres ansible task chain.

This is a new feature and not related to current behavior.

## What is the new behavior?

CREATE EXTENSION pgsodium;

## Additional context

This is just the base extension functions, changes to upstream supabase applications will depend on pgsodium (and libsodium).

Note that as installed here, pgsodium is not loaded at postgres boot and there is no root key.  All pgsodium functions are available but there is no key derivation or key ids in this mode.  Upstream applications can support key derivation by adding `pgsodium` to `shared_preload_libraries` and `pgsodium.getkey_script=...` to their `postgresql.conf`.


kiwicopple:
💯 thanks @michelp - could you also please update the readme with a link to your extension?

dragarcia (approved):
Hi @michelp 👋 

LGTM! Thanks for the PR, will be merging this in.

Issue #104, opened 2021-11-19
ARM Docker image


sixers:
# Feature request
Build & publish ARM64 version of the Docker image.

## Is your feature request related to a problem? Please describe.
AMD64 images have terrible performance on M1 Macs (even M1 Pro/M1 Max)

## Describe the solution you'd like
It would be cool to build & publish linux/arm64 version of the Docker image to Docker Hub, it would help a lot with local development on M1 Macs.

## Describe alternatives you've considered

Running existing linux/amd64 image.

## Additional context

`sysbench` CPU benchmark results for `linux/amd64` version of the base `postgres` image:

```
root@c0af6e852479:/# sysbench cpu --threads=2 run
sysbench 1.0.20 (using system LuaJIT 2.1.0-beta3)

Running the test with following options:
Number of threads: 2
Initializing random number generator from current time


Prime numbers limit: 10000

Initializing worker threads...

Threads started!

CPU speed:
    events per second:   360.03

General statistics:
    total time:                          10.0029s
    total number of events:              3602

Latency (ms):
         min:                                    5.33
         avg:                                    5.55
         max:                                   11.34
         95th percentile:                        5.67
         sum:                                19994.19

Threads fairness:
    events (avg/stddev):           1801.0000/0.00
    execution time (avg/stddev):   9.9971/0.00
```

`sysbench` CPU benchmark results for `linux/arm64` version of the base `postgres` image:

```
root@3cd6496ebf2a:/# sysbench cpu --threads=2 run
sysbench 1.0.20 (using system LuaJIT 2.1.0-beta3)

Running the test with following options:
Number of threads: 2
Initializing random number generator from current time


Prime numbers limit: 10000

Initializing worker threads...

Threads started!

CPU speed:
    events per second: 21270.60

General statistics:
    total time:                          10.0001s
    total number of events:              212719

Latency (ms):
         min:                                    0.09
         avg:                                    0.09
         max:                                    2.68
         95th percentile:                        0.10
         sum:                                19970.88

Threads fairness:
    events (avg/stddev):           106359.5000/5.50
    execution time (avg/stddev):   9.9854/0.00
```

60x faster!

soedirgo:
Resolved in #98

Issue #82, opened 2021-09-22
Use Dockerfile to build the image


vHanda:
Hi. Usually most Docker images are built using a Dockerfile. It would be nice if either -

* The `supabase/postgres` image could also be built via a Dockerfile, thereby simplifying the build process

or 

* The readme could be updated to educate users like me about the reason for choosing ansible over a Dockerfile.

olirice:
Sorry no one got back to you on this

We have recently revamped our docker build and are in the process of documenting all the changes.

In the short term you can build the docker image locally like this:
```
docker build -f Dockerfile-156 -t supa-15-6 .
```
better docs to follow

PR #108, opened 2021-11-25
Enhancements to docker images

dragarcia:
Objectives
- [x] Usage of Dockerfile
    - #82
- [ ] ~~Set up GA for pushing out to Docker Hub~~
    - ~~Utilising Dockerfile implementation above~~
    - Will leave this out for another PR once additional runners have been settled
- [x] Provide option for ARM64 docker image
    - #104  
- [x] General cleanup 
    - Use config files in `/etc/postgersql/` instead of copying them over to `/var/lib/postgresql/data/`
    - Remove init script that is no longer needed
    - Shift majority of cleaning up of the image to the Dockerfile instead of Ansible


.github/workflows/release-docker-arm64.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
+
+      - uses: docker/setup-buildx-action@v1
+
+      - uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
dragarcia: values not set yet

.github/workflows/release-docker-arm64.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          tags: supabase/postgres:latest,supabase/postgres:v${{ version }}
```
dragarcia: this one too

PR #111, opened 2021-12-06
only run CI GA when there is a change in common.vars.json

dragarcia:
A change in `common.vars.json` would signify a bump in the version. Right now, any change in the `develop` branch even from just the README would trigger a build of the AMI. This would cause the [GA to immediately fail](https://github.com/supabase/postgres/actions/runs/1543232332) as the image name already exists.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - develop
+    paths:
+      - common.vars*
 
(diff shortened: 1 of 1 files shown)


darora:
How about instead we start using the commit convention we use in the rest of our repos, and version off that?

Only building on `common.vars.json` has the downside of not building again if e.g. you're fixing a broken build (all too common).

PR #141, opened 2022-02-14
fastboot: Build arm64 AMI image

Lakshmipathi:
- Supports arm64 ext4fs AMI creation using EBS-Surrogate
- Reduce VM boot-time between 10 to 15 seconds
- Reduce VM boot time less than 10 seconds with optional "fast-commit"
- Shrink rootfs size from 16GB to 10GB
- Cleanup unused files and packages

Signed-off-by: Lakshmipathi Ganapathi <lakshmipathi.g@gmail.com>

### DIFF
--- a/amazon-arm64.pkr.hcl
+++ b/amazon-arm64.pkr.hcl
@@ -1,0 +1,218 @@
+variable "ami" {
+  type    = string
+  default = "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-*"
+}
+
+variable "profile" {
+  type    = string
+  default = "${env("AWS_PROFILE")}"
+}
+
+variable "ami_name" {
+  type    = string
+  default = "supabase-postgres"
+}
+
+variable "ami_regions" {
+  type    = list(string)
+  default = ["ap-southeast-2"]
+}
+
+variable "ansible_arguments" {
+  type    = string
+  default = "--skip-tags,install-postgrest,--skip-tags,install-pgbouncer,--skip-tags,install-supabase-internal,ebssurrogate_mode='true'"
+}
+
+variable "aws_access_key" {
+  type    = string
+  default = ""
+}
+
+variable "aws_secret_key" {
+  type    = string
+  default = ""
+}
+
+variable "environment" {
+  type    = string
+  default = "prod"
+}
+
+variable "region" {
+  type    = string
+}
+
+variable "build-vol" {
+  type    = string
+  default = "xvdc"
+}
+
+# ccache docker image details
+variable "docker_user" {
+  type    = string
+  default = ""
+}
+
+variable "docker_passwd" {
+  type    = string
+  default = ""
+}
+
+variable "docker_image" {
+  type    = string
+  default = ""
+}
+
+variable "docker_image_tag" {
+  type    = string
+  default = "latest"
+}
+
+locals {
+  creator = "packer"
+}
+
+variable "postgres-version" {
+  type = string
+  default = ""
+}
+
+# source block
+source "amazon-ebssurrogate" "source" {
+  profile = "${var.profile}"
+  #access_key    = "${var.aws_access_key}"
+  #ami_name = "${var.ami_name}-arm64-${formatdate("YYYY-MM-DD-hhmm", timestamp())}"
+  ami_name = "${var.ami_name}-${var.postgres-version}"
+  ami_virtualization_type = "hvm"
+  ami_architecture = "arm64"
+  ami_regions   = "${var.ami_regions}"
+  instance_type = "t4g.2xlarge"
+  region       = "${var.region}"
+  #secret_key   = "${var.aws_secret_key}"
+
+  # Use latest official ubuntu focal ami owned by Canonical.
+  source_ami_filter {
+    filters = {
+      virtualization-type = "hvm"
+      name = "${var.ami}"
+      root-device-type = "ebs"
+    }
+    owners = [ "099720109477" ]
+    most_recent = true
+   }
+  ena_support = true
+  launch_block_device_mappings {
+    device_name = "/dev/xvdf"
+    delete_on_termination = true
+    volume_size = 10
+    volume_type = "gp3"
+   }
+
+  launch_block_device_mappings {
+    device_name           = "/dev/${var.build-vol}"
+    delete_on_termination = true
+    volume_size           = 16
+    volume_type           = "gp2"
+    omit_from_artifact    = true
+  }
+
+  run_tags = {
+    creator = "packer"
+    appType = "postgres"
+  }
+  run_volume_tags = {
+    creator = "packer"
+    appType = "postgres"
+  }
+  snapshot_tags = {
+    creator = "packer"
+    appType = "postgres"
+  }
+  tags = {
+    creator = "packer"
+    appType = "postgres"
+  }
+
+  communicator = "ssh"
+  ssh_pty = true
+  ssh_username = "ubuntu"
+  ssh_timeout = "5m"
+
+  ami_root_device {
+    source_device_name = "/dev/xvdf"
+    device_name = "/dev/xvda"
+    delete_on_termination = true
+    volume_size = 10
+    volume_type = "gp2"
+  }
+}
+
+# a build block invokes sources and runs provisioning steps on them.
+build {
+  sources = ["source.amazon-ebssurrogate.source"]
+
+  provisioner "file" {
+    source = "ebssurrogate/files/sources-arm64.cfg"
+    destination = "/tmp/sources.list"
+  }
+
+  provisioner "file" {
+    source = "ebssurrogate/files/ebsnvme-id"
+    destination = "/tmp/ebsnvme-id"
+  }
+
+  provisioner "file" {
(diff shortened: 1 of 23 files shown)


amazon-arm64.pkr.hcl
```diff
@@ -0,0 +1,205 @@
+variable "ami" {
+  type    = string
+  default = "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-*"
+}
+
+variable "ami_name" {
+  type    = string
+  default = "supabase-postgres-arm-13.3.0.4"
```
darora: let's remove this default - fail if not provided

amazon-arm64.pkr.hcl
```diff
@@ -0,0 +1,205 @@
...
+}
+
+variable "ami_regions" {
+  type    = list(string)
+  default = ["us-west-2"]
```
darora: either remove, or default to all regions?

amazon-arm64.pkr.hcl
```diff
@@ -0,0 +1,205 @@
...
+}
+
+variable "docker_passwd" {
+  type    = string
+  default = ""
+}
+
+variable "docker_image" {
+  type    = string
+  default = "laks/ccache-arm64"
```
darora: Sounds like this can be any docker image-can we just use an upstream alpine (or ubuntu) image?

ebssurrogate/USAGE.md
```diff
@@ -0,0 +1,71 @@
+## Ext4 amd64 AMI creation
+
+`packer build -var "aws_access_key=$AWS_ACCESS_KEY_ID" -var "aws_secret_key=$AWS_SECRET_ACCESS_KEY" -var "region=$AWS_REGION" \
```
darora, line 3: Is there a reason we would want to retain all these different combinations? I'd be in favour of removing configurability here, beyond arm vs x86.

ebssurrogate/USAGE.md
```diff
@@ -0,0 +1,71 @@
...
+	```
+	
+	Now set DOCKER_IMAGE="<username>/ccache" or DOCKER_IMAGE="<username>/ccache-arm64v8" based on your AMI architecture.
+	
+	
+## EBS-Surrogate File layout
+
+```
+$ tree ebssurrogate/
+ebssurrogate/
```
darora, line 38: I'm assuming using the ebs surrogate builder unlocks some capabilities we're interested in (apologies if you delve into this elsewhere); if so, is there any reason we should retain the existing build machinery around the ebs builder?

ebssurrogate/scripts/chroot-bootstrap.sh
```diff
@@ -0,0 +1,202 @@
...
+
+cat << EOF > /etc/default/grub.d/50-aws-settings.cfg
+GRUB_RECORDFAIL_TIMEOUT=0
+GRUB_TIMEOUT=0
+GRUB_CMDLINE_LINUX_DEFAULT=" root=/dev/nvme0n1p2 rootfstype=ext4 rw noatime,nodiratime,discard console=tty1 console=ttyS0 ip=dhcp tsc=reliable net.ifnames=0 quiet module_blacklist=psmouse,input_leds,autofs4 ipv6.disable=1 nvme_core.io_timeout=4294967295 systemd.hostname=ubuntu"
+GRUB_TERMINAL=console
+GRUB_DISABLE_LINUX_UUID=true
+EOF
+	# remove initd
+	#rm -f /boot/microcode.cpio
```
darora: remove

ebssurrogate/scripts/chroot-bootstrap.sh
```diff
@@ -0,0 +1,202 @@
...
+}
+
+function disable_sshd_passwd_auth {
+	sed -i -E -e 's/^#?\s*PasswordAuthentication\s+(yes|no)\s*$/PasswordAuthentication no/g' \
+	  -e 's/^#?\s*ChallengeResponseAuthentication\s+(yes|no)\s*$/ChallengeResponseAuthentication no/g' \
+	 /etc/ssh/sshd_config
+}
+
+function create_admin_account {
+	groupadd admin
```
darora, line 167: what's this group used for?

ebssurrogate/scripts/surrogate-bootstrap.sh
```diff
@@ -0,0 +1,276 @@
...
+EOF
+	unset FMT
+}
+
+function setup_chroot_environment {
+	# Bootstrap Ubuntu into /mnt
+	debootstrap --arch ${ARCH} --variant=minbase focal /mnt
+
+	# Update ec2-region
+	REGION=$(curl --silent http://169.254.169.254/latest/meta-data/placement/availability-zone | sed -E 's|[a-z]+$||g')
```
darora: `--fail` as well

ansible/playbook.yml
```diff
@@ -63,14 +45,31 @@
...
+        dest: /tmp/{{ item.dest }}
+      loop: "{{ sql_files }}"
+
+    - name: Execute init SQL files
+      become: yes
+      become_user: postgres
+      shell:
+        cmd: /usr/lib/postgresql/bin/psql -f /tmp/{{ item.dest }}
+      loop: "{{ sql_files }}"
+      when: not ebssurrogate_mode
```
darora: Not sure I follow why we're not executing these files?

ansible/tasks/setup-pgbouncer.yml
```diff
@@ -100,6 +100,7 @@
   become_user: postgres
   shell:
     cmd: /usr/lib/postgresql/bin/psql -f /tmp/00-schema.sql
+  when: not ebssurrogate_mode
```
darora: ?

amazon-arm64.pkr.hcl
Lakshmipathi: ok, updated the PR

amazon-arm64.pkr.hcl
Lakshmipathi: removed this too.

amazon-arm64.pkr.hcl
Lakshmipathi: Yes, this can be any image.  one thing to note:  you need to `push`  for this image. Any ubuntu/alpine image from personal namespace should work. `(<dockerhub-username>/ubuntu)`

ebssurrogate/scripts/chroot-bootstrap.sh
Lakshmipathi: Without this `admin` group `tasks/internal/admin-api.yml` fails with error like "Group admin does not exist"

ansible/playbook.yml
Lakshmipathi: Inside `ansible-chroot` I'm unable to start `postgresql`  with systemd. This looks like known ansible issue. So I skipped these init part, thinking "we are not starting postgresql during first boot, may be this can be handled with the script which starts  postgresql ".  

Now found a work-around for this, instead of using `systemd` updated PR uses `pg_ctl` to start this process.

ansible/tasks/setup-pgbouncer.yml
Lakshmipathi: _same as above_

ebssurrogate/USAGE.md
Lakshmipathi: There are two different variants: 1. using `fast-commit` with ext4 and  2. ZFS support for amd64. 
1. Removed this option now, because we are not going to use `fast-commit`  without proper testing for this feature. 
2. Removed this too, ZFS supported only for amd64, I had issues with zfs+arm ami. This can be added later.

ebssurrogate/USAGE.md
Lakshmipathi: Yes, ebs surrogate allows us to created customised  image  from scratch instead of relying on base-image.
Once we test and verified that ebs-surrogate based ami works as expected, we can start removing ebs-builder along with anyother un-used AMI building components.

Lakshmipathi:
Tested this PR with `cli`  and verified disk details can be retrieved and expanded.

ebssurrogate/scripts/surrogate-bootstrap.sh
```diff
@@ -0,0 +1,267 @@
...
+}
+
+function update_systemd_services {
+	# Disable vector service and set timer unit.
+	cp -v /tmp/vector.timer /mnt/etc/systemd/system/vector.timer
+	rm -f /mnt/etc/systemd/system/multi-user.target.wants/vector.service
+	ln -s /etc/systemd/system/vector.timer /mnt/etc/systemd/system/multi-user.target.wants/vector.timer
+
+	# Uncomment below to Disable postgresql service during first boot.
+	# rm -f /mnt/etc/systemd/system/multi-user.target.wants/postgresql.service
```
darora: was this meant to be uncommented?

darora (approved):

ebssurrogate/scripts/surrogate-bootstrap.sh
Lakshmipathi: yes, right. Updated PR and tested the changes.

Issue #143, opened 2022-02-15
Incorrect linux/arm64 image


bhvngt:
# Bug report
## Describe the bug
When I pull `linux/arm64` version of [supabase/postgres](https://hub.docker.com/layers/supabase/postgres/14.1.0/images/sha256-22d7f41a55cb7410d156ae357ce36d8f17d9b62d919e6ab479db9ba2a291c60c?context=explore) and run it as a container, it gives me `x86_64` as platform on executing `uname -a`.

I executed this commands on my `Macbook M1 Pro`.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. docker pull supabase/postgres:latest
2. docker run --rm -it --platform linux/arm64 --name supabase supabase/postgres:14.1.0@sha256:22d7f41a55cb7410d156ae357ce36d8f17d9b62d919e6ab479db9ba2a291c60c /bin/bash
3. uname -a
4. Output is `Linux 2ee3f3bf12c3 5.10.76-linuxkit #1 SMP PREEMPT Mon Nov 8 11:22:26 UTC 2021 x86_64 GNU/Linux`

## Expected behavior

`uname -a` should give `aarch64` as platform architecture

## Screenshots

![image](https://user-images.githubusercontent.com/79074469/153996333-828a8b8b-5377-4d41-8105-60815fe01e5d.png)

## System information

- OS: Macbook M1 Pro OSX 12.2

## Additional context

Rest of the supabase images such as `supabase/gotrue`, `supabase/postgres-meta` etc pulls correct arch for arm based system.

I have also tried pulling the image by supplying the arm64 digest. It still gives me `x86` arch. While `x86` based image works fine with docker on mac, it runs on a much reduced performance with high cpu load.

I was trying to download these images in linux VM running on top of my M1 Pro. There it completely breaks since the VM does not support running `x86` based images.

sarumont:
Looks like #167 fixed this...any chance of getting a new build pushed soonish? :)

sarumont:
For anyone who needs this now, I have built a version of the image, available under `sarumont/supabase:v14`. Note that `pg_graphql` failed to build for me, so I removed it from the image.

soedirgo:
This is fixed as of 14.1.0.66. Thanks for the patience!

johanatan:
Still an issue:

```
jonathan@Mac-Studio scripts % docker pull supabase/postgres:latest
latest: Pulling from supabase/postgres
[...]

PR #147, opened 2022-02-19
fix: postgres bin directory location for pgsodium extension when running a docker build

bhvngt:

Bug Fix

docker build breaks on my system due to incorrect postgres `bin` directory location for `pgsodium` extension.

I faced the broken build while trying to create an arm based image for my M1 based laptop. Not sure if this is connected to #143. Current `arm` based docker image for supabase/postgres gives `x86` architecture.

## What is the new behavior?

build is successful with this change.


bhvngt:
This PR changes the hardcoding of bin directory from `/usr/lib/postgresql/bin` picking it up from `pg_config --bindir`

dragarcia (approved):
Thanks for the bug fix @bhvngt 🐛! Will be merging this in then modifying it slightly so that this change can accommodate the EC2 build as well.

Issue #45, opened 2021-06-02
RFC: new extensions


kiwicopple:
Possible new extensions:

- `is_jsonb_valid`: https://github.com/supabase/supabase/discussions/679#discussioncomment-348469
- `IP4R`: https://github.com/supabase/supabase/discussions/679#discussioncomment-566220
- https://github.com/postgrespro/rum RUM provided a enhanced GIN Index for common search function
- https://github.com/amutu/zhparser zhparser is a PostgreSQL extension for full-text search of Chinese language
- pgRouting https://github.com/supabase/supabase/discussions/679#discussioncomment-688348
- https://github.com/iCyberon/pg_hashids for obscuring ids
- supascript: https://github.com/burggraf/SupaScript

My current thoughts:

- We should do pgRouting now, since we want to be useful for mapping services in the future
- we should add a bunch of extensions around text search + dictionaries - great differentiator
- supascript could be useful

bhvngt:
Do you have any plans to include [pglogical](https://www.2ndquadrant.com/en/resources/pglogical/). From whatever little I know, it looks like it may bring in some more flexibility to supabase realtime in terms of filtering logical replication etc.

dragarcia:
Hi @bhvngt! 👋 

No plans at the moment but it could always be a possibility down the road. For now, could you help me add it to the list [here](https://github.com/supabase/supabase/discussions/679)? We keep track and consolidate all of our extension requests on this discussion page. 

Thanks!

PR #148, opened 2022-02-23
feat: zhparser

fengkx:

Feature, docs update, ...

no this extension
#45
## What is the new behavior?
Add ansible of zhparser

Try to add a new extension

### DIFF
--- a/ansible/tasks/postgres-extensions/19-zhparser.yml
+++ b/ansible/tasks/postgres-extensions/19-zhparser.yml
@@ -1,0 +1,57 @@
+# scws and zhparser
+- name: scws - download scws
+  get_url:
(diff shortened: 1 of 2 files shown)


kiwicopple:
Thanks for the PR @fengkx . We have been discussing PgRoonga internally. Would that work for you instead of zhparser? It seems like it's a better solution because it works with many languages

https://pgroonga.github.io/

fengkx:
I read about PgRoonga docs, It seems it doesn't work like postgrersql builtin tsquery and tsvector. I wonder if it can work with other builtin text search function like `ts_rank_cd`.

kiwicopple:
supabase now supports pgroonga, so I'll close this one

PR #151, opened 2022-03-01
chore: enable debug symbols on PG compile

dragarcia:
first iteration


darora (approved):
Suggest removing the flag enabling cassert ; let's get an rc build out

darora:
On second thoughts, let's leave it in, but only use this build for debugging - won't be releasing this for normal usage

darora:
In which case we can also use `-O0 -g`, since we're not pushing this for wider prod usage

PR #152, opened 2022-03-03
chore: install pg_graphql extension

dragarcia:
Installs the `pg_graphql` extension.

### DIFF
--- a/ansible/tasks/postgres-extensions/19-pg_graphql.yml
+++ b/ansible/tasks/postgres-extensions/19-pg_graphql.yml
@@ -1,4 +1,13 @@
 # pg_graphql
+- name: pg_graphql - download & install dependencies
+  apt:
+    pkg:
+      - clang-11
+      - build-essential
+      - python
+    update_cache: yes
+    install_recommends: no
+
 - name: pg_graphql - download libgraphqlparser
   git:
     repo: https://github.com/graphql/libgraphqlparser.git
--- a/ansible/tasks/postgres-extensions/19-pg_graphql.yml
+++ b/ansible/tasks/postgres-extensions/19-pg_graphql.yml
@@ -37,12 +37,6 @@
     version: "{{ pg_graphql_release }}"
   become: yes
 
-- name: pg_graphql - execute ./bin/pgc build
-  shell:
-    cmd: "./bin/pgc build"
-    chdir: /tmp/pg_graphql
-  become: yes
-
 - name: pg_graphql - build
   make:
     chdir: /tmp/pg_graphql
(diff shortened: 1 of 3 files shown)


ansible/tasks/postgres-extensions/19-pg_graphql.yml
```diff
@@ -0,0 +1,24 @@
+# pg_graphql
+- name: pg_graphql - download libgraphqlparser
+  git:
+    repo: https://github.com/graphql/libgraphqlparser.git
+    dest: /tmp/libgraphqlparser
+    version: master
+  become: yes
```
dragarcia: @olirice do you have a recommendation/ preferred way to install `libgraphqlparser`? Their repository hasn't had any new releases since 2017 so the above is just building off from the master branch.

olirice: > preferred way to install libgraphqlparser

I've been pinning it to the last realease `0.7.0`. Heres the install line in the dockerfile used for the quickstart
https://github.com/supabase/pg_graphql/blob/75565af8ef81f0d4e1cf281356d7270c2c6d5a1d/dockerfiles/db/Dockerfile#L13

ansible/tasks/postgres-extensions/19-pg_graphql.yml
```diff
@@ -0,0 +1,33 @@
+# pg_graphql
+- name: pg_graphql - download & install dependencies
+  apt:
+    pkg:
+      - clang-11
+      - build-essential
+      - python
```
olirice: `python2` is a dependency for installing `libgraphqlparser`. It shouldn't be required by `pg_graphql`. did it cause an issue for you?

dragarcia: Ah, this is more of a catch all for all packages that would be used by the extension and any other dependencies.

ansible/tasks/postgres-extensions/19-pg_graphql.yml
dragarcia: got it 👍

dragarcia:
I've updated the steps to instead emulate the steps taken in the [Dockerfile](https://github.com/supabase/pg_graphql/blob/75565af8ef81f0d4e1cf281356d7270c2c6d5a1d/dockerfiles/db/Dockerfile#L13). One additional step is to run `/sbin/ldconfig -v` to update the links and cache for the shared libraries, in particular, `/usr/local/lib/libgraphqlparser.so`. Beforehand, the following error was returned upon executing `create extension if not exists pg_graphql cascade;`:


>ERROR:  could not load library "/usr/lib/postgresql/lib/pg_graphql.so": libgraphqlparser.so: cannot open shared object file: No such file or directory

dragarcia:
All good to go. @oli will let you approve and merge this one in once the final `pg_graphql` is ready (or if `v0.1.1` is already the one).

olirice (approved):

PR #174, opened 2022-03-28
Direct dockerbuild to self-hosted runner

dragarcia:


.github/workflows/dockerhub-release.yml
```diff
@@ -5,11 +5,14 @@ on:
...
     paths:
-      - docker.vars*
+      - '.github/workflows/dockerhub-release.yml'
+      - 'docker.vars*'
 
 jobs:
   docker_release:
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, linux]
+    timeout-minutes: 360
```
dragarcia, line 14: making this big for now just in case. Will bring it down after the first successful run

darora (approved):

Issue #173, opened 2022-03-27
Docker image seems broken past adminapi driven optimizations


darora:
```                                                                                                   
initdb: warning: enabling "trust" authentication for local connections                                                                                                                                                             
You can change this by editing pg_hba.conf or using the option -A, or                                                                                                                                                              
--auth-local and --auth-host, the next time you run initdb.                                                                                                                                                                        
2022-03-27 20:11:16.778 GMT [43] LOG:  could not open configuration file "/etc/postgresql-custom/generated-optimizations.conf": No such file or directory                                                                          
2022-03-27 20:11:16.778 GMT [43] LOG:  could not open configuration file "/etc/postgresql-custom/custom-overrides.conf": No such file or directory                                                                                 
2022-03-27 20:11:16.778 GMT [43] FATAL:  configuration file "/etc/postgresql/postgresql.conf" contains errors         
```

PR #179, opened 2022-03-29
Revert dockerbuild configurations for now

dragarcia:
Will postpone multi-arch docker images for now and revisit it earliest next week. PR also introduces fixes for #173

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -10,9 +10,7 @@ on:
 
 jobs:
   docker_release:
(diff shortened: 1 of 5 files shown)


soedirgo:
I think we can keep removing `PLATFORM` build arg and just remove `linux/arm64` from `setup-qemu-action` and `build-push-action` - easier to enable it back later

dragarcia:
let me change that now

dragarcia:
alright, updated

soedirgo (approved):

PR #192, opened 2022-04-15
llvm: Set llvm package as manual to prevent autoremoval

Lakshmipathi:
Signed-off-by: Lakshmipathi Ganapathi <lakshmipathi.g@gmail.com>


Lakshmipathi:
Verified that missing file (`libLLVM-11.so.1`) is now available and verified that `ldd llvmjit.so` doesn't report `file not found` error

darora (approved):

PR #210, opened 2022-05-20
Changes for PITR process

dragarcia:


ansible/files/admin_api_scripts/commence_walg_backup.sh
```diff
@@ -2,6 +2,6 @@
 
 set -euo pipefail
 
-WALG_SENTINEL_USER_DATA="{ \"backup_id\": $1, \"project_id\": $2 }" nohup wal-g backup-push /var/lib/postgresql/data --config /etc/wal-g/config.json --verify >> /var/log/wal-g/backup-push.log 2>&1 &
+WALG_SENTINEL_USER_DATA="{ \"project_id\": $1, \"backup_id\": $2 }" nohup wal-g backup-push /var/lib/postgresql/data --config /etc/wal-g/config.json --verify >> /var/log/wal-g/backup-push.log 2>&1 &
```
darora, line 5: How much data do we expect this script to spit out? Do we need to set up logrotate on this file?

ansible/files/admin_api_scripts/commence_walg_restore.sh
```diff
@@ -0,0 +1,39 @@
...
+    recovery_target_time=$2
+
+    echo $recovery_target_time
+    
+    # Stop database and empty it
+    systemctl stop postgresql
+    rm -rf /var/lib/postgresql/data/*
+
+    # Download base backup
+    wal-g backup-fetch /var/lib/postgresql/data $backup_name --config /etc/wal-g/config.json
```
darora: quote $backup_name (consider running shellcheck on these scripts to catch these issues)

Personally I like to enable https://www.shellcheck.net/wiki/SC2250 as well but that's more of a debatable preference.

ansible/files/walg_helper_scripts/wal_change_ownership.sh
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+string=$1
```
darora: can we come up with a better name for this than $string?

ansible/files/walg_helper_scripts/wal_change_ownership.sh
```diff
@@ -0,0 +1,42 @@
...
+
+if [[ $string = "" ]]; then
```
darora: More explicit with `[[ -z "$string" ]]`

ansible/files/walg_helper_scripts/wal_change_ownership.sh
```diff
@@ -0,0 +1,42 @@
...
+	exit 1
+fi
+
+base_dir=$(readlink -f $full_path | cut -d'/' -f2)
+
+# Checks if directory/ file to be manipulated 
+# is indeed within the /tmp directory
+# For example, "/tmp/../var/lib/postgresql/..." 
+# will return "var" as the value for $base_dir
+if [ $base_dir != "tmp" ]; then
```
darora: Let's consistently use bash comparisons `[[ ]]` consistently

ansible/files/walg_helper_scripts/wal_fetch.sh
```diff
@@ -0,0 +1,12 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+# Fetch the WAL file and temporarily store them in /tmp
+sudo -u wal-g wal-g wal-fetch $1 /tmp/$1 --config /etc/wal-g/config.json
```
darora: quoting etc; just verify with shellcheck

darora: Can we nest the file under a more specific dir under /tmp ?

e.g. /tmp/supabase-restores or similar? And then verify the same before ownership reassignment

darora (approved):

ansible/files/admin_api_scripts/commence_walg_backup.sh
dragarcia: It's dependent on the size the database but for the smaller ones, it'll be approx 20 lines. For a large one (~15 GB) it takes up around 50 lines per full backup (incremental backups take approximately 20 lines). Logrotate has already been set up [here](https://github.com/supabase/postgres/blob/develop/ansible/files/logrotate_config/logrotate-walg.conf).

ansible/files/walg_helper_scripts/wal_change_ownership.sh
dragarcia: Have updated it to `$filename` instead.

PR #215, opened 2022-05-31
Make Updates To AMI To Support pg_stat_monitor

J0:

Removes pg_stat_statements and installs pg_stat_monitor.

We use pg_stat_statements.

## What is the new behavior?

We will instead use pg_stat_monitor

### DIFF
--- a/ansible/tasks/postgres-extensions/20-pg_stat_monitor.yml
+++ b/ansible/tasks/postgres-extensions/20-pg_stat_monitor.yml
@@ -6,12 +6,12 @@
     version: "{{ pg_stat_monitor_release }}"
     become: yes
 
-- name: pg_stat_monitor build and install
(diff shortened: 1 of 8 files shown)


ansible/tasks/postgres-extensions/20-pg_stat_monitor.yml
```diff
@@ -0,0 +1,18 @@
+# pg_stat_monitor
+- name: pg_stat_monitor - download and install dependencies
+  git:
+    repo: https://github.com/percona/pg_stat_monitor.git
+    dest: /tmp/pg_stat_monitor
+    version: "{{ pg_stat_monitor_release }}"
+    become: yes
+
+- name: pg_stat_monitor build and install
```
darora: nit: first one is build, second is install

darora (approved):

PR #225, opened 2022-06-08
chore: update custom walg conf

dragarcia:
* Sets param `recovery_target_inclusive = off` which ensures that recovery to a target time does **not** stop after the specified time. Doing so could bring a database back to an undesirable state if the specified time is too close.
* Turning it off would halt the recovery right before the target time.
* Bumps image version to `14.1.0.38`.


ansible/files/postgresql_config/custom_walg.conf.j2
```diff
@@ -14,3 +14,4 @@
 #recovery_target_time = ''
 #recovery_target_action = 'promote'
 #recovery_target_timeline = 'current'
+#recovery_target_inclusive = off
```
darora: newline

darora (approved):

ansible/files/postgresql_config/custom_walg.conf.j2
dragarcia: d'oh

PR #228, opened 2022-06-16
Bump pg_graphql to v0.3.2

olirice:
Increases pg_graphql version from v0.3.1 to v0.3.2


dragarcia (approved):
Could you bump up to `14.1.40` [here](https://github.com/supabase/postgres/blob/develop/common.vars.pkr.hcl#L1) so that the new image will be built?

PR #230, opened 2022-06-23
Bs/misc upgrades

soedirgo:
* bump pgtap to v1.2.0
* supautils: move config to custom .conf
* supautils: add authenticator to reserved_memberships


dragarcia (approved):
Just bump up to `14.1.0.41` [here]( https://github.com/supabase/postgres/blob/develop/common.vars.pkr.hcl) so that a new build can be triggered.

PR #237, opened 2022-07-13
Move installation of python3 outside of Ansible for Docker

dragarcia:
* This should fix the issue of building the Docker image locally


david-plugge:
Is there an issue with merging this PR? I'm building my own image for arm and I had to move the python installation to make it work.

darora:
Docker images are being built off CI now without any issues

PR #238, opened 2022-07-14
feat: add support to allow expanding multiple volumes

pcnc:

- Adds support to allow expanding multiple volumes

### DIFF
--- a/ansible/tasks/internal/node-exporter.yml
+++ b/ansible/tasks/internal/node-exporter.yml
@@ -33,7 +33,7 @@
       - cpu
       - diskstats
       - filesystem
-      - filesystem.ignored-mount-points='^/.+'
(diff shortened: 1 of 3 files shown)


ansible/tasks/internal/node-exporter.yml
```diff
@@ -33,7 +33,7 @@
       - cpu
       - diskstats
       - filesystem
-      - filesystem.ignored-mount-points='^/.+'
+      - filesystem.ignored-mount-points='^/(boot|sys|dev|run).*'
```
pcnc, line 36: Unfortunately have to use this regex since node_exporter uses Golang's regexp library, which doesn't support lookaheads, such as `/(?!data).+`, which would've worked in this case

darora (approved):

PR #243, opened 2022-07-19
chore: use stdout logging for docker container

darora:


ansible/files/postgresql_config/postgresql.conf.j2
```diff
@@ -428,39 +428,7 @@ effective_cache_size = 128MB
...
-					# 0 disables.
-#log_truncate_on_rotation = off		# If on, an existing log file with the
-					# same name as the new log file will be
-					# truncated rather than appended to.
-					# But such truncation only occurs on
-					# time-driven rotation, not on restarts
-					# or size-driven rotation.  Default is
-					# off, meaning append to existing files
-					# in all cases.
+include = '/etc/postgresql/logging.conf'
```
dragarcia, line 431: nit: Can we store it with the rest in `/etc/postgresql-custom` dir?

dragarcia (approved):

ansible/files/postgresql_config/postgresql.conf.j2
darora: The custom dir is currently used for config that can be dynamically updated (whether by user, or internal tooling). Going to leave as-is unless there's a strong objection.

quangtd95:
@darora  may I ask if the latest docker image of supabase/postres applied this fix yet?

darora:
Doesn't look like there's been a release since then; making one now https://github.com/supabase/postgres/pull/265

quangtd95:
@darora  please inform us when the new image is ready to pull. really appreciate this

Issue #235, opened 2022-07-12
XML functions unsupported


boredom101:
# Bug report

## Describe the bug

When I try to use an XML function, it states that the XML feature is unsupported

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Run an SQL snippet like: `SELECT xmlelement(name foo);`
2. See error

## Expected behavior

I expected that the XML function would return expected result.

## Screenshots

![screenshot of error](https://user-images.githubusercontent.com/20235408/178384359-354f505e-730c-44ec-93f8-fdd331dd966f.png)

## Additional context

I think this is caused by postgres not being compiled with libxml.

boredom101:
Hello, I don't want to be a bother, but it seems the only change needed for this bug is to change this line to add `--with-libxml`: https://github.com/supabase/postgres/blob/develop/ansible/tasks/setup-postgres.yml#L71

steve-chavez:
@boredom101 Yes, that should be enough. I suggest to make a PR.

boredom101:
The fix was released, and it works.

PR #249, opened 2022-07-26
configure postgres to include libxml

boredom101:

Bug fix

See issue #235


darora (approved):
Thanks!

PR #250, opened 2022-07-27
feat: add database upgrade scripts

pcnc:

- adds scripts in support of database upgrade oeprations


pcnc:
@supabase/backend ready for review

ansible/files/admin_api_scripts/pg_upgrade_prepare.sh
```diff
@@ -0,0 +1,7 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+systemctl stop postgresql
+umount /data
```
darora, line 13: I guess this upgrade isn't going to work for really old projects, since they didn't use this mount point?

ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
```diff
@@ -0,0 +1,82 @@
...
+    STATEMENT=$1
+    psql -h localhost -U supabase_admin -d postgres -c "$STATEMENT"
+}
+
+cleanup() {
+    UPGRADE_STATUS=${1:-"failed"}
+    EXIT_CODE=${?:-0}
+
+    systemctl start postgresql
+    run_sql "CREATE EXTENSION IF NOT EXISTS pg_graphql CASCADE;"
```
darora: oh that sweet special casing

darora (approved):

ansible/files/admin_api_scripts/pg_upgrade_prepare.sh
pcnc: This runs on the "new" instance - the upgraded one which would replace the database instance. 
Could use more comments throughout these scripts to describe what each one does

ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
pcnc: Hah, yep. It's the only extension that we use that uses regtypes, referencing system OIDs. 

Will break this out into a list of extensions and make the script iterate over that so it's easier to notice and manage. Also comments.

PR #254, opened 2022-08-02
Install pg_jsonschema v0.1.0

olirice:
Installs [pg_jsonschema](https://github.com/supabase/pg_jsonschema) v0.1.0

The ansible task is a best effort but might have some issues. Is there a way to check it in CI without merging to develop?


ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
```diff
@@ -0,0 +1,51 @@
...
+
+- name: download rust/cargo installer
+  when: cargo_exists is failed
+  get_url:
+    url: https://sh.rustup.rs
+    dest: /tmp/sh.rustup.rs
+    mode: '0755'
+    force: 'yes'
+  tags:
+    - rust
```
dragarcia: Would there be a situation wherein we exclude the tag `rust` and not run these steps?

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
```diff
@@ -0,0 +1,51 @@
...
+      - clang
+      - postgresql-server-dev-{{ postgresql_major }}
+      - libssl-dev
+    update_cache: yes
+    install_recommends: no
+
+- name: check if cargo is installed
+  shell: command -v cargo
+  register: cargo_exists
+  ignore_errors: yes
```
dragarcia: It looks like by default `cargo` would not exist beforehand during the build. Could we omit this step and proceed with the rest without the conditional?

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
```diff
@@ -0,0 +1,51 @@
...
+
+- name: install rust/cargo
+  when: cargo_exists is failed
+  shell: /tmp/sh.rustup.rs -y
+  tags:
+    - rust
+
+- name: install cargo-pgx
+  shell:
+    cmd: "cargo install --version '={{ pgx_version }}' cargo-pgx && cargo pgx init --pg{{ postgres_major }} pg_config"
```
dragarcia:
```suggestion
    cmd: "~/.cargo/bin/cargo install --version '={{ pgx_version }}' cargo-pgx && ~/.cargo/bin/cargo pgx init --pg{{ postgres_major }} pg_config"
```

At this point, `cargo` is not yet included in the PATH environment. We'll just run it from the default directory `~/.cargo/bin/cargo`

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
```diff
@@ -0,0 +1,51 @@
...
+- name: pg_jsonschema - download release
+  git:
+    repo: https://github.com/supabase/pg_jsonschema.git
+    dest: /tmp/pg_jsonschema
+    version: "{{ pg_jsonschema_release }}"
+  become: yes
+
+- name: pg_jsonschema - install
+  shell:
+    cmd: "cargo pgx install --release"
```
dragarcia:
```suggestion
    cmd: "~/.cargo/bin/cargo pgx install --release"
```

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
```diff
@@ -0,0 +1,51 @@
...
+  tags:
+    - rust
+
+- name: install rust/cargo
+  when: cargo_exists is failed
+  shell: /tmp/sh.rustup.rs -y
+  tags:
+    - rust
+
+- name: install cargo-pgx
```
dragarcia: The following error was returned upon running this step:
```
Error: 
   0: problem running initdb: "/usr/lib/postgresql/bin/initdb" "-D" "/root/.pgx/data-14"
      initdb: error: cannot be run as root
      Please log in (using, e.g., "su") as the (unprivileged) user that will
      own the server process.
```

Would there be any implications if all the steps here aside from the downloading and installation of dependencies were to be performed by the `postgres` Linux user (which I'm assuming should be owning the server process)?

dragarcia: Have tested running everything as the `postgres` Linux user and it looks to be working. Only hitch is that we would have to temporarily grant the `postgres` Linux user ownership of `/usr/lib/postgresql/lib` to ensure `cargo pgx install --release` is fully installed. Else, the following error would be returned:
```
Error: 
   0: failed copying `/tmp/pg_jsonschema/target/release/libpg_jsonschema.so` to `/usr/lib/postgresql/lib/pg_jsonschema.so`
   1: Permission denied (os error 13)

Location:
   /home/postgres/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.4.5/src/command/install.rs:234

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgx::command::install::install_extension with pg_version=14.14 release=true test=false features=["pg14"]
      at /home/postgres/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.4.5/src/command/install.rs:84
   1: cargo_pgx::command::install::execute
      at /home/postgres/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-pgx-0.4.5/src/command/install.rs:51
```

Should be okay since we're returning ownership to the `root` Linux user afterward. @olirice which OS user do you use for these steps when following the installation in the [pg_jsonschema readme](https://github.com/supabase/pg_jsonschema#installation)?

dragarcia (changes requested):
> The ansible task is a best effort but might have some issues. Is there a way to check it in CI without merging to develop?

What I usually do is spin up the latest image, SSH in, and attempt to install the extension using the steps outlined in the task.

---
Other than that, we'll just need to:
* Remove `ansible/tasks/postgres-extensions/.21-pg_jsonschema.yml.swp` and add `*.swp` in `.gitignore`.
* Rename `21-pg_jsonschema.yml` and anything related to it name-wise to `22-pg_jsonschema.yml` as [21-auto_explain.yml](https://github.com/supabase/postgres/blob/develop/ansible/tasks/postgres-extensions/21-auto_explain.yml) already exists.
* Bump [common.vars.pkr.hcl](https://github.com/supabase/postgres/blob/develop/common.vars.pkr.hcl) to the next version.

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
olirice: the intent was to avoid re-installing after the first run but sure

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
olirice: nope, will remove

ansible/tasks/postgres-extensions/21-pg_jsonschema.yml
olirice: The installation instructions in the readme are more for trying it vs a release so the OS user can be the user who ran `cargo pgx init` when setting up the `pgx` framework

> tested running everything as the postgres Linux user and it looks to be working. Only hitch is that we would have to temporarily grant the postgres Linux user ownership of /usr/lib/postgresql/lib to ensure cargo pgx install --release is fully installed

Could you please please push that change to this branch? I'm not clear how to do it

olirice:
@dragarcia thanks for the feedback! I made the changes you mentioned in all of the comments except the one that is still unresolved

> What I usually do is spin up the latest image, SSH in, and attempt to install the extension using the steps outlined in the task.

do you run the actual ansible script for that, or copy/paste the commands?

dragarcia (approved):

dragarcia:
> do you run the actual ansible script for that, or copy/paste the commands?

The latter. More precisely, I would first come up with the steps in the form of shell commands, run each of these steps to ensure that they're working as expected, then translate them into Ansible tasks.

PR #259, opened 2022-08-06
fix: start postgresql post-resize if not running

darora:


darora:
Closing in favour of #260

PR #279, opened 2022-08-22
Separate --skip-tags parameters from Ansible arguments

dragarcia:
* Noticed that `--skip-tags` parameters [are not observed ](https://github.com/supabase/postgres/runs/7894989613?check_suite_focus=true#step:4:6036) when building an image.
    * By right, with this configuration, certain steps would be omitted in the build process. Instead, all steps are still included. 
* For internal builds, this value is set to `""` as we will want to include all steps.

### DIFF
--- a/amazon-arm64.pkr.hcl
+++ b/amazon-arm64.pkr.hcl
@@ -20,7 +20,12 @@ variable "ami_regions" {
 
 variable "ansible_arguments" {
   type    = string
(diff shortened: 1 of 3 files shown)


pashkinelfe:
I think it's fair enough to be reviewed. @Lakshmipathi

Lakshmipathi (approved):

Lakshmipathi:
Sorry somehow I missed this PR. @dragarcia  @pashkinelfe

darora:

PR #286, opened 2022-09-02
chore: add init scripts to docker build

sweatybridge:

Include sql init scripts in docker image

initial schemas in ami build are missing from docker build

## What is the new behavior?

include all init scripts in docker image

### DIFF
--- a/ansible/playbook-docker.yml
+++ b/ansible/playbook-docker.yml
@@ -5,12 +5,24 @@
   vars_files:
     - ./vars.yml
 
+  vars:
+    sql_files:
+      - { source: "pgbouncer_config/pgbouncer_auth_schema.sql", dest: "00-schema.sql" }
+      - { source: "stat_extension.sql", dest: "01-extension.sql" }
+      - { source: "sodium_extension.sql", dest: "02-sodium-extension.sql" }
+
   tasks:
     - name: Setup container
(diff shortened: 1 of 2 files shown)


soedirgo:
~Let's create a new image for this, possibly based on `supabase/postgres`. `supabase/postgres` is supposed to be "Postgres + goodies", but without auth, realtime, storage, etc. etc.~

Just saw this is only for pgbouncer + pg_stat_statements + pgsodium. I guess that's... fine? I'll leave it to @dragarcia or @darora to decide

ansible/playbook-docker.yml
```diff
@@ -5,13 +5,25 @@
   vars_files:
     - ./vars.yml
 
+  vars:
+    sql_files:
+      - { source: "pgbouncer_config/pgbouncer_auth_schema.sql", dest: "00-schema.sql" }
```
dragarcia, line 10: For now, the docker images don't have PgBouncer attached to it. Would you still want this included to ensure as much parity as possible with the hosted image?

dragarcia (approved):

ansible/playbook-docker.yml
soedirgo: yeah I think that's preferrable for now

PR #289, opened 2022-09-06
chore: update services

dragarcia:
In particular, the following:
* Gotrue
* Vector

### DIFF
--- a/ansible/tasks/setup-supabase-internal.yml
+++ b/ansible/tasks/setup-supabase-internal.yml
@@ -47,18 +47,14 @@
     cmd: |
       usermod -a -G postgres vector
 
-- name: create vector overrides dir
-  file:
-    state: directory
-    owner: root
-    group: root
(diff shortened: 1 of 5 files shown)


ansible/tasks/setup-supabase-internal.yml
```diff
@@ -47,18 +47,14 @@
...
-  file:
-    state: directory
-    owner: root
-    group: root
-    path: /etc/systemd/system/vector.service.d
-    mode: '0700'
-
-- name: Custom systemd overrides - ensure Vector gets restarted
-  copy:
-    src: files/vector.service.conf
```
darora, line 60: remove the corresponding file if it's not being used

darora (approved):

ansible/tasks/setup-supabase-internal.yml
dragarcia: gotcha

PR #295, opened 2022-09-15
test: Add basic postgres database image test using pgTAP

Lakshmipathi:

### DIFF
--- a/ansible/playbook.yml
+++ b/ansible/playbook.yml
@@ -9,7 +9,10 @@
 
   vars:
     sql_files:
-      - { source: "pgbouncer_config/pgbouncer_auth_schema.sql", dest: "00-schema.sql" }
+      - {
+          source: "pgbouncer_config/pgbouncer_auth_schema.sql",
+          dest: "00-schema.sql",
+        }
       - { source: "stat_extension.sql", dest: "01-extension.sql" }
       - { source: "sodium_extension.sql", dest: "02-sodium-extension.sql" }
 
@@ -51,7 +54,7 @@
       tags:
         - install-postgrest
         - install-supabase-internal
-    
+
     - name: Install Kong
       import_tasks: tasks/setup-kong.yml
       tags:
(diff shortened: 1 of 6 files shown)


Lakshmipathi:
tested locally and introduced errors. Build fails with error like:  
err

ebssurrogate/files/unit-tests/unit-test-01.sql
```diff
@@ -0,0 +1,47 @@
...
+-- Check roles
+SELECT has_role('anon');
+SELECT has_role('dashboard_user');
+SELECT has_role('service_role');
+SELECT has_role('supabase_admin');
+SELECT has_role('supabase_auth_admin');
+SELECT has_role('supabase_storage_admin');
+
+-- Check auth schema relations
+SELECT fk_ok( 'auth', 'identities',   'user_id', 'auth',  'users', 'id');
```
darora: what's with the odd spacing between the params? intentional?

ansible/playbook.yml
```diff
@@ -150,6 +153,23 @@
       shell:
         cmd: "for fl in /usr/lib/postgresql/bin/* ; do ln -sf $fl /usr/bin/$(basename $fl) ; done"
 
+    - name: Enable pgTAP extension
```
darora: let's move these out to a separate task file that gets included

ansible/playbook.yml
```diff
@@ -150,6 +153,23 @@
...
+      become: yes
```
darora: `become: yes` should be unnecessary given it's configured for the entire playbook?

darora (approved):

ansible/playbook.yml
Lakshmipathi: okay, will update the PR

ebssurrogate/files/unit-tests/unit-test-01.sql
Lakshmipathi: fixed it.

pcnc (approved):

PR #305, opened 2022-09-27
fix: Ensure fail2ban disabled during first boot

Lakshmipathi:

### DIFF
--- a/ebssurrogate/scripts/surrogate-bootstrap.sh
+++ b/ebssurrogate/scripts/surrogate-bootstrap.sh
@@ -211,6 +211,9 @@ function update_systemd_services {
 
 	# Disable postgresql service during first boot.
 	rm -f /mnt/etc/systemd/system/multi-user.target.wants/postgresql.service
+
+	# Disable auditd
+	rm -f /mnt/etc/systemd/system/multi-user.target.wants/auditd.service
 }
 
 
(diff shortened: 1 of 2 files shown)


Lakshmipathi:
`fail2ban` during first boot adds 5secs to the start-up time.

ebssurrogate/scripts/surrogate-bootstrap.sh
```diff
@@ -211,6 +211,9 @@ function update_systemd_services {
 
 	# Disable postgresql service during first boot.
 	rm -f /mnt/etc/systemd/system/multi-user.target.wants/postgresql.service
+
+	# Disable fail2ban service during first boot.
+	rm -f /mnt/etc/systemd/system/multi-user.target.wants/fail2ban.service
```
darora: systemd disable in `setup-fail2ban`?

darora: hmm is there a reason to handle even e.g. postgres here instead of where we're setting it up initially?

Lakshmipathi: ok, disabled it via ansible task. Ansible task for postgres shows `enabled: yes`

Lakshmipathi:
Added `auditd` to this list.

darora (approved):

PR #310, opened 2022-09-29
chore: build arm docker image

darora:
Tested to build successfully in roughly the same time as the x86 images, using the newly spun up arm64 workers.


darora:
(hold off on this; breaks the multi-arch aspect of things, it looks like)

sweatybridge:
Should be ready to merge now.

sweatybridge (approved):

PR #339, opened 2022-10-23
chore: Run tests from migration

Lakshmipathi:


sweatybridge (approved):

dragarcia (approved):

Lakshmipathi:
Increased plv8 re-tries to 40 from 20. Because while running inside docker - tests taking more time because of less resources than ami builder.

Lakshmipathi:
Even after 40 retries - async within docker fails to find plv8 status. Possible issue with async inside docker https://github.com/ansible/ansible/issues/77046 Will revert retry to 20

@sweatybridge ,  will add a new flag `use_async_mode`. We can disable this flag for docker builds and let docker use linear build, while enabling this flag for ami build to use async mode. Will send this as separate PR

PR #341, opened 2022-10-24
Update pg_graphql to v0.5.0

olirice:
Updates `pg_graphql` from `v0.4.1` to `v0.5.0`

Note, this is a more significantly change than usual version bumps because the tooling to install `pg_graphql` has changed


darora (approved):
Going to approve this for now, but I think we can extract out the rust install as it's now shared across multiple extensions

PR #354, opened 2022-11-01
fix: bump gotrue to v2.23.2

J0:

Bump Gotrue to v2.23


common.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "14.1.0.83-rc1"
+postgres-version = "14.1.0.84"
```
hf, line 1: I'm assuming this should be this way.

hf (approved):

dragarcia (approved):

PR #365, opened 2022-11-07
chore: Add Migration scripts along with extensions test

Lakshmipathi:
After building AMI, Perform extension install test. If tests passed, push migrations files to s3. Also create migration_output.sql file.

### DIFF
--- a/.github/workflows/ami-release.yml
+++ b/.github/workflows/ami-release.yml
@@ -17,3 +17,17 @@ jobs:
       - name: Build AMI
         run: |
           packer build -var-file="development-arm.vars.pkr.hcl" -var-file="common.vars.pkr.hcl" amazon-arm64.pkr.hcl
+
+      - name: Merging migration files
+        run: cat $(ls -1) > ../migration-output.sql
+        working-directory: ${{ github.workspace }}/migrations/db/migrations
+
+      - name: Push migration files to S3
+        uses: jakejarvis/s3-sync-action@master
+        env:
+          AWS_S3_BUCKET: ${{ secrets.PG_INIT_SCRIPT_S3_BUCKET_STAGING }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.PG_INIT_SCRIPT_ACCESS_KEY_ID_STAGING }}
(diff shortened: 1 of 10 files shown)


Lakshmipathi:
Extension test passed. https://github.com/supabase/postgres/actions/runs/3419028280/jobs/5692036378  (ignore the s3 failure, its due to staging key rotation) 

pljava test passed after switching to `supabase_admin` instead of `postgres` though.

.github/workflows/ami-release.yml
```diff
@@ -17,3 +17,17 @@ jobs:
       - name: Build AMI
         run: |
           packer build -var-file="development-arm.vars.pkr.hcl" -var-file="common.vars.pkr.hcl" amazon-arm64.pkr.hcl
+
+      - name: Merging migration files
+        run: cat $(ls -1) > ../migration-output.sql
+        working-directory: ${{ github.workspace }}/migrations/db/migrations
+
+      - name: Push migration files to S3
+        uses: jakejarvis/s3-sync-action@master
```
sweatybridge, line 26:
```suggestion
        uses: jakejarvis/s3-sync-action@master
        with:
          args: --delete
```

Might want to replace existing files

.github/workflows/ami-release.yml
```diff
@@ -17,3 +17,17 @@ jobs:
       - name: Build AMI
         run: |
           packer build -var-file="development-arm.vars.pkr.hcl" -var-file="common.vars.pkr.hcl" amazon-arm64.pkr.hcl
+
+      - name: Merging migration files
+        run: cat $(ls -1) > ../migration-output.sql
+        working-directory: ${{ github.workspace }}/migrations/db/migrations
```
sweatybridge, line 23: Probably not needed anymore if we use migrate.sh

sweatybridge (approved):

darora (approved):

.github/workflows/ami-release.yml
Lakshmipathi: ok, added.

.github/workflows/ami-release.yml
sweatybridge: discussed offline: concatenating migrations into a single file could introduce unexpected behaviour as session configs, ie. `set [session | local] ...`, are persistent for the whole file when executed by psql. Will need to update `restore.ts` in a separate PR to run each migration individually.

PR #397, opened 2022-11-30
Install from .deb for supabase pgx projects

olirice:
Updates the install process for supabase pgx projects to install from .deb packages attached to the GitHub release instead of installing using dev toolchain

### DIFF
--- a/ansible/tasks/postgres-extensions/19-pg_graphql.yml
+++ b/ansible/tasks/postgres-extensions/19-pg_graphql.yml
@@ -1,69 +1,14 @@
 # pg_graphql
-- name: pg_graphql - download & install dependencies
-  apt:
-    pkg:
-      - make
-      - gcc
-      - pkg-config
-      - clang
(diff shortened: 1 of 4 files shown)


pcnc (approved):

ansible/tasks/postgres-extensions/19-pg_graphql.yml
```diff
@@ -1,69 +1,14 @@
...
-    - /usr/lib/postgresql/14/lib
-    - /usr/share/postgresql/14
+- name: Setting arch (x86)
+  set_fact:
+    arch: "amd64"
+  when: platform == "amd64"
+
+- name: Setting arch (arm)
+  set_fact:
+    arch: "arm64"
```
darora: you can just use `{{ platform }}` directly; the separate fact is only needed when e.g. the artifact url uses x86 or aarch instead

darora (approved):

PR #405, opened 2022-12-05
add Trusted Language Extensions

michelp:

Adds Trusted Language Extensions (TLE) Extension


kiwicopple (approved):

michelp:
There's a `pgtle_admin` role that is necessary to install/update/uninstall TLEs that should be conveyed to the `postgres` role at some point in the setup flow, not 100% positive where that should go someone on @supabase/backend have any suggestions?

darora:
Initial instinct would be to suggest `migrations/db/init-scripts/00000000000000-initial-schema.sql` unless this role is categorically different from the ones we currently deal with there?

michelp:
The role is created by the extension at CREATE EXTENSION time, so I don't think anything special needs to be done?  As postgres has CREATEROLE already it should be able to self-grant it, maybe this requires a tweak to supautils?

https://github.com/aws/pg_tle/blob/c12bfc37fb5c1066339aa66c5bad9d04baebe96c/pg_tle--1.0.0.sql#L283

darora (approved):
@sweatybridge fyi doesn't add the extension to the separate docker build

dragarcia:
* Attempt PR for installing TLE on Docker here: https://github.com/supabase/postgres/pull/542

soedirgo:
Re: `pgtle_admin`, we can handle it the same way we handle [`pgsodium_*` roles](https://github.com/supabase/postgres/blob/47c60bf1fb45440d7ab4bcd7405770c39bdffb00/migrations/db/migrations/20221207154255_create_pgsodium_and_vault.sql).

ansible/Dockerfile
```diff
@@ -681,6 +682,27 @@ RUN --mount=type=cache,source=/ccache,target=/ccache,from=public.ecr.aws/t3w2s2c
...
+ARG pg_tle_release
+ARG pg_tle_release_checksum
+ADD --checksum=${pg_tle_release_checksum} \
+    "https://github.com/aws/pg_tle/archive/refs/tags/v${pg_tle_release}.tar.gz" \
+    /tmp/pg_tle.tar.gz
+RUN tar -xvf /tmp/pg_tle.tar.gz -C /tmp && \
+    rm -rf /tmp/pg_tle.tar.gz
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    flex \
+    && rm -rf /var/lib/apt/lists/*
```
dragarcia:
```suggestion
```

PR #411, opened 2022-12-07
Feat/pgsodium vault updates

michelp:

updates to pgsodium and vault.


soedirgo:
Can also confirm this works locally. I think the error is because the test is using an [older build](https://github.com/supabase/postgres/blob/e9584d5b0be3d8a39c84f649b77bf4e7f64a7603/.github/workflows/test.yml#L26) which already has `pgsodium` installed.

I'll modify the PR a bit to make it idempotent ~and to make the tests pass~ (nvm, can't make the test pass because `supabase_vault` isn't on the old image).

sweatybridge (approved):

pcnc (approved):

PR #426, opened 2022-12-12
feat: add `/etc/gotrue.overrides.env` to gotrue systemd unit

hf:
Often Auth needs to modify GoTrue settings that need to persist across restarts / config changes. It's somewhat risky to be changing the systemd unit file on the fly. This makes it easy to add the config to an override file only.

(Note: I'm not sure what happens to systemd units if the [`EnvironmentFile=`](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) does not exist.)

### DIFF
--- a/ansible/files/gotrue.service.j2
+++ b/ansible/files/gotrue.service.j2
@@ -13,6 +13,7 @@ MemoryAccounting=true
 MemoryMax=50%
 
 EnvironmentFile=/etc/gotrue.env
+EnvironmentFile=-/etc/gotrue.overrides.env
 
 Slice=services.slice
 


hf:
Version needs to be bumped, but it's now with `-rc1` so not sure what I should do.

ansible/files/gotrue.service.j2
```diff
@@ -13,6 +13,7 @@ MemoryAccounting=true
 MemoryMax=50%
 
 EnvironmentFile=/etc/gotrue.env
+EnvironmentFile=/etc/gotrue.override.env
```
pcnc: >(Note: I'm not sure what happens to systemd units if the [EnvironmentFile=](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) does not exist.)

Value should be prefixed with a minus sign (`-`) to avoid any errors or warnings related to the file being missing:
> The argument passed should be an absolute filename or wildcard expression, optionally prefixed with "-", which indicates that if the file does not exist, it will not be read and no error or warning message is logged.

---

> Version needs to be bumped, but it's now with `-rc1` so not sure what I should do.

PG15 isn't yet rolled out to prod; full testing finished today. Additionally waiting on the Postgres team to confirm Vault and related extensions are stable, after which we will deploy the AMI. The final image will get rebuilt and also contain this change

pcnc (dismissed):

ansible/files/gotrue.service.j2
darora: For consistency with other files with overrides, use plural `/etc/gotrue.overrides.env`

darora (approved):

PR #443, opened 2022-12-15
fix: grant pgsodium functions to service_role

soedirgo:
Using supautils's extension custom scripts. The script will be run on `create extension pgsodium`.

### DIFF
--- a/migrations/schema.sql
+++ b/migrations/schema.sql
@@ -511,25 +511,6 @@ $$;
 
 
 --
--- Name: key_encrypt_secret(); Type: FUNCTION; Schema: pgsodium; Owner: -
---
-
-CREATE FUNCTION pgsodium.key_encrypt_secret() RETURNS trigger
-    LANGUAGE plpgsql
-    AS $$
-    BEGIN
-            new.raw_key = CASE WHEN new.raw_key IS NULL THEN NULL ELSE
-                CASE WHEN new.parent_key IS NULL THEN NULL ELSE
-                        pgsodium.crypto_aead_det_encrypt(new.raw_key::bytea, pg_catalog.convert_to((new.id::text || new.associated_data::text)::text, 'utf8'),
-                new.parent_key::uuid,
-                new.raw_key_nonce
-              ) END END;
-    RETURN new;
-    END;
-    $$;
-
-
---
 -- Name: extension(text); Type: FUNCTION; Schema: storage; Owner: -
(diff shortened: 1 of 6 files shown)


migrations/schema.sql
```diff
@@ -510,25 +510,6 @@ END;
 $$;
 
 
---
--- Name: key_encrypt_secret(); Type: FUNCTION; Schema: pgsodium; Owner: -
---
-
-CREATE FUNCTION pgsodium.key_encrypt_secret() RETURNS trigger
```
soedirgo, line 517: The `schema` test has been failing for a while, so I guess this just reflects the latest pgsodium version.

migrations/schema.sql
```diff
@@ -754,35 +735,6 @@ CREATE TABLE auth.users (
 COMMENT ON TABLE auth.users IS 'Auth: Stores user login data within a secure schema.';
 
 
---
--- Name: decrypted_key; Type: VIEW; Schema: pgsodium; Owner: -
---
-
-CREATE VIEW pgsodium.decrypted_key AS
```
soedirgo, line 761: @michelp do you know why this got dropped now? If I recreate `pgsodium` it gets dumped again.

migrations/schema.sql
soedirgo: https://github.com/michelp/pgsodium/blob/060e27fc93a53999d94c58170627a443f0eb8d0c/sql/pgsodium--3.1.0--3.1.1.sql#L178

migrations/schema.sql
soedirgo: I think I got it - this was fixed in pgsodium 3.1.1 but now [this issue](https://github.com/supabase/postgres/blob/00837a237134dc0cb7ba3d48fa96b67cc55da2a3/migrations/db/migrations/20221207154255_create_pgsodium_and_vault.sql#L13-L17) is back, so we need to revert the migration.

soedirgo: (I'll hold off on reverting it until Michel's done with the pgsodium fixes)

pcnc (approved):

pcnc:
Looks good to me
@soedirgo is this ready to merge and test on staging, or are we waiting on feedback from Michel beforehand?

soedirgo:
I guess we can merge and test this on staging and ask him later - one less thing to fix at once

PR #451, opened 2022-12-16
Fix: Retain migrations sql file under /opt/migration/db path as part of AMI

Lakshmipathi:


sweatybridge (approved):

migrations/db/migrate.sh
```diff
@@ -56,3 +56,6 @@ fi
 
 # once done with everything, reset stats from init
 psql -v ON_ERROR_STOP=1 --no-password --no-psqlrc -U supabase_admin -c 'SELECT extensions.pg_stat_statements_reset(); SELECT pg_stat_reset();' || true
+
+# Create migration-output.sql file
+cd "$db"/migrations && ls -1 > /opt/migrations/db/migration-output.sql
```
sweatybridge, line 61:
```suggestion
cat "$db"/migrations/*.sql > /opt/migrations/db/migration-output.sql
```

Should be ok since wildcard expansion is alphabetical

sweatybridge (changes requested):
Actually a cat is missing

PR #454, opened 2022-12-16
fix: pgsodium extension custom script

soedirgo:
Add back tests and address build failures.

### DIFF
--- a/ansible/files/postgresql_extension_custom_scripts/pgsodium/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/pgsodium/after-create.sql
@@ -1,4 +1,3 @@
 grant execute on function pgsodium.crypto_aead_det_decrypt(bytea, bytea, uuid, bytea) to service_role;
 grant execute on function pgsodium.crypto_aead_det_encrypt(bytea, bytea, uuid, bytea) to service_role;
 grant execute on function pgsodium.crypto_aead_det_keygen to service_role;
-grant execute on function pgsodium.crypto_aead_det_noncegen to service_role;
--- a/ansible/tasks/internal/supautils.yml
+++ b/ansible/tasks/internal/supautils.yml
@@ -49,11 +49,17 @@
 
 - name: supautils - copy extension custom scripts
   copy:
-    src: files/postgresql_extension_custom_scripts
+    src: files/postgresql_extension_custom_scripts/
     dest: /etc/postgresql-custom/extension-custom-scripts
-    mode: 0664
+  become: yes
+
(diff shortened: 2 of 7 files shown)


ansible/files/postgresql_extension_custom_scripts/pgsodium/after-create.sql
```diff
@@ -1,4 +1,3 @@
 grant execute on function pgsodium.crypto_aead_det_decrypt(bytea, bytea, uuid, bytea) to service_role;
 grant execute on function pgsodium.crypto_aead_det_encrypt(bytea, bytea, uuid, bytea) to service_role;
 grant execute on function pgsodium.crypto_aead_det_keygen to service_role;
-grant execute on function pgsodium.crypto_aead_det_noncegen to service_role;
```
soedirgo, line 4: This is granted to `public`, so no need for the grant

ansible/tasks/internal/supautils.yml
```diff
@@ -49,11 +49,17 @@
 
 - name: supautils - copy extension custom scripts
   copy:
-    src: files/postgresql_extension_custom_scripts
+    src: files/postgresql_extension_custom_scripts/
```
soedirgo, line 52: the bug

migrations/db/migrations/20221207154255_create_pgsodium_and_vault.sql
```diff
@@ -6,6 +6,10 @@ grant pgsodium_keyiduser to postgres with admin option;
 grant pgsodium_keyholder to postgres with admin option;
 grant pgsodium_keymaker  to postgres with admin option;
 
+grant execute on function pgsodium.crypto_aead_det_decrypt(bytea, bytea, uuid, bytea) to service_role;
+grant execute on function pgsodium.crypto_aead_det_encrypt(bytea, bytea, uuid, bytea) to service_role;
+grant execute on function pgsodium.crypto_aead_det_keygen to service_role;
```
soedirgo, line 11: I have no idea why, but `pgsodium/after-create.sql` are not run during `create extension pgsodium` above, but it's run if I recreate the extension (tested on staging). This was the cause of the build failures.

I'll just add these since the grants are supposed to be idempotent.

soedirgo:
Build seems to run fine: https://github.com/supabase/postgres/actions/runs/3712752347/jobs/6294627349

pcnc:
Looks good - let's merge then roll it on staging to check if it's stable and all good and then we can promote it to a non-rc release. 
We can revert to the Vault image on staging afterwards if needed

pcnc (approved):

PR #461, opened 2022-12-22
feat: grant select on information_schema to supabase_auth_admin

hf:
GoTrue needs access to the information schema so that it can identify whether certain migrations or changes have occurred or are safe to run. This is often necessary because users add references or modify the schema in ways that could cause migrations to fail, and it needs a way to identify if migrations are safe to run.


soedirgo:
Is this still relevant? `supabase_auth_admin` has access to information schema in my testing. Which table/view does it not have access to?

hf:
@soedirgo Yes. This made it apparent: https://github.com/supabase/helper-scripts/pull/323

There are also some migrations in GoTrue that are broken because of this. For example this: https://github.com/supabase/gotrue/blob/master/migrations/20221011041400_add_mfa_indexes.up.sql#L4-L14 fails consistently on projects.

soedirgo:
How can I test that `supabase_auth_admin` is missing the privileges, and that this migration solves the issue? Can you also link to these migration failures? Testing on a new project, I'm getting:

```
~ 
❯ sac psql -p aexwgqchieipvynfujls
psql postgresql://supabase_admin:*****@db.aexwgqchieipvynfujls.supabase.co:5432/postgres
psql (15.1 (Homebrew))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
Type "help" for help.

postgres=# select * from information_schema.table_privileges where table_schema = 'information_schema' and table_name = 'check_constraints' and grantee = 'PUBLIC';
 grantor  | grantee | table_catalog |    table_schema    |    table_name     | privilege_type | is_grantable | with_hierarchy 
----------+---------+---------------+--------------------+-------------------+----------------+--------------+----------------
 postgres | PUBLIC  | postgres      | information_schema | check_constraints | SELECT         | NO           | YES
(1 row)
^^^^^ SELECT privilege is granted to PUBLIC, so all roles should be able to read the view
postgres=# set session authorization supabase_auth_admin;
SET
postgres=> select * from pg_catalog.pg_authid;
ERROR:  permission denied for table pg_authid
^^^^^ checking that supabase_auth_admin doesn't inherit privs from supabase_admin after `set session authorization`
postgres=> select * from information_schema.check_constraints;
 constraint_catalog | constraint_schema |             constraint_name             |                                 check_clause                 
                 
--------------------+-------------------+-----------------------------------------+--------------------------------------------------------------
-----------------
 postgres           | auth              | 16487_26962_3_not_null                  | updated_at IS NOT NULL
<trimmed>
 postgres           | auth              | 16487_26989_1_not_null                  | id IS NOT NULL
(46 rows)
^^^^^ success
postgres=> 
```

Or maybe it only fails on older projects?

sweatybridge:
I added a test for this using the latest docker image.

```bash
docker compose -f migrations/docker-compose.yml up
```

Seems like the `supabase_auth_admin` already has `select` permission to `information_schema.table_constraints` without this migration. May be rolling out the helper script alone is enough?

soedirgo:
> added a test

🔥

> May be rolling out the helper script alone is enough?

I'd rather not roll out permission changes only on existing projects - if we don't do it on new projects, the schema for older & newer projects will diverge.

We need to find out how it failed in the first place - my suspicion is it's `postgres` revoking the privileges from `public` or some such.

darora:

PR #470, opened 2023-01-07
Use GitHub action runner instead

sweatybridge:

## What is the new behavior?

## Additional context

### DIFF
--- a/.github/workflows/dockerhub-release.yml
+++ b/.github/workflows/dockerhub-release.yml
@@ -22,7 +22,7 @@ jobs:
 
   docker_x86_release:
     needs: settings
-    runs-on: [self-hosted, X64]
+    runs-on: ubuntu-latest
     timeout-minutes: 120
     env:
       arch: amd64


darora (commented):
Have we tested this out? Whats the reason we're changing this? Iirc we switched away to internal workers because the builds never finished on the GH hosted workers (CPU constraints or w/e)

sweatybridge:
Yes this is the run from my own fork using ubuntu-latest runner https://github.com/sweatybridge/postgres/actions/runs/3856067392/jobs/6571831699. The run built the image in 30 mins but failed to push because I didn't give it the token.

sweatybridge:
I think with ccache, the runner doesn't need as much cpu as before.

sweatybridge:
https://github.com/sweatybridge/postgres/actions/runs/3860186275/jobs/6580353194 here's a successful run without push. 

I suspect we are missing some configuration for our self hosted x86 runner. May be the docker credentials were not mounted as secrets?

darora:
> I think with ccache, the runner doesn't need as much cpu as before.

Wouldn't that cause things to fail again if a PR changed enough things that the cache couldn't be re-used?

I'm happy to approve this if you'd like, but it doesn't sound like we understand the root cause for the failures, given things were building OK three weeks ago.

darora (approved):
approving, if you prefer unblocking with this PR

sweatybridge:
Thanks, I will check with @pcnc next week. May be the runner had an old version of ccache image in its docker cache.

PR #472, opened 2023-01-10
Add pgvector extension

gregnr:

Feature

Popular ML extension [`pgvector`](https://github.com/pgvector/pgvector) is unavailable in Supabase. The closest related extension today is [`cube`](https://www.postgresql.org/docs/current/cube.html), but its dimension limit (100) makes it impractical for modern vector math tasks. 

## What is the new behavior?

Adds [`pgvector`](https://github.com/pgvector/pgvector) extension, one of the [highest voted extensions](https://github.com/supabase/supabase/discussions/679#discussioncomment-2850265) on the [Vote for Postgres extensions](https://github.com/supabase/supabase/discussions/679) discussion.

## Additional context

As the language model ecosystem is increasing in popularity (eg. GPT-3), there is becoming a need for efficient vector operations (like vector similarity search) on large vector-based datasets (eg. [GPT-3 Embeddings](https://beta.openai.com/docs/guides/embeddings)). The ability to use Postgres for this type of work vs. outsourcing to another system would make Supabase a compelling service offering for the AI/ML community.

This PR mirrors best practices and conventions used by other extensions to install `pgvector` through Ansible.

Please let me know if there is any additional work I can do to help bring this extension to Supabase.


olirice:
Thanks for the PR!

`pgvector` looks great. I'm comparing it with the (available) `cube` extension and there is a lot of overlap. Is the only issue with using `cube` the 100 element limit?

I see in the [cube docs](https://www.postgresql.org/docs/current/cube.html)
> To make it harder for people to break things, there is a limit of 100 on the number of dimensions of cubes. This is set in cubedata.h if you need something bigger.

so we could potentially bump that cap over the standard embedding sizes to something like 4096

the other features I see in `pgvector` that aren't in `cube` are some niche index operations and cosine distance support. Since euclidean distance (supported by both) is equal to cosine distance when operating on unit vectors that doesn't seem like too bad of a constraint to work around.

If there are other differences that are significant for common uses cases could you please add a few notes wrt what they are?

michelp:
+1 for pgvector. cube has not hada major update since 2006, if we patch it then we own it.  pgvector is targeted directly at very common vector embedding use cases and is actively maintained.

olirice (approved):

olirice (approved):

olirice:
@supabase/backend please +1 and merge when you get a sec

mattaylor:
bump

Lakshmipathi (approved):

gregnr:
@olirice @Lakshmipathi Thanks for taking the time to review & thanks for the approvals. There were a few new merge conflicts as of yesterday, but I have resolved these now.

olirice:
thanks @gregnr, looks great!
sorry for the slow cycle time on this one

harshit0209:
cheers

PR #474, opened 2023-01-10
chore: cleanup extension sources before proceeding to next task

pcnc:

* cleans up extension sources to avoid builder disk filling up
* speeds up Groonga builds
* adds parallelization support to the Docker build
  * this requires a new version of Ansible's make task type, provided by installing the general Ansible collection


sweatybridge (approved):

dragarcia (approved):

ansible/tasks/postgres-extensions/13-plv8.yml
```diff
@@ -58,4 +64,3 @@
   poll: 0
   register: plv8_build
   when: async_mode
```
Lakshmipathi, line 66: For async_mode, we need to perform cleanup from [99-finish_async_tasks.yml](https://github.com/supabase/postgres/blob/develop/ansible/tasks/postgres-extensions/99-finish_async_tasks.yml)

Lakshmipathi (approved):

ansible/tasks/postgres-extensions/13-plv8.yml
pcnc: Good catch - thanks!

PR #481, opened 2023-01-13
Add pg_netstat extension

burmecia:

This PR is to add [pg_netstat extension](https://github.com/supabase/pg_netstat), which is used to gather data for database egress metric.

Currently there is no reliable way to gather accurate database egress.

## What is the new behavior?

By using this extension, we can capture network packets in postgres and expose it as a view so Victoria Metrics can gather.

## Additional context

This extension is configured to capture both default device like `ens5` and loopback device, capture interval is set to 60 seconds.


burmecia:
Wait for arm release is ready in https://github.com/supabase/pg_netstat/releases/tag/v0.1.0, then we're good to go.

PR #485, opened 2023-01-18
chore: update pg upgrade scripts

pcnc:
## Fixes

- [x]  PG_VERSION file did not exist
- [x]  Initiate script doesn’t handle loop devices existing or something (unclear)
    - initiate script assumes the only present block devices are located on `/dev/nvme*` and excludes `nvme[0-1]`, mounting the following device
    - older instances (pre-ebssurrogate?) also have `/dev/loop*` devices present
- [x]  Stop Postgrest, gotrue while we’re upgrading (or restoring)
    - retries for this job end up stopping the newer instance as well
- [x]  poll upgrade completion status
- [x]  use pgsodium key stored in middleware db if it does not exist
- [x]  handle `special device` errors
- [x] run generated `.sql` files such as `update-extensions.sql ` on new instance
- [x] bail if complete operation already in progress


ansible/files/admin_api_scripts/pg_upgrade_complete.sh
```diff
@@ -6,40 +6,65 @@
...
         run_sql "CREATE EXTENSION IF NOT EXISTS ${EXTENSION} CASCADE;"
     done
-    
-    sleep 5
-    service postgresql restart
 
     sleep 5
     service postgresql restart
+
+    if [[ $(systemctl is-active gotrue) == "inactive" ]]; then
```
darora: wouldn't this incorrectly start these services for projects with a separate kps instance?

pcnc: Yup. removed in updated script versions. It's up to the worker to manage services

Lakshmipathi (approved):

PR #486, opened 2023-01-19
feat: add pg_egress_collect service

burmecia:

This PR is to add service for postgres egress metric collection. See more context: https://www.notion.so/supabase/Fix-DB-egress-metric-190164a5c4e5444cbbe20c726cea241d

Currently there is no reliable way to gather accurate database-only egress, so the whole server egress is used instead.

## What is the new behavior?

The solution is using `tcpdump` to capture outgoing TCP packets on port 5432 and 6543, and a perl script extract packet length and sum up to one minute interval. The result is saved to text file at `/tmp/pg_egress_collect.txt` by default. Admin API then can read metric from that file and expose it to Victoria Metrics.

By using this approach, we can collect accurate network egress for postgres and expose it in admin api so Victoria Metrics can gather for downstream pipeline.

## Additional context

This metric will also be as data source for admin api, see more details: https://github.com/supabase/supabase-admin-api/pull/130

### DIFF
--- a/ansible/files/admin_api_scripts/pg_egress_collect.pl
+++ b/ansible/files/admin_api_scripts/pg_egress_collect.pl
@@ -1,0 +1,110 @@
+#!/usr/bin/env perl
+
+# This script receive tcpdump output through STDIN and does:
+#
+# 1. extract outgoing TCP packet length on all devices port 5432 and 6543
+# 2. sum the length up to one minute
+# 3. save the total length to prom file (default is /tmp/pg_egress_collect.prom)
+#
+# Usage:
+#
+# tcpdump -s 128 -Q out -i any -nn -tt -vv -p -l 'tcp and (port 5432 or port 6543)' | perl pg_egress_collect.pl /tmp/output.prom
+#
+
+use POSIX;
+use List::Util qw(sum);
+use IO::Async::Loop;
+use IO::Async::Stream;
+use IO::Async::Timer::Periodic;
+
+use strict;
+use warnings;
+
+# total captured packets lenth in a time frame
+my $captured_len = 0;
+
+# extract tcp packet length captured by tcpdump
+#
+# Sample input lines:
+#
+# 1674013833.940253 IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 60)
+#     10.112.101.122.5432 > 220.235.16.223.62599: Flags [S.], cksum 0x5de3 (incorrect -> 0x63da), seq 2314200657, ack 2071735457, win 62643, options [mss 8961,sackOK,TS val 3358598837 ecr 1277499190,nop,wscale 7], length 0
+# 1674013833.989257 IP (tos 0x0, ttl 64, id 24975, offset 0, flags [DF], proto TCP (6), length 52)
+#     10.112.101.122.5432 > 220.235.16.223.62599: Flags [.], cksum 0x5ddb (incorrect -> 0xa25b), seq 1, ack 9, win 490, options [nop,nop,TS val 3358598885 ecr 1277499232], length 0
+sub extract_packet_length {
+    my ($line) = @_;
+
+    #print("debug: >> " . $line);
+
+    if ($line =~ /^(\d+)\.\d+/) {
+        # extract packet timestamp, do nothing for now
+        my $ts = $1;
+    } elsif ($line =~ /^\s+\d+\.\d+\.\d+\.\d+\..*, length (\d+)$/) {
+        # extract tcp packet length and add it up
+        my $len = $1;
+        $captured_len += $len;
+    }
+}
+
+# write total length to prom file
+#
+# ref: https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format
+sub write_prom_file {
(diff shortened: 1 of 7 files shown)


darora:
@burmecia have you collected data on performance overhead for this across a few projects with different utilizations?

burmecia:
I tried several projects including my personal projects and RevOps project, not much overhead I can see. CPU usage is unnoticeable, memory is about 10M and only write one line to a tmp file per minute.

burmecia:
Tested its performance on my project, below are the setups:

- Used sql to generate random serial data and read it to local using psql.
  ```sql
  SELECT time, random()*100 as cpu_usage
  FROM generate_series(now() - INTERVAL '12 months',now(),INTERVAL '1 minute') as time
  ```
- Kept monitoring egress for about 1 hour
- Total egress is about 5GB
- No other database activities

And the test result is:

- Average CPU usage is around 10% including postgres
- Average memory usage by the egress collector is about 15M (tcpdump + perl script)
- Disk space used: <10 bytes
- EBS IO balance: 99% all the time

ansible/files/admin_api_scripts/pg_egress_collect.pl
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env perl
```
darora, line 1: needs to be reviewed by someone familiar with perl

ansible/files/adminapi.sudoers.conf
```diff
@@ -25,3 +26,4 @@ Cmnd_Alias PGBOUNCER = /bin/systemctl start pgbouncer.service, /bin/systemctl st
 %adminapi ALL= NOPASSWD: POSTGREST
 %adminapi ALL= NOPASSWD: GOTRUE
 %adminapi ALL= NOPASSWD: PGBOUNCER
+%adminapi ALL= NOPASSWD: PG_EGRESS_COLLECT
```
darora: changes in this file are not needed afaict?

ansible/files/admin_api_scripts/pg_egress_collect.pl
```diff
@@ -0,0 +1,106 @@
...
+#     10.112.101.122.5432 > 220.235.16.223.62599: Flags [S.], cksum 0x5de3 (incorrect -> 0x63da), seq 2314200657, ack 2071735457, win 62643, options [mss 8961,sackOK,TS val 3358598837 ecr 1277499190,nop,wscale 7], length 0
+# 1674013833.989257 IP (tos 0x0, ttl 64, id 24975, offset 0, flags [DF], proto TCP (6), length 52)
+#     10.112.101.122.5432 > 220.235.16.223.62599: Flags [.], cksum 0x5ddb (incorrect -> 0xa25b), seq 1, ack 9, win 490, options [nop,nop,TS val 3358598885 ecr 1277499232], length 0
+sub extract_packet_length {
+    my ($line) = @_;
+
+    #print("debug: >> " . $line);
+
+    if ($line =~ /^(\d+)\.\d+/) {
+        # extract packet timestamp, do nothing for now
```
darora: not familiar with perl -- when do we do something with the timestamp?

ansible/files/admin_api_scripts/pg_egress_collect.pl
```diff
@@ -0,0 +1,106 @@
...
+                extract_packet_length($line);
+            }
+
+            return 0;
+        },
+    );
+
+    # schedule file writer per minute
+    my $writer = IO::Async::Timer::Periodic->new(
+        interval => 60,
```
darora: parameterize as an arg - we might be changing this soon

ansible/files/adminapi.sudoers.conf
burmecia: ok, removed.

ansible/files/admin_api_scripts/pg_egress_collect.pl
burmecia: that was supposed to be used in the future if we need timestamp, but for now it seems not need, will remove it.

ansible/files/admin_api_scripts/pg_egress_collect.pl
burmecia: done

ansible/files/admin_api_scripts/pg_egress_collect.pl
```diff
@@ -0,0 +1,106 @@
...
+        on_tick => sub {
+            write_file($output);
```
michelp, line 88: If there's an error here on write, the default (IIR my Perl) is to close the file, not sure if this will cause the event loop to close and systemd will restart it, or should you add an on_write_error to die explicitly?  Might end up being the same either way but something to consider.

michelp (approved):

ansible/files/admin_api_scripts/pg_egress_collect.pl
burmecia: yes, if the perl script exits the pipe will fail, thus systemd will restart the whole service. It has been tested on my project. To be more safe, I added `die` check when closing the file handle.

darora (approved):

PR #515, opened 2023-01-27
fix: enable pg_egress_collect service

burmecia:

This PR is to enable `pg_egress_collect` service, which was introduced in #486 but not enabled yet.

Currently the service is not enabled, so it is not started when boot.

## What is the new behavior?

This service should start when server is booted.

## Additional context

N/A


darora:
We don't want it enabled until after everything critical has been brought up - this will need to be done in init-scripts

burmecia:
Raised https://github.com/supabase/infrastructure/pull/11427 to enable `pg_egress_collect` service on boot.

PR #519, opened 2023-02-02
chore: create tags and GH releases when we build an AMI

darora:
Adding tags makes it easier to diff the changes that have been made
between releases. At the moment, this information only lives in the
contents of the repository (and we haven't really maintained a
changelog).


darora:
Assuming we're OK with this approach, I'll likely backfill our more recent releases manually.

sweatybridge (approved):

pcnc (approved):

PR #527, opened 2023-02-06
fix: capture db egress traffic on non-loopback interface only

burmecia:

This PR is to change db egress collection config to let it capture on non-loopback interface only.

Currently the db egress collection capture traffic on all network devices, it will be double counted when combined bill with API egress. So the db egress collection should only capture traffic on interface like `ens5` not `lo`.

## What is the new behavior?

The tcpdump will automatically detect the 1st non-loopback interface and only capture packets on it.

## Additional context

N/A


darora (approved):
Approving, with the note that this could lead to your pipeline silently failing or collecting incorrect data if something changes about our server builds.

PR #528, opened 2023-02-07
chore: improve build cache with multistage docker

sweatybridge:

chore

When an extension is upgraded, it's infeasible to test the new version because the PR build takes 2h.

## What is the new behavior?

Separate compilation into its own docker stage. When one dependency changes, only that dependency is recompiled. The GHA cache ensures that default branch is used to load other dependencies. A cached build now takes less than a minute.

Also reduces uncompressed image size on x86 macOS from 4GB to 1.3GB (without plv8).

## Additional context

- [x] use ccache to build plv8
- [x] tested on my [own fork](https://github.com/sweatybridge/postgres/actions/runs/4104565638/jobs/7080219111)


Lakshmipathi (approved):

ansible/Dockerfile
```diff
@@ -0,0 +1,705 @@
...
+ARG groonga_release=12.0.8
+ARG groonga_release_checksum=sha256:7770c0ff6804ef4b47b015b15736cd973cffced977c20991b16b2daa4fea6eeb
+ARG pgroonga_release=2.4.0
+ARG pgroonga_release_checksum=sha256:5baaae0e7d81f8167e278e9a34c6ed56aece8b34f5ab98f228c64408093417b3
+ARG wrappers_release=0.1.7
+ARG hypopg_release=1.3.1
+ARG hypopg_release_checksum=sha256:e7f01ee0259dc1713f318a108f987663d60f3041948c2ada57a94b469565ca8e
+ARG pg_repack_release=1.4.8
+ARG pg_repack_release_checksum=sha256:18b4d871c1abf78cf0b1b1fe6081d435d183a8dc5eb977576e7a47fe113dd4ec
+ARG pgvector_release=0.4.0
```
darora, line 57: can we please avoid duplicating these version strings?

darora (commented):
I dunno about duplicating the entire build process in this file

ansible/Dockerfile
sweatybridge: These are default values which we can change via `--build-arg` flag. I will dedupe in a separate PR by passing in values from vars.yaml.

Currently there's a few inconsistencies in vars.yaml with using sha1 checksum and git branch ref. I will update everything to sha256 and commit ref in the follow up PR.

darora: Is there a reason we can't remove the defaults altogether? Can we update this PR with the vars.yml plumbing?

In general, I'm wary of the duplicated build process. Is there a path forward to getting rid of the duplication?

sweatybridge: > Is there a path forward to getting rid of the duplication?

One possible path forward is to package each extension as a .deb file and publish to an internal registry / s3. Then it would be trivial (see pg_graphql for eg) for docker and ansible to install these packages separately from a single vars file.

Another possibility is to extract the docker image contents into a base VM. Fly.io uses this approach to bootstrap arbitrary firecracker VM. Our ansible tasks can work on top of this base VM to do other configurations, like updating postgresql config.

sweatybridge: Another possibility to extract the docker image contents into a base VM. Fly.io uses this approach to bootstrap arbitrary firecracker VM.

Our ansible tasks can work on top of this base VM to do other configurations, like updating postgresql config.

sweatybridge: > Can we update this PR with the vars.yml plumbing?

I've created a new PR based on top of this to illustrate the plumbing we can do: https://github.com/supabase/postgres/pull/529

> Is there a reason we can't remove the defaults altogether?

These variables are currently inconsistent with the rest of extensions. Once we update them, should be ok to remove the defaults.
- `wal2json_commit_sha`: we probably want to rename as `wal2json_release` for consistency
- `pljava_release`/`pljava_release_checksum`: update sha256 after upstream PR merge
- `plv8_commit_version`: rename to `plv8_release`
- `pg_plan_filter_commit_version`: rename to `pg_plan_filter_release`
- `libsodium_release`/`libsodium_release_checksum`: remove after vault release
- `hypopg_commit_sha`: rename to `hypopg_release` and use release tag
- `pg_stat_monitor_release_checksum`/`pgvector_release_checksum`: currently missing from vars.yaml

PR #532, opened 2023-02-09
fix: add runtime dependency to pg_net package

sweatybridge:

chore

```bash
waiting for server to start.... 2023-02-09 05:01:47.990 UTC [157] LOG:  pgaudit extension initialized
 2023-02-09 05:01:47.993 UTC [157] FATAL:  could not load library "/usr/lib/postgresql/15/lib/pg_net.so": libcurl-gnutls.so.4: cannot open shared object file: No such file or directory
```

## What is the new behavior?

- specify runtime dependency for `pg_net` and `pgsodium` via checkinstall
- remove pljava (bloats up the docker image by 50% and rarely used)
  - reduces [compressed size](https://hub.docker.com/layers/sweatybridge/postgres/latest/images/sha256-3b01d5cedb1ae6cf0da8669287e0b3b99ec74fc7280d7d9c8d2271f3b9d8279b?context=explore) from 730MB to 207.58MB
  - memory usage reduced by 150MB (see screenshot)
- builds plv8 separately because the source layer occupies 24GB of disk space and takes 10 mins to download

## Additional context

Screenshot 2023-02-09 at 5 08 00 PM


darora (commented):
Are we deprecating and removing pljava from the platform? And plv8, for that matter?

darora:
To back up a bit, my impression is that the docker image supports local dev workflows for the hosted platform. Is that not the case? Or is removing these extensions somehow not detrimental to that goal?

sweatybridge:
I turned off plv8 because it's taking too long to build and not really needed for PR tests. Let me check with @pcnc on how to update ccache image before enabling plv8. This should be supported on the platform.

We discussed briefly during cli meeting to deprecate pljava on the platform. Do you think it's ok to do so from infra side?

> To back up a bit, my impression is that the docker image supports local dev workflows for the hosted platform.

Yes, this is still true. Sorry about any confusion introduced by this PR. My intention is to test this new image a bit more before switching over our self-hosted [supabase/postgres](https://hub.docker.com/r/supabase/postgres) image to this one. Currently the docker release action is still using the [root Dockerfile](https://github.com/supabase/postgres/blob/develop/Dockerfile) so no change is exposed to users yet.

ansible/tasks/postgres-extensions/13-plv8.yml
```diff
@@ -37,6 +37,13 @@
   when: platform == "arm64"
   ignore_errors: yes # not needed for docker build
 
+- name: plv8 - enable ccache
+  become: yes
+  replace:
+    path: /tmp/plv8/Makefiles/Makefile.docker
+    regexp: '^GN_ARGS ='
+    replace: GN_ARGS = cc_wrapper="env CCACHE_SLOPPINESS=time_macros ccache"
```
sweatybridge: Using instructions from macos which also applies to linux https://chromium.googlesource.com/chromium/src/+/HEAD/docs/ccache_mac.md

darora (approved):

PR #533, opened 2023-02-09
chore: update upgrade scripts

pcnc:

* Initiate script now also disables extensions selectively per initial pg version
  * Generates script to re-enable extensions, if they exist on the resulting upgraded instance
* Covers case in which latest libpq wasn't available for binaries used during the upgrade process 
* Improved logging
* Dry-run mode


ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
```diff
@@ -37,78 +69,143 @@ cleanup() {
...
-        run_sql "CREATE EXTENSION IF NOT EXISTS ${EXTENSION} CASCADE;"
-    done
+    if [ "$IS_DRY_RUN" = false ]; then
+        echo "Restarting postgresql"
+        systemctl restart postgresql
+    fi
+
+    echo "Re-enabling extensions"
+    if [ -f $POST_UPGRADE_EXTENSION_SCRIPT ]; then
+        run_sql -f $POST_UPGRADE_EXTENSION_SCRIPT
```
darora, line 82: might be good to run these scripts through shellcheck -- it'd probably get mad about all the non-quoted variables

ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
```diff
@@ -37,78 +69,143 @@ cleanup() {
...
+    if [ -d "/var/lib/postgresql/extension/" ]; then
+        cp /root/pg_upgrade_pgsodium_getkey.sh "/var/lib/postgresql/extension/pgsodium_getkey"
+        chown postgres:postgres "/var/lib/postgresql/extension/pgsodium_getkey"
+    fi
+
+    chown -R postgres:postgres "/tmp/pg_upgrade_bin/$PGVERSION"
+
+    # Make latest libpq available to pg_upgrade
+    mkdir -p /usr/lib/postgresql/lib/aarch64
+    if [ ! -L /usr/lib/postgresql/lib/aarch64/libpq.so.5 ]; then
```
darora, line 157: hmm do we block upgrades on the older x86 instances at some early stage?

darora (approved):

ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
pcnc: Good point! Added a [shellcheck workflow](https://github.com/supabase/postgres/actions/runs/4162177961/jobs/7201028544) to help keep the admin_api_scripts folder in check

ansible/files/admin_api_scripts/pg_upgrade_initiate.sh
pcnc: Not an explicit one - it's tied down to version eligibility since the last version supporting x86 was 12.5.
Will add additional checks in the admin cli and upcoming API support

PR #542, opened 2023-02-16
chore: pg_tle in Docker

dragarcia:


ansible/tasks/docker/setup.yml
```diff
@@ -21,7 +21,6 @@
       - systemtap-sdt-dev
       - ccache
       - clang
-      - flex
```
dragarcia, line 24: @michelp I have moved this over [here](https://github.com/supabase/postgres/blob/49563a6eb20a82446117d171637741018813e9ca/ansible/Dockerfile#L693) for now. Also, is this a Docker only dependency or should this also be included in [this file](https://github.com/supabase/postgres/blob/49563a6eb20a82446117d171637741018813e9ca/ansible/tasks/postgres-extensions/29-pg_tle.yml)?

sweatybridge (approved):
lgtm. I will refresh the build cache after merging

sweatybridge:
Let's also add a test for enabling this extension? similar to https://github.com/supabase/postgres/blob/develop/migrations/tests/extensions/28-pgvector.sql

PR #550, opened 2023-02-20
chore: fix PG AMI build

pcnc:

* `pg_prove` is now run as supabase_admin, which might create objects that cannot be dropped by the postgres role: https://github.com/supabase/postgres/commit/6ecf278dd3cbf24370fd19694d5ae0e790d0f257
* already built off this branch

## What is the new behavior?

## Additional context

### DIFF
--- a/ansible/tasks/test-image.yml
+++ b/ansible/tasks/test-image.yml
@@ -52,11 +52,11 @@
   when: ebssurrogate_mode
 
 - name: Drop pgTap extension
-  shell: /usr/lib/postgresql/bin/psql -U postgres -h localhost -d postgres -c "DROP extension if exists pgtap";
+  shell: /usr/lib/postgresql/bin/psql -U supabase_admin -h localhost -d postgres -c "DROP extension if exists pgtap";
   when: ebssurrogate_mode
 
 - name: Drop extension test function
-  shell: /usr/lib/postgresql/bin/psql -U postgres -h localhost -d postgres -c "DROP FUNCTION IF EXISTS install_available_extensions_and_test";
+  shell: /usr/lib/postgresql/bin/psql -U supabase_admin -h localhost -d postgres -c "DROP FUNCTION IF EXISTS install_available_extensions_and_test";
   when: ebssurrogate_mode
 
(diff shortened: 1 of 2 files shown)


sweatybridge (approved):

darora (approved):

ansible/tasks/test-image.yml
```diff
@@ -52,11 +52,11 @@
   when: ebssurrogate_mode
 
 - name: Drop pgTap extension
-  shell: /usr/lib/postgresql/bin/psql -U postgres -h localhost -d postgres -c "DROP extension if exists pgtap";
+  shell: /usr/lib/postgresql/bin/psql -U supabase_admin -h localhost -d postgres -c "DROP extension if exists pgtap";
```
sweatybridge, line 55: probably better to wrap creation of pgtap extension in the txn instead  https://github.com/supabase/postgres/blob/develop/migrations/tests/test.sql#L4 

~Then we can remove this step.~ seems like it's created again [here](https://github.com/supabase/postgres/pull/550/files#diff-2d01095c3e13c0a570e306aa7f34fbb31b8e1e1612708dceb3b0b752628fe89bR21)

ansible/tasks/test-image.yml
```diff
@@ -52,11 +52,11 @@
...
   when: ebssurrogate_mode
 
 - name: Drop extension test function
-  shell: /usr/lib/postgresql/bin/psql -U postgres -h localhost -d postgres -c "DROP FUNCTION IF EXISTS install_available_extensions_and_test";
+  shell: /usr/lib/postgresql/bin/psql -U supabase_admin -h localhost -d postgres -c "DROP FUNCTION IF EXISTS install_available_extensions_and_test";
```
sweatybridge, line 59: Also on closer look, this function is created by [unit-tests](https://github.com/supabase/postgres/pull/550/files#diff-2d01095c3e13c0a570e306aa7f34fbb31b8e1e1612708dceb3b0b752628fe89bR25) which is run as postgres user. So this change is probably unnecessary?

PR #552, opened 2023-02-21
chore: rollback extensions from pgtap tests

sweatybridge:

tests

## What is the new behavior?

wrap individual extension tests in transactions

## Additional context

considered wrapping extensions.sql but that leads to error 
```
migrations-pg_prove-1  | psql:/tests/extensions/11-wal2json.sql:2: ERROR:  cannot create logical replication slot in transaction that has performed writes
```


darora (approved):
Please increment the version to create a new release

PR #560, opened 2023-02-24
fix: user permissions for extension schemas

### LINKED ISSUE #12588: `postgres` role should have `select` permissions on tables created/managed by extensions
# Bug report

## Describe the bug

Hey there, I'm trying to create a table of normalized address using the extensions `address_standardizer`, `address_standardizer_data_us`, and `postgis` in the `extensions` schema and `postgis_tiger_geocoder` in the `tiger` schema. I can access the schemas & functions, but not tables in the schema.

For example, with the following query:
```sql
> SELECT tiger.pprint_addy(tiger.pagc_normalize_address('710 E Ben White Blvd, Austin, TX 78704'));

ERROR:  permission denied for table street_type_lookup
CONTEXT:  SQL function "is_pretype" statement 1
PL/pgSQL function pprint_addy(norm_addy) line 9 at assignment
```

And indeed, if I try to select directly from the table in question I do not have permissions.
```sql
> select * from tiger.street_type_lookup;

ERROR:  permission denied for table street_type_lookup
```

I'm connected to the database as the `postgres` user. And, as there is no access to the DB superuser, I cannot grant `postgres` select access to the tables.

This problem does not occur on local setup using supabase-cli, only on cloud supabase instances.

Also, this issue occurs with tables created by `address_standardizer_data_us`. One of these tables is called `us_lex`.

## To Reproduce

Create a new supabase project, connect via psql, then run:

```plpgsql
BEGIN;

CREATE SCHEMA IF NOT EXISTS extensions;
CREATE SCHEMA IF NOT EXISTS private;
CREATE SCHEMA IF NOT EXISTS tiger;

CREATE EXTENSION IF NOT EXISTS address_standardizer WITH SCHEMA extensions;
CREATE EXTENSION IF NOT EXISTS address_standardizer_data_us WITH SCHEMA extensions;
CREATE EXTENSION IF NOT EXISTS postgis WITH SCHEMA extensions;
CREATE EXTENSION IF NOT EXISTS fuzzystrmatch WITH SCHEMA extensions;
CREATE EXTENSION IF NOT EXISTS postgis_tiger_geocoder WITH SCHEMA tiger;

SELECT tiger.pprint_addy(tiger.pagc_normalize_address('710 E Ben White Blvd, Austin, TX 78704'));

ROLLBACK;
```

## Expected behavior

I expect to have select permissions on tables in added extensions.

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

sweatybridge:

Bug fix https://github.com/supabase/supabase/issues/12588

## What is the new behavior?

- Schemas created by postgis tiger extension are owned by supabase_admin. We now give full grant option to postgres so that platform users can further grant permissions to other roles. 
- Tables, functions, and sequences in extensions schema are now accessible to postgres by default.

## Additional context

### DIFF
--- a/migrations/db/migrations/20230224042246_grant_extensions_perms_for_postgres.sql
+++ b/migrations/db/migrations/20230224042246_grant_extensions_perms_for_postgres.sql
@@ -1,0 +1,10 @@
+-- migrate:up
+grant all privileges on all tables in schema extensions to postgres with grant option;
+grant all privileges on all routines in schema extensions to postgres with grant option;
+grant all privileges on all sequences in schema extensions to postgres with grant option;
+alter default privileges in schema extensions grant all on tables to postgres with grant option;
+alter default privileges in schema extensions grant all on routines to postgres with grant option;
+alter default privileges in schema extensions grant all on sequences to postgres with grant option;
+
+-- migrate:down
+
--- a/ansible/files/postgresql_extension_custom_scripts/postgis_tiger_geocoder/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/postgis_tiger_geocoder/after-create.sql
@@ -1,9 +1,10 @@
+-- These schemas are created by extension to house all tiger related functions, owned by supabase_admin
 grant usage on schema tiger, tiger_data to postgres with grant option;
-
+-- Give postgres permission to all existing entities, also allows postgres to grant other roles
 grant all on all tables in schema tiger, tiger_data to postgres with grant option;
 grant all on all routines in schema tiger, tiger_data to postgres with grant option;
 grant all on all sequences in schema tiger, tiger_data to postgres with grant option;
-
+-- Update default privileges so that new entities are also accessible by postgres
 alter default privileges in schema tiger, tiger_data grant all on tables to postgres with grant option;
 alter default privileges in schema tiger, tiger_data grant all on routines to postgres with grant option;
 alter default privileges in schema tiger, tiger_data grant all on sequences to postgres with grant option;
--- a/ansible/files/postgresql_extension_custom_scripts/postgis_tiger_geocoder/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/postgis_tiger_geocoder/after-create.sql
@@ -1,0 +1,9 @@
+grant usage on schema tiger to postgres with grant option;
+
+grant all on all tables in schema tiger to postgres with grant option;
+grant all on all routines in schema tiger to postgres with grant option;
+grant all on all sequences in schema tiger to postgres with grant option;
+
+alter default privileges in schema tiger grant all on tables to postgres with grant option;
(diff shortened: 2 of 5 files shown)


migrations/db/migrations/20230224042246_grant_extensions_perms_for_postgres.sql
```diff
@@ -0,0 +1,8 @@
+-- migrate:up
+grant select on all tables in schema extensions to postgres, anon, authenticated, service_role;
+grant execute on all functions in schema extensions to postgres, anon, authenticated, service_role;
+alter default privileges in schema extensions grant select on tables to postgres, anon, authenticated, service_role;
+alter default privileges in schema extensions grant execute on functions to postgres, anon, authenticated, service_role;
```
soedirgo: Not sure about the implications of doing this - maybe it's safer to add custom scripts per extension as needed? There may be some extensions like pgsodium where it's not safe to expose certain functions to some of the roles.

sweatybridge: How about giving postgres grant option? I think minimally platform users should be able to control the roles that can access their enabled extensions. That way they can self serve such use cases in the future.

```sql
grant select on all tables in schema extensions to postgres with grant option;
grant execute on all functions in schema extensions to postgres with grant option;
alter default privileges in schema extensions grant select on tables to postgres with grant option;
alter default privileges in schema extensions grant execute on functions to postgres with grant option;
```

sweatybridge: seems like postgres already has these grant options on extensions schema.

sweatybridge: Pgsodium creates its own schema rather than installing to extensions. As a general rule, we can make extensions schema public and if there's something users shouldn't change, they can be installed in their own schema.

soedirgo: Yeah, giving `postgres` grant option sounds fine.

soedirgo (approved):

soedirgo:
Not sure if we should also run grants on `tiger_data`, seems to be an empty schema

sweatybridge:
> Another schema called tiger_data is also created which houses all the census data for each state that the loader downloads from Census site and loads into the database. In the current model, each set of state tables is prefixed with the state code e.g ma_addr, ma_edges etc with constraints to enforce only that state data. Each of these tables inherits from the tables addr, faces, edges, etc located in the tiger schema.

Seems like some kinda of runtime cache https://postgis.net/docs/Extras.html. I will add it just in case.

dragarcia (approved):

PR #574, opened 2023-03-06
feat: add supautils to docker image

sweatybridge:

feature

## What is the new behavior?

Add supautils extension to docker image for feature parity with hosted ami. Most notably this allows extension permissions to be setup properly via after-create hooks.

## Additional context

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ ARG hypopg_release=1.3.1
 ARG pg_repack_release=1.4.8
 ARG pgvector_release=0.4.0
 ARG pg_tle_release=1.0.1
+ARG supautils_release=1.7.2
 
 FROM postgres:${postgresql_release} as base
 # Redeclare args for use in subsequent stages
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -714,6 +715,33 @@ RUN --mount=type=cache,target=/ccache,from=public.ecr.aws/supabase/postgres:ccac
...
+# Install build dependencies
+RUN apt-get update && apt-get install -y \
+    libicu-dev \
+    && rm -rf /var/lib/apt/lists/*
+# Build from source
+WORKDIR /tmp/supautils-${supautils_release}
+RUN --mount=type=cache,target=/ccache,from=public.ecr.aws/supabase/postgres:ccache \
+    make -j$(nproc)
+# Create debian package
+RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --nodoc
```
soedirgo: Do we still need to build it from source?

sweatybridge: It's mostly for documentation at this point since the production stage doesn't depend on supautils-source stage. But I can also remove it to align with graphql extension

dragarcia (approved):

PR #591, opened 2023-03-27
chore: trigger build

inian:


inian:
Made a boo-boo in testing this end to end :\

Forgot to add a semver version while triggering a build from the branch
https://github.com/supabase/postgres/actions/runs/4529140811/jobs/7976573350#step:11:14

We can either cancel the workflow when this PR is merged or I can clean up the AMIs generated. Former is better I think. 

Tested the new AMIs works as expected on staging though

darora (approved):

PR #598, opened 2023-03-29
chore: enable Vault

dragarcia:
* Reversal of https://github.com/supabase/postgres/pull/566


common.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "15.1.0.62"
+postgres-version = "15.1.0.63-rc"
```
darora: if we need to test this on staging, could you create the build off this branch and do that before merge?

dragarcia: sure, roger that

pcnc (approved):

PR #604, opened 2023-04-02
WIP:  Adding pg_ivm extension to Postgres

allquantor:

Adding a PG extension 

pg_ivm is not included. 

## What is the new behavior?
pg_ivm is included

## Additional context

Is there any doc on tests or other conventions I should add?


allquantor:
The build process could be more straightforward. I need help putting things together. It would be great to have some indication of how to build locally and reuse in different contexts.

olirice:
hi @allquantor 
thanks for opening & in sorry for the delay getting back to you.
In the near term we've decided not to include pg_ivm in the supabase/postgres image

pyrotechnick:
@olirice I'm curious what were there reasons for that decision, and if/when that may be reconsidered?

olirice:
The bar for adding new extensions very high as it introduces complexity/maintenance/support/risk.

For example, the project has to add support for major versions before we can release that major version on the platform. Our support team needs to know how to handle common issues that will occur. Engineering needs to be comfortable enough with the source to monitor releases for breaking changes and resolve issues ourselves if maintainers aren't sufficiently responsive, etc etc

pg_ivm is very cool when it fits your use case, but that use case can be quite [constrained](https://github.com/sraoss/pg_ivm?tab=readme-ov-file#supported-view-definitions-and-restriction).

If the performance of standard views is insufficient, a more general solution would be something like [materialize](https://github.com/MaterializeInc/materialize).

pyrotechnick:
@olirice thanks, appreciate the detailed answer :)

PR #614, opened 2023-04-10
chore: Build Ubuntu 18 binaries

pcnc:
* Lots of Ubuntu 18-specific handling
* Also fixes handling for the `ansible_arguments` Packer var since that never actually worked


migrations/schema.sql
```diff
@@ -135,6 +142,20 @@ CREATE EXTENSION IF NOT EXISTS pgjwt WITH SCHEMA extensions;
 COMMENT ON EXTENSION pgjwt IS 'JSON Web Token API for Postgresql';
 
 
+--
```
darora: schema changes seem unrelated to u18 concerns?

darora (approved):

migrations/schema.sql
pcnc: Yup, those were likely generated by a previous merge commit.
Fixed by merging `develop` into this branch

PR #629, opened 2023-05-02
chore: address possible race conditions encountered when running a PG upgrade

pcnc:
* If project was previously updated, the poll script might return previous upgrade `complete` status
* Kernel might not register newly attached disk's partition table - waiting and running an explicit check & repair forces it to properly register the new volume


supaPatrick (approved):
LGTM!

darora (approved):

PR #643, opened 2023-05-22
feat: build AIO image

pcnc:
* still needs deduplication of some files in `/docker/all-in-one/etc`
  * e.g. fail2ban configs

### DIFF
--- a/docker/all-in-one/Dockerfile
+++ b/docker/all-in-one/Dockerfile
@@ -1,0 +1,234 @@
+ARG postgres_version=15.1.0.83
+
+ARG pgbouncer_release=1.18.0
+ARG postgrest_release=10.1.2
+ARG gotrue_release=2.47.0
+ARG kong_release=2.8.1
+ARG adminapi_release=0.44.3
+ARG adminmgr_release=0.4.0
+ARG vector_release=0.22.3
+ARG postgres_exporter_release=0.9.0
+
+FROM supabase/postgres:${postgres_version} as base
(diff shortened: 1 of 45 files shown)


docker/all-in-one/Dockerfile
```diff
@@ -0,0 +1,233 @@
...
+COPY --from=gotrue /usr/local/etc/gotrue /opt/gotrue/
+COPY --from=adminapi /tmp/supabase-admin-api /opt/
+COPY --chown=root:root --from=adminmgr /tmp/admin-mgr /usr/bin/
+COPY --from=exporter /tmp/postgres_exporter /opt/postgres_exporter/
+COPY docker/all-in-one/opt/postgres_exporter /opt/postgres_exporter/
+
+# Scripts for adminapi
+COPY ansible/files/admin_api_scripts /root 
+COPY --chown=adminapi:adminapi docker/all-in-one/etc/adminapi /etc/adminapi
+COPY docker/all-in-one/etc/sudoers.d /etc/sudoers.d/
```
pcnc, line 153: Not sure about hardcoding the path here - was thinking of pulling the path into a build arg but then you still won't be able to build the image without the files in this repo

darora (approved):

sweatybridge (approved):

docker/all-in-one/Dockerfile
darora: that seems ok? why would you need to be able to build w/o the files from here?

PR #649, opened 2023-05-26
Add postgresql anonymizer extension

olirice:
Adds the PostgreSQL anonymizer extension in support of improving `seed.sql` experience during local development with the supabase CLI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -788,6 +788,25 @@ RUN --mount=type=cache,target=/ccache,from=public.ecr.aws/supabase/postgres:ccac
 RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --nodoc
 
 ####################
+# 30-postgresql-anonymizer.yml
+####################
+FROM ccache as postgresql_anonymizer-source
+ARG postgresql_anonymizer_release
+ARG postgresql_anonymizer_release_checksum
+ADD --checksum=${postgresql_anonymizer_release_checksum} \
+    "https://gitlab.com/dalibo/postgresql_anonymizer/-/archive/${postgresql_anonymizer_release}/postgresql_anonymizer-1.1.0.tar.gz" \
+    /tmp/postgresql_anonymizer.tar.gz
+RUN tar -xvf /tmp/postgresql_anonymizer.tar.gz -C /tmp && \
+    rm -rf /tmp/postgresql_anonymizer.tar.gz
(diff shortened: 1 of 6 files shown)


darora:
@olirice have we confirmed that backups and restores work without any issues?

Dockerfile
```diff
@@ -787,6 +787,25 @@ RUN --mount=type=cache,target=/ccache,from=public.ecr.aws/supabase/postgres:ccac
...
 RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --nodoc
 
+####################
+# 30-postgresql-anonymizer.yml
+####################
+FROM ccache as postgresql_anonymizer-source
+ARG postgresql_anonymizer_release
+ARG postgresql_anonymizer_release_checksum
+ADD --checksum=${postgresql_anonymizer_release_checksum} \
+    "https://gitlab.com/dalibo/postgresql_anonymizer/-/archive/${postgresql_anonymizer_release}/postgresql_anonymizer-1.1.0.tar.gz" \
```
darora: version string hardcoded

ansible/tasks/postgres-extensions/30-postgresql-anonymizer.yml
```diff
@@ -0,0 +1,30 @@
+# postgresql-anonymizer
+- name: postgresql-anonymizer - download latest release
+  get_url:
+    url: "https://gitlab.com/dalibo/postgresql_anonymizer/-/archive/{{ postgresql_anonymizer_release }}/postgresql_anonymizer-1.1.0.tar.gz"
```
darora: version string hardcoded

olirice:
When settings up both the "create extension" statement and an initialization

```
create extension anon;
select anon.init();
```

@soedirgo do you know if we currently have a way to add a callback to `anon.init()` on creation of an extension?

olirice:
> @olirice have we confirmed that backups and restores work without any issues?

@darora I installed it on an instance, dumped and restored without issue. Caveat being that the restored instance did not have the extension available anymore, but it didn't cause any problems

soedirgo:
@olirice just saw this sorry!

supautils only supports custom scripts before or after `create extension`, so you'd need to e.g. create an event trigger in the after-create script to catch the `anon.init()`

olirice:
closing this for now to keep open PRs clean

PR #653, opened 2023-06-01
chore: add suggestions for testing extension changes

darora:


pcnc (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,9 @@
+# Guidelines for upgrading or adding new extensions
+
+One of the common worklow breakages caused by new extensions, or upgraded extensions, is failures in logical restores. A few workflows to verify when making changes to extensions are essentially various combinations straddling the change:
+
+- [ ] backup/restore on a vanilla project
```
egor-romanov, line 5: i am not sure that it is so much of a difference rn, but it can be a bigger thing in the future so i would add explicitly 

```suggestion
- [ ] backup/restore on a vanilla project
- [ ] pause/restore a vanilla project
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,9 @@
...
+- [ ] backup/restore on a project with the extension enabled
+- [ ] backup/restore on a project, after the extension has been enabled and then disabled
+- [ ] backup from an old project (w/o change), restore to a new project (w/ change)
+- [ ] backup from older project (e.g. early 13/14/15 build), restore to a new project (w/ change)
```
egor-romanov, line 10: should we also backup and restore with all/the majority of other extensions enabled as well as the new/updated one?

egor-romanov:
would be great addition to PR template in this repo, also, but i am not sure if youd like it there too

Like this one https://github.com/supabase/.github/blob/main/PULL_REQUEST_TEMPLATE.md

egor-romanov (approved):

CONTRIBUTING.md
darora: If there becomes a difference between the two, we'll need to think about it at that point. Right now there's no difference from a PG perspective.

CONTRIBUTING.md
darora: Have we seen any such failure cases? At the limit you'll end up testing all combinations of extensions, which isn't very feasible.

egor-romanov: i would just suggest all.
yep, i think there were some conflicts when we initially tried adding vault

egor-romanov: all or just top 20, we had csv in slack with usage stats

olirice:

PR #661, opened 2023-06-07
feat: add `auth` schema protections

hf:
Adds an event trigger for DDL commands on the `auth` schema that:

- Prevents commands from running if `auth.enable_schema_ddl_protection()` has been called.
- Increments the sequence `auth.changes_seq` on each applied change.

These can be used by Supabase Auth to:

- Prevent users from accidentally changing the `auth` schema in a way that would break their project. Users can call `auth.disable_schema_ddl_protection()` at any time to make any changes they wish.
- Giving Supabase Auth the ability to detect that the schema has been changed so that when service interruption does occur, the root cause can be identified more easily.

Schema protection is disabled by default with this change, and will be enabled upon further testing with a new PR.

### DIFF
--- a/migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
+++ b/migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
@@ -1,0 +1,115 @@
+-- migrate:up
+
+drop sequence if exists auth.changes_seq;
+
+create sequence
+  auth.changes_seq increment by 1;
+
+comment on sequence auth.changes_seq is 'Tracks the number of changes of the auth schema.';
+
+grant all on sequence
+  auth.changes_seq
+  to supabase_auth_admin
+  with grant option;
+
+drop sequence if exists auth.lock_seq;
+
+create sequence
+  auth.lock_seq increment by -1 no minvalue no maxvalue start with -1;
+  -- starting off with protection disabled
+
+comment on sequence auth.lock_seq is 'Used to protect against accidental schema changes in the auth schema.';
+
+grant all on sequence
+  auth.lock_seq
+  to supabase_auth_admin
+  with grant option;
+
+create or replace function
+  auth.enable_schema_ddl_protection()
+  returns void
+  language sql
+  as $$
+    alter sequence if exists auth.lock_seq increment by 1 no minvalue no maxvalue restart with 1;
+  $$;
+
+grant all on function
+  auth.enable_schema_ddl_protection()
+  to supabase_auth_admin
+  with grant option;
+
+create or replace function
+  auth.disable_schema_ddl_protection()
+  returns void
+  language sql
+  as $$
+    alter sequence if exists auth.lock_seq increment by -1 no minvalue no maxvalue restart with -1;
+  $$;
+
+grant all on function
+  auth.disable_schema_ddl_protection()
+  to supabase_auth_admin
+  with grant option;
+
+create or replace function
+  auth.schema_ddl_protection_command_end()
+  returns event_trigger
+  security definer
+  language plpgsql
+  as $$
+    declare
+      cmd record;
+    begin
+      for cmd in select * from pg_event_trigger_ddl_commands()
+      loop
+        if cmd.schema_name = 'auth' then
+          if nextval('auth.lock_seq') > 0 then
+            raise exception 'auth schema is protected from unintended changes. To unlock run SELECT auth.disable_schema_ddl_protection(); ';
+          end if;
+
+          perform nextval('auth.changes_seq');
+        end if;
+      end loop;
+    end;
+  $$;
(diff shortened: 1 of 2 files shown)


migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -0,0 +1,84 @@
...
+
+grant all on sequence
+  auth.changes_seq
+  to supabase_auth_admin
+  with grant option;
+
+grant usage, select on sequence
+  auth.changes_seq
+  to supabase_admin;
+  -- supabase_admin can advance the sequence only
```
olirice: as a superuser, supabase_admin will have full access to the seq

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -0,0 +1,84 @@
...
+
+create sequence if not exists
```
olirice: nit:
since this sequence should never exist prior to this migration, remove `if not exists`. If something has gone wrong and the sequence does exist, its better to fail loudly during a migration than to hide the issue

olirice: same for `changes_seq`

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -0,0 +1,84 @@
...
+  auth.lock_seq increment by -1;
```
olirice: setting the increment to `-1` will cause the min sequence value to be `-9223.....` and a max sequence value to `-1`

That causes 
```
select auth.enable_schema_ddl_protection();
```
to raise
```
ERROR:  RESTART value (1) cannot be greater than MAXVALUE (-1)
CONTEXT:  SQL function "enable_schema_ddl_protection" statement 1 

SQL state: 22023
```

to avoid that error, manually specify a large positive maximum when creating the seq

```sql
create seq auth.lock_seq minvalue -999999 maxvalue 999999  increment by -1;
```

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -0,0 +1,84 @@
...
+        end if;
+      end loop;
+    end;
+  $$;
+
+drop event trigger if exists protect_auth_schema_ddl cascade;
+
+create event trigger
+  protect_auth_schema_ddl
+  on ddl_command_start
```
olirice: `ddl_command_start` is not compatible with `pg_event_trigger_ddl_commands()`

currently that function will always return `null`, if you change it to `ddl_command_end` it will work as you expect

olirice (changes requested):
Weirdly, event triggers have a separate category for drops so drop statements wouldn't be detected with this setup.

Here's the full matrix of what `ddl_command_begin`, `ddl_command_end` and `sql_drop` detect https://www.postgresql.org/docs/current/event-trigger-matrix.html

to handle drops you'll want to add another event trigger on `sql_drop` that refers to `pg_event_trigger_dropped_objects()` rather than `pg_event_trigger_ddl_commands()`


[full example](https://github.com/supabase/pg_graphql/blob/34cc266da972d356365f953503deb7a39960da0a/pg_graphql--0.4.0.sql#L4906-L4986)

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
hf: Ok, then this is not necessary. Will remove it.

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
soedirgo: These migrations are also run when doing restores, so we need `if not exists` to make them idempotent

hf:
@olirice I addressed some of your comments, however I don't think `sql_drop` is necessary as `ddl_command_end` seems to support `DROP` statements too.

As for `if not exists` looks like the tests here keep a state of the DB, so I have to use them to get a clean schema.

olirice:
>  I don't think sql_drop is necessary as ddl_command_end seems to support DROP statements too

sorry, I forgot the specifics of the issue. The event trigger does fire, but `pg_event_trigger_ddl_commands()` will be empty so it won't work for your purpose without another event trigger for `sql_drop` and `pg_event_trigger_dropped_objects()`. 

I double checked and confirmed that the existing setup isn't protecting from e.g. dropping a column on an auth schema table.

you can read more about it in [this thread](https://github.com/PostgREST/postgrest-docs/issues/445#issuecomment-1010269737)

darora:

PR #665, opened 2023-06-13
chore: fly fixes; PITR + logging + shutdown + perms

pcnc:

## What is the new behavior?

## Additional context

### DIFF
--- a/docker/all-in-one/Dockerfile
+++ b/docker/all-in-one/Dockerfile
@@ -231,5 +231,5 @@ HEALTHCHECK --interval=3s --timeout=2s --start-period=4s --retries=10 CMD [ "hea
 COPY docker/all-in-one/init /init
 COPY docker/all-in-one/entrypoint.sh /usr/local/bin/
 COPY docker/all-in-one/postgres-entrypoint.sh /usr/local/bin/
-COPY docker/all-in-one/shutdown.sh /usr/local/bin/supa-shutdown
+COPY docker/all-in-one/shutdown.sh /usr/local/bin/supa-shutdown.sh
 ENTRYPOINT [ "entrypoint.sh" ]
--- a/docker/all-in-one/Dockerfile
+++ b/docker/all-in-one/Dockerfile
@@ -5,7 +5,7 @@ ARG postgrest_release=10.1.2
 ARG gotrue_release=2.47.0
 ARG kong_release=2.8.1
 ARG adminapi_release=0.44.3
-ARG adminmgr_release=0.4.0
+ARG adminmgr_release=0.5.0
 ARG vector_release=0.22.3
 ARG postgres_exporter_release=0.9.0
 
@@ -230,4 +230,6 @@ HEALTHCHECK --interval=3s --timeout=2s --start-period=4s --retries=10 CMD [ "hea
 
 COPY docker/all-in-one/init /init
 COPY docker/all-in-one/entrypoint.sh /usr/local/bin/
+COPY docker/all-in-one/postgres-entrypoint.sh /usr/local/bin/
+COPY docker/all-in-one/shutdown.sh /usr/local/bin/supa-shutdown
 ENTRYPOINT [ "entrypoint.sh" ]
--- a/docker/all-in-one/entrypoint.sh
+++ b/docker/all-in-one/entrypoint.sh
@@ -125,7 +125,6 @@ fi
 
 if [ "${PGDATA:-}" ]; then
   if [ "${PGDATA_REAL:-}" ]; then
-    mkdir -p "$(dirname "${PGDATA}")"
     ln -s "${PGDATA}" "${PGDATA_REAL}"
   else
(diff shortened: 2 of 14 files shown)


docker/all-in-one/Dockerfile
```diff
@@ -230,4 +230,6 @@ HEALTHCHECK --interval=3s --timeout=2s --start-period=4s --retries=10 CMD [ "hea
 
 COPY docker/all-in-one/init /init
 COPY docker/all-in-one/entrypoint.sh /usr/local/bin/
+COPY docker/all-in-one/postgres-entrypoint.sh /usr/local/bin/
+COPY docker/all-in-one/shutdown.sh /usr/local/bin/supa-shutdown
```
darora: any reason not to preserve the `.sh` ext?

docker/all-in-one/shutdown.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
+
```
darora, line 2: Let's give a brief overview of what the script does

docker/all-in-one/shutdown.sh
```diff
@@ -0,0 +1,60 @@
...
+  if [ "$ACTIVE_CONN_COUNT" = "0" ] && [ $TIME_SINCE_LAST_DISCONNECT -gt "$((MAX_IDLE_TIME_MINUTES * 60))" ]; then
+    LAST_WAL_FILE_NAME=$(run_sql -tA -c "SELECT pg_walfile_name(pg_switch_wal())")
+    NEW_WAL_FILE_NAME=$(run_sql -tA -c "SELECT pg_walfile_name(pg_current_wal_lsn())")
+
+    echo "$(date): No active connections for $MAX_IDLE_TIME_MINUTES minutes. Shutting down."
+
+    supervisorctl stop postgresql
+
+    # Postgres ships the latest WAL file using archive_commAND during shutdown, in a blocking operation
```
darora:
```suggestion
    # Postgres ships the latest WAL file using archive_command during shutdown, in a blocking operation
```

docker/all-in-one/etc/postgresql/logging.conf
```diff
@@ -31,3 +31,5 @@ log_rotation_size = 0		# Automatic rotation of logfiles will
 					# or size-driven rotation.  Default is
 					# off, meaning append to existing files
 					# in all cases.
+
+log_disconnections = on			# log end of a session, including duration
```
darora: intentional?

docker/all-in-one/entrypoint.sh
```diff
@@ -138,10 +175,28 @@ fi
 
 # Download and extract init payload from s3
 export INIT_PAYLOAD_PATH=${INIT_PAYLOAD_PATH:-/tmp/payload.tar.gz}
-export CONFIGURED_FLAG_PATH=${CONFIGURED_FLAG_PATH:-$PGDATA/../machine.configured}
 
 if [ "${INIT_PAYLOAD_PRESIGNED_URL:-}" ]; then
-  curl -sSL "$INIT_PAYLOAD_PRESIGNED_URL" -o "$INIT_PAYLOAD_PATH"
+  curl -sSL "$INIT_PAYLOAD_PRESIGNED_URL" -o "/tmp/payload.tar.gz"
```
darora: `-f` as well?

docker/all-in-one/entrypoint.sh
```diff
@@ -138,10 +175,28 @@ fi
...
+  mv "/tmp/payload.tar.gz" "$INIT_PAYLOAD_PATH"
+fi
+
+if [ "${DATA_VOLUME_MOUNTPOINT}" ]; then
+  LOGS_FOLDER="${DATA_VOLUME_MOUNTPOINT}/logs"
+
+  mkdir -p "${LOGS_FOLDER}/postgresql"
```
darora: could be a loop over postgresql, services, wal-g, but this is fine too

darora (approved):

docker/all-in-one/etc/postgresql/logging.conf
pcnc: Yes. 
This param is disabled by default and can't track disconnections any other way since the connections are terminated and thus not available in `pg_stat_activity`.
This allows the shutdown script to track when the last disconnection happened and shutdown the machine after X minutes.

Postgres output related to disconnections is filtered out by Vector, so it won't additionally be shipped to Logflare.

This approach diverges the configs, thus we could have the shutdown script run an `ALTER SYSTEM` at startup to set `log_disconnections` to on, since it's a shutdown script dependency after all.

docker/all-in-one/Dockerfile
pcnc: no reason - will add extension

docker/all-in-one/entrypoint.sh
pcnc: fair point - refactored into a loop

PR #667, opened 2023-06-13
Auth protection tweaks

olirice:
Updates auth schema change protection resolve edge cases around sequence bounds and idempotence 

Behavior 1
```
begin;
	 set local role postgres;

        -- raises exception
	create function auth.foo() returns int language sql as $$ select 1 $$;
rollback;
```

Behavior 2
```
begin;
	 set local role postgres;

	select auth.disable_schema_ddl_protection()

	create function auth.foo() returns int language sql as $$ select 1 $$;

        -- re-enable protection is allowed but change_seq has been incremented		
 	select auth.enable_schema_ddl_protection()
rollback;
```

### DIFF
--- a/migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
+++ b/migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
@@ -1,5 +1,8 @@
 -- migrate:up
 
+drop event trigger if exists protect_auth_schema_ddl_command_end;
+drop event trigger if exists protect_auth_schema_ddl_sql_drop;
+
 drop sequence if exists auth.changes_seq;
 
 create sequence
@@ -14,8 +17,7 @@ grant all on sequence
 
 drop sequence if exists auth.lock_seq;
 
-create sequence
-  auth.lock_seq increment by -1 no minvalue no maxvalue start with -1;
+create sequence auth.lock_seq increment by -1 minvalue -999999999999999 maxvalue 999999999999999 start with -1;
   -- starting off with protection disabled
 
 comment on sequence auth.lock_seq is 'Used to protect against accidental schema changes in the auth schema.';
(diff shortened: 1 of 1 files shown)


migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -98,18 +111,17 @@ create or replace function
...
 
-drop event trigger if exists protect_auth_schema_ddl_sql_drop cascade;
-
 create event trigger
   protect_auth_schema_ddl_sql_drop
   on sql_drop
   execute function auth.schema_ddl_protection_drop();
 
+-- Enable the lock
+select auth.enable_schema_ddl_protection()
```
olirice, line 125: note, security enabled after migration runs

hf: Let's keep it off for now, we can roll it out individually on projects as a first stage, then once confident it's a good idea we can enable it in the migrations.

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -87,7 +100,7 @@ create or replace function
     begin
       for cmd in select * from pg_event_trigger_dropped_objects()
       loop
-        if cmd.schema_name = 'auth' and cmd.object_identity not in ('auth.changes_seq', 'auth.lock_seq') then
+        if cmd.schema_name = 'auth' then
```
hf, line 103: Actually this is somewhat intentional as it appears that when you do a DROP the sequence/object is no longer visible when this event trigger runs. Essentially if we don't have that, we can never ever drop the sequences which is not a good idea.

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
```diff
@@ -14,8 +17,7 @@ grant all on sequence
 
 drop sequence if exists auth.lock_seq;
 
-create sequence
-  auth.lock_seq increment by -1 no minvalue no maxvalue start with -1;
+create sequence auth.lock_seq increment by -1 minvalue -999999999999999 maxvalue 999999999999999 start with -1;
```
hf, line 20: Let's go with this.

hf (approved):

migrations/db/migrations/20230607094822_add_auth_schema_protections.sql
olirice: the concern is that dropping the sequences doesn't cascade to the event triggers which causes the event triggers to raise an error and prevent all migrations

PR #674, opened 2023-06-21
chore: bump postgrest to v11.1.0

dragarcia:


steve-chavez (approved):

delgado3d (approved):

steve-chavez:
So looks like `v11.1` is not apt for Supabase cloud bc of https://github.com/PostgREST/postgrest/pull/2834.

I'll make a new `v11.2` next which should be fine to deploy.

steve-chavez:
> So looks like v11.1 is not apt for Supabase cloud bc of https://github.com/PostgREST/postgrest/pull/2834.

Correction, the above is not true. It's just a small percentage of Supabase instances (old ones) that are on Ubuntu 18.04 and most are on Ubuntu 20.04. So we can proceed with the deployment of `v11.1`.

> I'll make a new v11.2 next which should be fine to deploy.

I'll not rush a new version for PostgREST. But a next version will support the old Supabase instances.

darora (approved):

PR #690, opened 2023-07-05
chore: fix and enable pljava

pcnc:
* Enables pljava
* Some project upgrades were failing since the extension was missing in the upgrade binaries (and in the distribution itself), and databases targeted for upgrade had pljava enabled


migrations/tests/extensions/12-pljava.sql
```diff
@@ -1,3 +1,3 @@
 BEGIN;
-create extension if not exists pljava with schema "sqlj";
```
pcnc, line 2: pljava creates the sqlj schema automatically

.github/workflows/ami-build-ubuntu-18.yml
```diff
@@ -80,11 +82,12 @@ jobs:
             ubuntu_release_no=${{ matrix.ubuntu_version }}
             postgresql_major=${{ steps.version.outputs.postgresql_major }}
             postgresql_release=${{ steps.version.outputs.postgresql_release }}
-            CPPFLAGS=-mcpu=${{ matrix.mcpu }}
+            CPPFLAGS=
+            DEB_BUILD_PROFILES=pkg.postgresql.nozstd
```
pcnc, line 86: Building Postgres from source requires libzstd 1.4.0+, whereas Ubuntu 18 only has versions up to 1.3.3 available.
Sadness all around, although it doesn't affect us since these binaries are only used to upgrade from PG13/14 (which didn't have zstd compression support)

.github/workflows/ami-build-ubuntu-18.yml
```diff
@@ -46,6 +46,8 @@ jobs:
       - uses: docker/setup-buildx-action@v2
         with:
           endpoint: builders          
+          driver-opts: image=moby/buildkit:v0.11.6
```
pcnc, line 49: Updated buildkit just in case it'd help us avoid connection reset errors; default was running under v0.10.0

.github/workflows/ami-build-ubuntu-18.yml
```diff
@@ -56,7 +58,7 @@ jobs:
           tags: supabase/postgres:extensions
           platforms: linux/${{ matrix.arch }}
           cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-to: type=gha,mode=max,ignore-error=true
```
pcnc, line 61: Also no point crashing the entire workflow if the build cache push fails

Dockerfile
```diff
@@ -403,11 +403,10 @@ RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --pkgver
...
 FROM builder as pljava-source
 # Download and extract
-# TODO: revert to using main repo after PG15 support is merged: https://github.com/tada/pljava/pull/413
-ARG pljava_release=master
-ARG pljava_release_checksum=sha256:e99b1c52f7b57f64c8986fe6ea4a6cc09d78e779c1643db060d0ac66c93be8b6
+ARG pljava_release=1_6_5
+ARG pljava_release_checksum=sha256:e1492b237eac8aaed6792feb26b8fa06b9c6ec3bd0a61ecd7a90e175dd1f1c5a
 ADD --checksum=${pljava_release_checksum} \
-    "https://github.com/supabase/pljava/archive/refs/heads/${pljava_release}.tar.gz" \
+    "https://github.com/tada/pljava/archive/refs/tags/V${pljava_release}.tar.gz" \
```
pcnc, line 409: Reverting to original repo since PG15 support was merged in the meantime

darora (approved):
We've tested pause/restore and upgrades from older builds that had pljava enabled?

Dockerfile
```diff
@@ -422,15 +421,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-RUN cp pljava-packaging/target/pljava-pg${postgresql_major}.jar /tmp/
-
-FROM base as pljava
-# Download pre-built packages
-RUN apt-get update && apt-get install -y --no-install-recommends --download-only \
-    default-jdk-headless \
-    postgresql-${postgresql_major}-pljava \
-    && rm -rf /var/lib/apt/lists/*
-RUN mv /var/cache/apt/archives/*.deb /tmp/
+RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --pkgname=pljava --pkgversion="\${pljava_release//_/.}" --requires=libssl-dev,default-jdk --exclude /proc/ --nodoc java -jar pljava-packaging/target/pljava-pg${postgresql_major}.jar
```
sweatybridge:
```suggestion
RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --pkgname=pljava --pkgversion="\${pljava_release//_/.}" --requires=default-jdk-headless --exclude /proc/ --nodoc java -jar pljava-packaging/target/pljava-pg${postgresql_major}.jar
```

Do you need libssl-dev at runtime? It should be a build dependency.

Also prefers the headless jdk since it's a smaller installation.

pcnc: Didn't check - assumed it was needed since it was not removed [here](https://github.com/supabase/postgres/pull/690/files#diff-f9985ee5a6ca94a95ad0bcfc670ebf7d1114bce1003675c4383068c1e3bf2eb8R41-R43).

Might've also been needed by some other extension further down the line and we didn't properly contain extension deps, so definitely worth a check if it's needed or not.

pcnc:
> We've tested pause/restore and upgrades from older builds that had pljava enabled?

Not yet - this is pending today - upgrades breaking were the main driver for this PR

PR #695, opened 2023-07-11
feat: automate pg_upgrade testing

pcnc:
* Removes the `DRY_RUN` flag and introduces an `IS_CI` flag
* Runs the pg_upgrade dry-run command (`--check`) before every upgrade, should make it easier to spot if any prerequisites are are failing, before the migration actually starts
* Splits the pg_upgrade script publishing workflow in separate staging/prod deploy jobs to make testing easier, less error prone
  * can be deployed from any branch to staging via workflow dispatch

To FLUP:
* tests targeting encrypted columns
* tests targeting problematic extensions (ltree, others)


darora (approved):
Sorry I missed this earlier - gave it a quick skim and looks great. Any changes needed to bring it up to speed?

@supabase/postgres any concerns with this change?

olirice:
lgtm, @soedirgo could you give this a once over?

soedirgo:
Looks great! 👍I guess we don't need to include [Testinfra](https://github.com/supabase/postgres/blob/32e93b510517ecb87c77a6cbbddf358acf3f0e10/.github/workflows/testinfra.yml#L8C19-L8C19) tests since this should only affect PG stuff?

steve-chavez:
Is it possible to run these tests locally?

pcnc:
> Sorry I missed this earlier - gave it a quick skim and looks great. Any changes needed to bring it up to speed?
> 
> @supabase/postgres any concerns with this change?

Need to follow up on this and merge the more recent upgrade script changes into it.
Also want to add tests around encrypted columns (since those are more critical), some extensions such as ltree now that it is possibly problematic, and have pick it up the last ~5 PG releases to cover, besides the configured baseline

@steve-chavez running `debug.sh` would run this locally, but it requires the scripts and binaries stored in the dev account. I’ll add some docs around this and ping the team in Slack to confirm that everybody can run it

pcnc:
> Looks great! 👍I guess we don't need to include [Testinfra](https://github.com/supabase/postgres/blob/32e93b510517ecb87c77a6cbbddf358acf3f0e10/.github/workflows/testinfra.yml#L8C19-L8C19) tests since this should only affect PG stuff?

Yep! This would run as a parallel check to those, so no dependency between the two workflows

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -104,11 +112,16 @@ function complete_pg_upgrade {
     sleep 5
 
     echo "5. Restarting postgresql"
-    retry 3 service postgresql restart
-
-    echo "5.1. Restarting gotrue and postgrest"
-    retry 3 service gotrue restart
-    retry 3 service postgrest restart
+    if [ -z "$IS_CI" ]; then
+        service postgresql restart
```
pcnc:
```suggestion
        retry 3 service postgresql restart
```

darora (approved):

pcnc:
Will merge after tomorrow's (Jun 13th) prod deploy, to have more ample time for testing the pg_upgrade scripts

PR #697, opened 2023-07-14
chore: fix for sar making sure directory is available before service …

supaPatrick:
changing when the directory for sysstat gets created to insure it exists before the service starts

Bug fix: should ensure sysstat logs are in the correct locations

### DIFF
--- a/ansible/tasks/setup-system.yml
+++ b/ansible/tasks/setup-system.yml
@@ -32,6 +32,11 @@
     update-alternatives --set ebtables /usr/sbin/ebtables-nft
     systemctl restart ufw
 
+- name: Create Sysstat log directory
+  file:
+    path: /var/log/sysstat
+    state: directory
+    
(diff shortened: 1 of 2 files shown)


ansible/tasks/setup-system.yml
```diff
@@ -32,6 +32,11 @@
     update-alternatives --set ebtables /usr/sbin/ebtables-nft
     systemctl restart ufw
 
+- name: Create Sysstat log directory
+  file:
+    path: /var/log/sysstat
+    state: directory
```
darora, line 38: I don't think it's the placement of this clause, but the cleanup performed afterwards by `scripts/91-log_cleanup.sh` that's tripping it up.

supaPatrick: agree updated to create it after its getting nuked.

darora (approved):

PR #703, opened 2023-07-18
chore: Adding Envoy to the all-in-one image, not running be default

supaPatrick:
This is adding envoy as service to the all-in-one docker image. Envoy is is not enabled by default and should not effect/change any functionality with the container at the moment but will make it easier to change between them when testing envoy at fly.

## What is the new behavior?

## Additional context

### DIFF
--- a/docker/all-in-one/init/configure-envoy.sh
+++ b/docker/all-in-one/init/configure-envoy.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -eou pipefail
+
+ENVOY_CONF=/etc/envoy/envoy.yml
+touch /var/log/services/envoy.log
+
+
+
+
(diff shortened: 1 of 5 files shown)


docker/all-in-one/init/configure-envoy.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+# Inject project specific configuration
+sed -i -e "s|anon_key|$ANON_KEY|g" \
+  -e "s|service_key|$SERVICE_ROLE_KEY|g" \
+  -e "s|supabase_admin_key|$ADMIN_API_KEY|g" \
+  $ENVOY_CONF
+
+# Update envoy ports
+sed -i "s|80 |:$KONG_HTTP_PORT |g" /etc/envoy/envoy.yml
```
darora: stray colons?

darora (approved):

PR #707, opened 2023-07-21
fix: ensure that the test slot gets dropped

darora:
Based on a quick test, a rollback of the txn does not remove the
logical replication slot that gets created.

### DIFF
--- a/migrations/tests/extensions/11-wal2json.sql
+++ b/migrations/tests/extensions/11-wal2json.sql
@@ -1,4 +1,5 @@
 BEGIN;
 select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name = 'test_slot';
 select * from pg_create_logical_replication_slot('test_slot', 'wal2json');
+select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name = 'test_slot';
 ROLLBACK;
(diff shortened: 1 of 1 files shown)


dragarcia (approved):

darora:
Red build is a cache failure

migrations/tests/extensions/11-wal2json.sql
```diff
@@ -1,4 +1,5 @@
 BEGIN;
 select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name = 'test_slot';
 select * from pg_create_logical_replication_slot('test_slot', 'wal2json');
+select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name = 'test_slot';
```
soedirgo, line 5:
```suggestion
-- a rollback of the txn does not remove the logical replication slot that gets created, so we need to manually drop it
select pg_drop_replication_slot(slot_name) from pg_replication_slots where slot_name = 'test_slot';
```

olirice:
Was `test_slot` scanned for and dropped on instances launched in the impacted window?

A support ticket was opened today where `pg_statistic` was growing in bloat and `vacuum full` would not resolve it due to reference from an inactive `test_slot`. It was only a few hundred MB but thats an issue with free tier usage limits

darora:
Has not been rolled out yet. Probably some work that needs to be done to determine the slot isn't being used by something (even though it was created inadvertently by us).

olirice:
Tickets relating to this issue are still trickling coming in and the size of pg_statistic is getting larger and larger (4.6 GB today)

> determine the slot isn't being used by something (even though it was created inadvertently by us).

Using slot status = `lost` as an acceptance criterion for deleting should work (at least it would for every case I've seen so far)

PR #708, opened 2023-07-21
chore: preserve binaries across reboots

pcnc:

## What is the new behavior?

## Additional context


docker/all-in-one/init/configure-admin-mgr.sh
```diff
@@ -0,0 +1,20 @@
...
+set -eou pipefail
+
+ADMINMGR_CUSTOM_DIR="${DATA_VOLUME_MOUNTPOINT}/admin-mgr"
+
+if [ "${DATA_VOLUME_MOUNTPOINT}" ]; then
+  mkdir -p "${ADMINMGR_CUSTOM_DIR}"
+
+  if [ ! -f "${ADMINMGR_CUSTOM_DIR}/admin-mgr" ]; then
+    echo "Copying existing admin-mgr binary from /usr/bin/admin-mgr to ${ADMINMGR_CUSTOM_DIR}"
+    cp "/usr/bin/admin-mgr" "${ADMINMGR_CUSTOM_DIR}/admin-mgr"
```
darora: Not sure what the ideal place is since we're doing this for each of the services, but the general flow of operations here could use a comment describing it

docker/all-in-one/init/configure-admin-mgr.sh
```diff
@@ -0,0 +1,20 @@
...
+  fi
+
+  rm -f "/usr/bin/admin-mgr"
```
darora: All this has to be executed each time the machine starts up?

What if we shipped a dangling symlink as part of the image, and the init script moved the base version of the binary to the data vol as part of the very first boot? Maybe that's complicating things unnecessarily though

darora (approved):

docker/all-in-one/init/configure-admin-mgr.sh
pcnc: I think that'd be better - copying over a few config files doesn't bring much overhead, but copying over tens of MB on every boot can bring a noticeable overhead, especially if we aim for our boot times to generate minimal friction

docker/all-in-one/init/configure-admin-mgr.sh
pcnc: Though about pulling this out in a generic function - but its maintainability might take a hit given the number of args and options per service. Will take a different approach, so that we also save on boot time overhead

PR #726, opened 2023-08-24
pre-install dbdev client

olirice:
pre-installs the [dbdev](https://database.dev) client so users can immediately run 

```sql
select dbdev.install('some-package')
```

rather than first having to run the scary looking [installer script](https://database.dev/installer)

### DIFF
--- a/migrations/db/migrations/20230824080821_dbdev_installer.sql
+++ b/migrations/db/migrations/20230824080821_dbdev_installer.sql
@@ -2,33 +2,203 @@
 
 create extension if not exists http with schema extensions;
 create extension if not exists pg_tle;
+
 select
     pgtle.install_extension(
         'supabase-dbdev',
-        resp.contents ->> 'version',
-        'PostgreSQL package manager',
-        resp.contents ->> 'sql'
-    )
-from http(
-    (
-        'GET',
-        'https://api.database.dev/rest/v1/'
-        || 'package_versions?select=sql,version'
-        || '&package_name=eq.supabase-dbdev'
-        || '&order=version.desc'
-        || '&limit=1',
-        array[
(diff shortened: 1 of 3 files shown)


kiwicopple (approved):

sweatybridge (approved):
Nice! Let me know if the test failures are difficult to fix.

olirice:
this is ready to go but I've been unable to get an image spun up on staging (this one or others). going to wait for that to work so I can manually test loading a package from dbdev

migrations/db/migrations/20230824080821_dbdev_installer.sql
```diff
@@ -0,0 +1,211 @@
...
+        create schema dbdev;
+
+        create or replace function dbdev.install(package_name text)
+            returns bool
+            language plpgsql
+        as $$
+        declare
+            -- Endpoint
+            base_url text = 'https://api.database.dev/rest/v1/';
+            apikey text = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6InhtdXB0cHBsZnZpaWZyYndtbXR2Iiwicm9sZSI6ImFub24iLCJpYXQiOjE2ODAxMDczNzIsImV4cCI6MTk5NTY4MzM3Mn0.z2CN0mvO2No8wSi46Gw59DFGCTJrzM0AQKsu_5k134s';
```
darora, line 26: This is from a supabase project? And is present in all clients using dbdev?

What's the path forward for if the jwt secret needs to be rolled due to a leak?

olirice: It's intentional. We'd prefer not to use an apikey at all but afaik that's still not possible.

Good point about if the jwt secret needs to be rolled. I'll bring it up

This PR is not going to be merged b/c it requires adding http and pgtle as default on on new projects so we opted for a template instead for now

closing

PR #730, opened 2023-08-30
Adds pg archivecleanup

olirice:
### What
Adds pg_archivecleanup command to delete unused WAL on each checkpoint.

Currently it also logs to a file every time the command executes so we can investigate how frequently its running but that will be removed before this (maybe) merges.

### Why
 users frequently try to reclaim disk space by deleting or truncating tables to stay under supabase disk usage limits. They are surprised to find that up to 4 GB of disk usage is potentially consumed by WAL files. Those files are ultimately recycled, but they can't be removed, so the dbsize does not appear to be under user control. pg_archivecleanup is a utility that deletes WAL files no longer referenced (including by replication slots).


olirice:
not needed after billing changes to remove WAL from billed space

PR #732, opened 2023-09-04
fix: inform PG after log rotation

darora:


pcnc (approved):

steve-chavez:
@darora Can we include a description of the change here? What's the problem and why is this the solution?

This will help any possible integration testing we might do on [nix-postgres](https://github.com/supabase/nix-postgres)

darora:
Additional details can be found at https://www.notion.so/supabase/logrotate-broken-as-pg_ctl-not-on-path-b19a114598b141f2a84dd450da8e0dd8?pvs=4

PR #735, opened 2023-09-17
chore(aio): add simple testinfra test for postgrest

soedirgo:

chore

## What is the new behavior?

Add initial setup for testinfra integration tests.

### DIFF
--- a/docker/all-in-one/healthcheck.sh
+++ b/docker/all-in-one/healthcheck.sh
@@ -38,6 +38,7 @@ fail2ban-client status
 curl -sSfI "http://localhost:$PGEXPORTER_PORT/metrics"
 
 # vector is up (if starting logflare)
+# TODO: make this non-conditional once we set up local logflare for testinfra
 if [ -n "${LOGFLARE_API_KEY:-}" ]; then
   curl -sSfI "http://localhost:$VECTOR_API_PORT/health"
 fi
--- a/docker/all-in-one/healthcheck.sh
+++ b/docker/all-in-one/healthcheck.sh
@@ -37,5 +37,7 @@ fail2ban-client status
 # prometheus exporter up
 curl -sSfI "http://localhost:$PGEXPORTER_PORT/metrics"
 
-# vector is up
-curl -sSfI "http://localhost:$VECTOR_API_PORT/health"
+# vector is up (if starting logflare)
+if [ -n "${LOGFLARE_API_KEY:-}" ]; then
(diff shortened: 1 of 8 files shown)


docker/all-in-one/healthcheck.sh
```diff
@@ -37,5 +37,7 @@ fail2ban-client status
 # prometheus exporter up
 curl -sSfI "http://localhost:$PGEXPORTER_PORT/metrics"
 
-# vector is up
-curl -sSfI "http://localhost:$VECTOR_API_PORT/health"
+# vector is up (if starting logflare)
+if [ -n "${LOGFLARE_API_KEY:-}" ]; then
+  curl -sSfI "http://localhost:$VECTOR_API_PORT/health"
+fi
```
soedirgo: We don't currently spin up a local Logflare for Vector tests, but we still need the healthcheck to pass before we run testinfra tests.

docker/all-in-one/init/configure-gotrue.sh
```diff
@@ -5,6 +5,7 @@ touch /var/log/services/gotrue.log
 
 /usr/local/bin/configure-shim.sh /dist/gotrue /opt/gotrue/gotrue
 
+sed -i "s|api_external_url|${API_EXTERNAL_URL:-http://localhost}|g" /etc/gotrue.env
```
soedirgo, line 8: Seems like this was recently made to be mandatory; this fallback URL won't work, but not a problem until we add a test that uses it.

(on hosted platform this would be `https://ref.supabase.co/auth/v1`)

pcnc:
🔥

pcnc (approved):

testinfra/test_all_in_one.py
```diff
@@ -0,0 +1,79 @@
...
+        sleep(1)
+
+    # return a testinfra connection to the container
+    yield testinfra.get_host("docker://" + cast(str, container.name))
+
+    # at the end of the test suite, destroy the container
+    container.remove(v=True, force=True)
+
+
+def test_postgrest_service(host):
```
darora: Should probably go a step further and ensure it's able to respond to requests, rather than e.g. catching it in a "running" state just at the right time of a crashloop

docker/all-in-one/healthcheck.sh
darora: Should add a TODO to remove this guard once logflare is spun up; adding this guard here makes the setup more fragile

darora (approved):

PR #746, opened 2023-10-04
chore(testinfra): initial ami test

soedirgo:
Initial Testinfra integration tests for AMI builds.

TODO:
- [ ] allow GHA AWS role to run instances
- [x] add steps for locally running tests

### DIFF
--- a/.github/workflows/testinfra.yml
+++ b/.github/workflows/testinfra.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  test-all-in-one:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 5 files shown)


testinfra/test_ami.py
```diff
@@ -0,0 +1,307 @@
...
+import boto3
+import gzip
+import pytest
+import requests
+import testinfra
+from ec2instanceconnectcli.EC2InstanceConnectLogger import EC2InstanceConnectLogger
+from ec2instanceconnectcli.EC2InstanceConnectKey import EC2InstanceConnectKey
+from time import sleep
+
+postgresql_schema_sql_content = """
```
soedirgo, line 11: We need to hardcode a lot of these since the setup logic lives in worker. This makes the tests brittle. One possible improvement would be to move some of the setup logic into AMI creation.

darora: The bulk of the content here is dependent on project specific data that'll only be available in the worker, no?

The improvements to be made here are in making the interface between the AMI and the worker more explicit, and testing on both sides (that the worker produces the right interface, and that the AMI here takes that interface and does the right things with it).

darora: To start with, we can add a note in the relevant spots on both sides that any change to one (worker config generation/postgres integration test) _must_ be captured by an update and ideally a test in the other

testinfra/test_ami.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+# scope='session' uses the same container for all the tests;
+# scope='function' uses a new container per test function.
+@pytest.fixture(scope="session")
+def host():
+    ec2 = boto3.resource("ec2", region_name="ap-southeast-1")
+    images = list(
+        ec2.images.filter(
+            Filters=[{"Name": "name", "Values": ["supabase-postgres-ci-ami-test"]}]
```
darora, line 157: We should be using the version string from the common.vars file here, yeah? Not sure which AMI this is currently set up to use?

darora (approved):
Looks like a good start! Are you planning to add the rest of the testing scenarios as well? Or would you want someone on infra to take over and spec the rest out?

testinfra/test_ami.py
soedirgo: The test run will always publish an AMI with AMI name `supabase-postgres-ci-ami-test`. Subsequent runs will deregister older AMIs through the `force_deregister` option.

One alternative I haven't explored is to move the test run into a Packer [provisioning step](http://celien.info/posts/ami_packer_serverspec.html), that way we might not even need to publish the AMI. Can be done as a follow up.

testinfra/test_ami.py
soedirgo: Yeah, what I mean is to move the file templating to supabase/postgres, so that UserData only contains things that need to be variable. I'll start with adding a note for now.

soedirgo:
I think it's better for the infra team to take over, since you have a better understanding of the requirements.

.github/workflows/testinfra.yml
```diff
@@ -25,5 +25,110 @@ jobs:
...
+      - name: Run tests
+        run: |
+          # TODO: use poetry for pkg mgmt
+          pip3 install boto3 boto3-stubs[essential] docker ec2instanceconnectcli pytest pytest-testinfra[paramiko,docker] requests
+          pytest -vv testinfra/test_ami.py
+
+      - name: Cleanup resources on build cancellation
+        if: ${{ cancelled() }}
+        run: |
+          aws ec2 --region ap-southeast-1 describe-instances --filters "Name=tag:packerExecutionId,Values=${GITHUB_RUN_ID}" --query "Reservations[].Instances[].InstanceId" --output text | xargs -I {} aws ec2 terminate-instances --instance-ids {}
```
olirice, line 134: since the setup to this point is the same as `ami-release` could we add these lines to that file instead & run it `on PR` to reduce duplication?

soedirgo:
Merging this for now - will dedupe some of the Action steps in another PR

PR #759, opened 2023-10-16
Chore: adding sysctl params for rebooting the OS after oomkiller kill…

supaPatrick:
adding sysctl params to restart the system on OOM.
tested the machine does restart on OOM.

### DIFF
--- a/ansible/tasks/setup-system.yml
+++ b/ansible/tasks/setup-system.yml
@@ -131,3 +131,18 @@
     mode: 0644
     owner: root
     group: root
+
+#Set Sysctl params for restarting the OS on oom after 10
+- name: Set vm.panic_on_oom=1
+  ansible.builtin.sysctl:
+    name: vm.panic_on_oom
+    value: '1'
+    state: present
+    reload: yes
+
+- name: Set kernel.panic=10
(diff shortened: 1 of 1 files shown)


ansible/tasks/setup-system.yml
```diff
@@ -131,3 +131,18 @@
     mode: 0644
     owner: root
     group: root
+
+#Set Sysctl params for restarting the OS on oom after 10
+- name: Set vm.panic_on_oom=1
+  ansible.builtin.sysctl:
+    name: vm.panic_on_oom
+    value: '1'
```
darora, line 139: Do we want 2? I guess 1 is fine right now since PG memory usage is not capped?

darora (approved):

ansible/tasks/setup-system.yml
supaPatrick: I thought 1 seemed better sense we aren't using cgroups.

darora: Some of our services use systemd resource controls (e.g. https://github.com/supabase/postgres/blob/develop/ansible/files/gotrue.service.j2#L13) which internally uses cgroups afaik

PR #763, opened 2023-10-18
fix: publish migrations workflow

darora:


darora:
Clsoing in favour of #765

PR #774, opened 2023-10-26
chore: fix wrappers' lib name

pcnc:
* Wrappers saves its lib `.so` files using the current version in its name - e.g. `wrappers-0.1.16.so`, which makes pg_upgrade crash whenever a version is upgraded
* post-upgrade PG loads wrappers just fine since it isn't keeping any equivalence of extname<>libfilename - just loads them all up, then searches for any registry entry

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -46,6 +46,7 @@ OLD_PGVERSION=$(run_sql -A -t -c "SHOW server_version;")
 
 POSTGRES_CONFIG_PATH="/etc/postgresql/postgresql.conf"
 PGBINOLD="/usr/lib/postgresql/bin"
+PGLIBOLD="/usr/lib/postgresql/lib"
 
 # If upgrading from older major PG versions, disable specific extensions
 if [[ "$OLD_PGVERSION" =~ ^14.* ]]; then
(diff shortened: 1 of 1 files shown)


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -247,6 +248,18 @@ function initiate_upgrade {
...
     cp --remove-destination "$PGLIBNEW"/*.control "$PGSHARENEW/extension/"
     cp --remove-destination "$PGLIBNEW"/*.sql "$PGSHARENEW/extension/"
+
+    if [ -d "$PGLIBOLD" ]; then
+        WRAPPERS_LIB_PATH=$(find "$PGLIBNEW" -name "wrappers*so")
+        if [ -f "$WRAPPERS_LIB_PATH" ]; then
+            OLD_WRAPPER_LIB_PATH=$(find "$PGLIBOLD" -name "wrappers*so")
+            if [ -f "$OLD_WRAPPER_LIB_PATH" ]; then
+                LIB_FILE_NAME=$(basename "$OLD_WRAPPER_LIB_PATH")
+                cp "$WRAPPERS_LIB_PATH" "$PGLIBNEW/${LIB_FILE_NAME}"
```
darora, line 269: ooof

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -247,6 +248,18 @@ function initiate_upgrade {
 
     cp --remove-destination "$PGLIBNEW"/*.control "$PGSHARENEW/extension/"
     cp --remove-destination "$PGLIBNEW"/*.sql "$PGSHARENEW/extension/"
+
+    if [ -d "$PGLIBOLD" ]; then
+        WRAPPERS_LIB_PATH=$(find "$PGLIBNEW" -name "wrappers*so")
+        if [ -f "$WRAPPERS_LIB_PATH" ]; then
+            OLD_WRAPPER_LIB_PATH=$(find "$PGLIBOLD" -name "wrappers*so")
```
darora: the deb upgrades ensure that we won't match multiple files here? (not sure of behaviour of `-f` if multiple matches are passed in )

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
darora: might be worth adding comments here on why we're doing this, and maybe also adding a few echo statements to enhance traceability

darora (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
pcnc: Good point - limited output.

PR #775, opened 2023-10-26
chore: remove postgres role perms post-upgrade

pcnc:
* pg_upgrade requires superuser perms
* we grant postgres superuser perms pre-upgrade
* perms are ported over to the new instance through the migrated data dir


bmpandrade (approved):
Nice catch 🥳 👍

PR #779, opened 2023-11-01
feat: add h3-pg (#664)

### LINKED ISSUE #664: Feature Request: h3-pg extension
A user has requested Uber's [h3-pg](https://github.com/zachasme/h3-pg)

Issue open to track interest/engagement in a single location

kav:

Adds h3-pg.

## Additional context

I've upgraded cmake in the h3-pq layer. If there is interest I can lift that to the builder. Also not sure if I'm doing the pgTap test quite correctly. Not seeing any other extension smoke tests

Fixes #664

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,8 @@ ARG pgvector_release=0.4.0
 ARG pg_tle_release=1.0.3
 ARG supautils_release=1.9.0
 ARG wal_g_release=2.0.1
+ARG h3_pg_release=4.1.3
+ARG h3_pg_release_checksum
 
 ####################
 # Setup Postgres PPA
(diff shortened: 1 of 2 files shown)


kav:
@olirice any advice on moving this one forward? Happy to add more tests or tweaking things but not sure what's needed

olirice:
hi @kav 

thank you for opening and the PR looks good

we're not currently considering new extension to the platform right now. There have been a couple of extensions we already support that have caused major headaches for our pg16 migration and we're working out how best to resolve them

Its possible (likely) that we will support h3-pg in the future given that its large/stable & supported by a few other hosted platforms but our long-term goal is to push adoption of [TLEs](https://github.com/aws/pg_tle) so we're not in users' way

kav:
That's fair, I did a rough POC of a TLE version of h3 for postgres using the rust h3 bindings. Seemed reasonable but not sure I've got the bandwidth to maintain that

kav:
Hi @olirice, just checking if h3 is still on the potential list? Happy to bring this back to life if so.

olirice:
it is still on the potential list, but not something we can add currently

new constraints, but same outcome
there's a lot of stability work in progress to e.g.  reduce the number of images we build + package multiple extensions in each image so users are in charge of upgrading

Extensions are the single biggest source of issues we see in user projects so we're taking the time to get the tooling best-in-class before adding more

we'll get back into feature mode, it'll just take some time!

PR #782, opened 2023-11-02
feat: added pg_idkit extension

chetan:

Adds the pg_idkit extension

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -825,7 +825,7 @@ RUN arch=$([ "$TARGETARCH" = "arm64" ] && echo "aarch64" || echo "$TARGETARCH") 
 # 30-pg_idkit
 ####################
 FROM ccache as pg_idkit-source
-WORKDIR /root/
+WORKDIR /tmp
 RUN apt-get update && apt-get install -y git libpq-dev \
         postgresql-15 \
         postgresql-server-dev-15 \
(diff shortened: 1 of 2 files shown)


chetan:
I should note that I didn't totally understand all the `ccache` stuff, so happy to make any changes to fix those bits.

kav:
I'm not a maintainer so take this with a grain of salt but I think the pattern in the repository would have you set the workdir to `/tmp/` rather than `/root/` and then add a step at the end to run  dpkg-deb to bundle your deb.
`RUN dpkg-deb --build --root-owner-group ./pg_idkit/target/release/pg_idkit-pg15/`

 Then instead of copying the individual files over you can just do a 
`COPY --from=pg_idkit-source /tmp/*.deb /tmp/`

Hope this gets merged soon through, it's a nice package for those of us who uuid

Note: using `dpkg-deb`, which you may need to install, rather than `checkinstall` since `pgrx` produces root based file layouts rather than using `make install`

chetan:
Thanks @kav. I didn't mess around with the toolchain too much to figure out if it would be possible to produce a deb. I'll give this a try and update the PR.

olirice:
thanks @chetan

we've decided not to support pg_idkit extension on the platform

It's a great project, but it's also a good candidate to be converted into a trusted language extension. Given the relative ease releasing TLEs compared to full extensions, that's the route we'd prefer for this functionality.

Much of it has already been ported [here](https://database.dev/kiwicopple/pg_idkit).

PR #785, opened 2023-11-08
fix: drop pg_cron event trigger

soedirgo:
We already have an after-create script in https://github.com/supabase/postgres/blob/13b317b3bc74e4b2da7c1860b6f43e432fb17c8d/ansible/files/postgresql_extension_custom_scripts/pg_cron/after-create.sql , so the event trigger is redundant.

Why move away from event triggers: https://www.notion.so/supabase/move-away-from-event-triggers-to-custom-extension-scripts-08cb6c4fae39422ba57fd11135d71659?pvs=4

15.1.0.132 hasn't been deployed so seems fine to edit the migration in place, if not I'll write the change as a new migration

### DIFF
--- a/common.vars.pkr.hcl
+++ b/common.vars.pkr.hcl
@@ -1,1 +1,1 @@
-postgres-version = "15.1.0.132"
+postgres-version = "15.1.0.133"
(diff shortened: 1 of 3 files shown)


soedirgo:
Seems like migrations are deployed on prod upon merging to develop. That means we can't merge this PR now, because restored projects won't have the after-create.sql until the new AMI is out on prod.

We can remove the event trigger in a new AMI.

PR #787, opened 2023-11-09
feat: build extensions using orioledb as base

sweatybridge:

feature

## What is the new behavior?

Extensions with modified Makefile
- [x] pg_cron
- [x] pg_repack
- [x] pgvector (arm64 / clang)

Extensions with patched source
- [x] timescaledb
- [x] rum

Extensions which are currently installed as deb but need to be rebuilt
- [ ] pg_graphql
- [ ] pg_jsonschema
- [ ] wrappers
- [x] supautils
- [ ] plv8

## Additional context

### DIFF
--- a/.github/workflows/dockerhub-release-orioledb.yml
+++ b/.github/workflows/dockerhub-release-orioledb.yml
@@ -1,0 +1,87 @@
+name: Release OrioleDB on Dockerhub
+
+on:
+  push:
+    branches:
+      - develop
+      - orioledb
+    paths:
+      - ".github/workflows/dockerhub-release-orioledb.yml"
+      - "docker/orioledb/*"
+      - "common.vars*"
+
+jobs:
+  settings:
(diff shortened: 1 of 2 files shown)


olirice:
@sweatybridge I understand the issue with the pgrx extensions but could you give me more context on why `supautils` needs to be rebuilt? it looks like the deb is being built on top of orioledb already

sweatybridge:
Yup, I updated supautils to build from source in this commit https://github.com/supabase/postgres/commit/f57c806aaf091d3cca58a0d8c8febdb53b5bc13c and checked off the todo. Previously it was installing from deb but orioledb couldn't locate the installed extension at runtime.

olirice:
@sweatybridge 
Raminder is going to take a look at adding the pgrx extensions but none of them are blocking for the CLI or Studio so this can merge whenever you're ready and we'll add those in a separate PR

pcnc (approved):

PR #792, opened 2023-11-15
chore: bump adminapi and admin-mgr binaries

dragarcia:
* Also initialize `/etc/postgresql-custom/read-replica.conf` and ensure it's included in `postgresql.conf`.


ansible/files/postgresql_config/postgresql.conf.j2
```diff
@@ -763,6 +763,9 @@ jit_provider = 'llvmjit'		# JIT library to use
 # WAL-G specific configurations
 #include = '/etc/postgresql-custom/wal-g.conf'
 
+# read replica specific configurations
+#include = '/etc/postgresql-custom/read-replica.conf'
```
darora: Is there any reason to include it commented out here, only to uncomment it using sed/replace in all the artifacts we're producing? iirc we used this pattern for e.g. supautils config because we wanted to enable it only for some of the artifacts.

dragarcia: yea - it can be left uncommented but I was just standardizing with the other custom config files we have above. Have made it uncommented from the start instead. We could kaizen for some of the other `.conf` files later on like `wal-g.conf`.

darora (approved):

PR #799, opened 2023-11-20
chore: upgrade to newest adminapi

darora:
Also updates permissions to allow for management of Vector config and Postgres Exporter queries.


darora:
Need to double check that the permissions are set up correctly

pcnc (approved):

PR #800, opened 2023-11-20
chore: Fly refactoring; fixes; extra LSN shipping coverage

pcnc:
* Refactors LSN checkpoint creation and handling
  * Introduces handler that regularly ships the checkpoint file - used as a backup for cases such as volume corruption
    * Currently at 10 minute intervals, if the latest LSN has updated in the meantime
  * Ships the checkpoint after supervisord exits, as a backup to when signal handling goes awry
* Some fixes around perms and order of operations when unpacking the init payload


docker/all-in-one/entrypoint.sh
```diff
@@ -51,13 +53,57 @@ function enable_swap {
...
+  if [ ! -f "${LSN_CHECKPOINT_FILE_PATH}" ]; then
+    echo -n "0/0" > "${LSN_CHECKPOINT_FILE_PATH}"
+    touch "${LSN_CHECKPOINT_FILE_PATH}.previous"
+    chown postgres:postgres "${LSN_CHECKPOINT_FILE_PATH}"
+
+    chmod 0300 "${LSN_CHECKPOINT_FILE_PATH}"
   fi
 }
 
+function push_lsn_checkpoint_file {
```
darora, line 55: Should we consider moving the LSN related complexity a level down into the admin-mgr, where it can also be better tested?

docker/all-in-one/entrypoint.sh
```diff
@@ -160,6 +206,8 @@ function start_supervisor {
 
   # Start supervisord
   /usr/bin/supervisord -c $SUPERVISOR_CONF
+
```
darora: stray newlines

docker/all-in-one/entrypoint.sh
darora: Could also consider moving down most of the entrypoint functionality, including the configure scripts etc. Or perhaps adding an image entrypoint binary in this repo, if admin-mgr (and cross-repo for entrypoint tasks) doesn't make sense

docker/all-in-one/shutdown_event_listener.py
```diff
@@ -0,0 +1,57 @@
...
+def write_stdout(s):
+    sys.stdout.write(s)
+    sys.stdout.flush()
+
+def write_stderr(s):
+    sys.stderr.write(s)
+    sys.stderr.flush()
+
+def main():
+    while 1:
```
darora: `while True`?

docker/all-in-one/shutdown_event_listener.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python3
```
darora: Needs tests

docker/all-in-one/shutdown_event_listener.py
```diff
@@ -0,0 +1,57 @@
...
+
+# This script is used to make sure the latest LSN checkpoint is persisted remotely
+#  before the container is stopped. It is triggered every 60 seconds, whichever comes first.
```
darora: Perhaps I'm misreading the script, but how does any invocation terminate? Or is it a single instance that's spawned, and gets new events fed in over stdin as they occur?

docker/all-in-one/shutdown_event_listener.py
```diff
@@ -0,0 +1,57 @@
...
+    sys.stdout.write(s)
+    sys.stdout.flush()
+
+def write_stderr(s):
+    sys.stderr.write(s)
+    sys.stderr.flush()
+
+def main():
+    while 1:
+        write_stdout('READY\n')
```
darora: This output is a contract with supervisord?

darora: do we need additional debugging output to differentiate when the script didn't persist things for w/e reason?

docker/all-in-one/shutdown_event_listener.py
```diff
@@ -0,0 +1,57 @@
...
+            if os.path.getmtime(checkpointFilePrevious) < time.time() - 60 * LSN_CHECKPOINT_SHIP_INTERVAL:
+                previousLSN = ''
+                with open(checkpointFilePrevious) as f:
+                    previousLSN = f.read()
+
+                # If the current LSN is different from the previous one, persist it remotely
+                with open(checkpointFile) as f:
+                    currentLSN = f.read()
+                    if currentLSN != '0/0' and currentLSN != previousLSN:
+                        subprocess.run(["/usr/bin/admin-mgr", "lsn-checkpoint-push"])
```
darora: I don't think this captures stdout/stderr, or even error out if admin-mgr runs into an error?

darora:
Testinfra AIO tests were passing in the past I think?

docker/all-in-one/shutdown_event_listener.py
pcnc: It's an ongoing script which supervisord keeps alive then feeds it events, yes.

docker/all-in-one/shutdown_event_listener.py
pcnc: >This output is a contract with supervisord?

Yep - `READY` means supervisord can send events
http://supervisord.org/events.html#event-listener-states

End-state of an event completing succesfully is 
```
RESULT <resultLen>
<result>
```

>do we need additional debugging output to differentiate when the script didn't persist things for w/e reason?

Yep. Would definitely be worth having it log to its own log file and log whenever it bails on persisting

docker/all-in-one/shutdown_event_listener.py
pcnc: admin-mgr logs all its output to `/var/log/walg/pitr.log` by default :/

Agreed that it doesn't capture any error states for further debugging. addressing

docker/all-in-one/entrypoint.sh
pcnc: Good point, going to move LSN handling fully into adminmgr - better for it to be self-contained rather than handled in a handful of other ways. Would simplify the event listener considerably as well

bmpandrade:
Still failing test-infra-aio with:
```
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url: /rest/v1/buckets (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xffffbc858f70>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

Looks like services are taking longer to start and it fails. 

@pcnc , will look into the test suite and see if there is a hold we can do until services are up and running, or if the service itself failed to start.

pcnc:
@soedirgo - added additional handling so that the testinfra AIO workflow will attempt checking if the container is alive for a max of 60 times, or until it's marked as `exited`, before outright failing + print container logs: https://github.com/supabase/postgres/blob/pcnc/fly-fixes-redux/testinfra/test_all_in_one.py#L71-L76

This should result in a bit faster feedback timing-wise

pcnc:
> Testinfra AIO tests were passing in the past I think?

Fixed - supervisord service entry missing an env var which was added in the later steps of developing the PR.
Also improved the testinfra script to fail whenever the container is marked as `exited` instead of continuously waiting for it to become healthy + output container logs on failure

.github/workflows/dockerhub-release.yml
```diff
@@ -7,7 +7,7 @@ on:
     paths:
       - ".github/workflows/dockerhub-release.yml"
       - "common.vars*"
-
+
```
bmpandrade:
```suggestion
 
```

docker/all-in-one/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG postgres_version=15.1.0.83
+ARG postgres_version=15.1.0.136
```
bmpandrade:
```suggestion
ARG postgres_version=15.1.0.148
```

Maybe we could bump into latest? Anyway, not a blocker since everything runs again.

docker/all-in-one/entrypoint.sh
```diff
@@ -50,14 +54,40 @@ function enable_swap {
...
+  kill -s TERM "$(supervisorctl pid)"
+}
+
+function enable_autoshutdown {
+    sed -i "s/autostart=.*/autostart=true/" /etc/supervisor/db-only/supa-shutdown.conf
+}
+
+function enable_lsn_checkpoint_push {
+    sed -i "s/autostart=.*/autostart=true/" /etc/supervisor/db-only/lsn-checkpoint-push.conf
```
bmpandrade: Isn't this required even if not on a box configured as `db-only`? 🤔  (at least for all Fly boxes)

Similar thought for the remaining `supa-shutdown` and `lsn-checkpoint-push` files for supervisord or references, that bring up `db-only`.

bmpandrade (commented):
In general it looks good, but I got a bit confused about the `db-only` as referred on the other code comments.

I thought, that only on special cases `db-only` was used, and the fly extensions was not one of them. Wouldn't this make us update the worker too, when deploying to Fly to treat it as a `db-only`, to ensure that LSN checkpoints were being properly uploaded?

docker/all-in-one/Dockerfile
pcnc: Good point - better to have an up-to-date default base

docker/all-in-one/entrypoint.sh
pcnc: Ah, the `db-only` folder is the default (and minimal) set of services which supervisord runs:
https://github.com/supabase/postgres/blob/develop/docker/all-in-one/etc/supervisor/supervisord.conf#L169-L170

During regular operation if `POSTGRES_ONLY` is not true, `services/*conf` also gets added:
https://github.com/supabase/postgres/blob/develop/docker/all-in-one/entrypoint.sh#L234-L238

Though, good point - this causes quite a bit of confusion - better to rename the `db-only` folder to `default` or `base-services`

bmpandrade: Got it. Thanks for the points it might have escaped me but, yeah, unless you are very comfortable already with the term `db-only` it might be a bit misleading. Change to `base-services`, since the other folder is already called `services` sounds good to me 👌 

Regarding checks/tests:
- `Testinfra Integration Tests / test-all-in-one ` is failing with
```
FAILED testinfra/test_all_in_one.py::test_postgrest_responds_to_requests - requests.exceptions.ConnectionError:
HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url:
 /rest/v1/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection 
object at 0x7f8a8357bf70>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

- `Testinfra Integration Tests / test-ami` is failing with 
```
Build 'amazon-ebssurrogate.source' errored after 11 minutes 52 seconds: 
Script exited with non-zero exit status: 2.Allowed exit codes are: [0]
```

darora (approved):

PR #801, opened 2023-11-21
fix: drop obsolete pg_graphql function signature

soedirgo:
It's breaking some restores because the `grant_pg_graphql_access` expects only one function signature on this line:

https://github.com/supabase/postgres/blob/87ee42dda7082be8957fd2936c2f23ff6ed7ab77/migrations/schema.sql#L270

Error:
```
ERROR:  function name "graphql_public.graphql" is not uniqueHINT:  Specify the argument list to select the function unambiguously.
CONTEXT:  SQL statement "DROP FUNCTION IF EXISTS graphql_public.graphql"
PL/pgSQL function grant_pg_graphql_access() line 22 at SQL statement
SQL statement "create extension if not exists pg_graphql"
PL/pgSQL function inline_code_block line 13 at SQL statement
```

### DIFF
--- a/.github/workflows/testinfra.yml
+++ b/.github/workflows/testinfra.yml
@@ -137,4 +137,4 @@ jobs:
       - name: Cleanup resources on build cancellation
         if: ${{ always() }}
         run: |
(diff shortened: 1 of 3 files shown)


darora:
Merging to develop should be enough to have it picked up for restores

darora (approved):
This would work afaik - but will wait for @olirice 's input in case there's a better overall mechanism

olirice:
sorry, this ping must have been buried in 100s of GitHub notifications. Yes, this looks good, thanks for knocking it out!

PR #805, opened 2023-11-27
chore: wait a few seconds before an additional psql stop during upgrades

pcnc:

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -300,9 +300,15 @@ EOF
         UPGRADE_COMMAND="$UPGRADE_COMMAND --check"
     else 
         echo "9. Stopping postgres; running pg_upgrade"
+
+        # Extra work to ensure postgres is actually stopped
+        #  Mostly needed for PG12 projects with odd systemd unit behavior
         retry 5 systemctl restart postgresql
         systemctl disable postgresql
         retry 5 systemctl stop postgresql
+
+        sleep 3
+        systemctl stop postgresql
     fi
 
     su -c "$UPGRADE_COMMAND" -s "$SHELL" postgres


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -300,9 +300,15 @@ EOF
...
         echo "9. Stopping postgres; running pg_upgrade"
+
+        # Extra work to ensure postgres is actually stopped
+        #  Mostly needed for PG12 projects with odd systemd unit behavior
         retry 5 systemctl restart postgresql
         systemctl disable postgresql
         retry 5 systemctl stop postgresql
+
+        sleep 3
+        systemctl stop postgresql
```
bmpandrade, line 311: Interesting.

Wouldn't bump the `retry 5` to `retry 10` actually helps, instead of adding a sleep + repeat command?

bmpandrade:
Also, might be missing a postgres version upgrade?

I've a version update pending though: https://github.com/supabase/postgres/pull/803

pcnc:
> Also, might be missing a postgres version upgrade?
> 
> I've a version update pending though: #803

The pg_upgrade scripts workflow update the scripts for whichever version is latest in common.vars.hcl.
The scripts also get downloaded on-demand before every upgrade - thus this would patch the scripts for the version specified in common.vars.hcl.

Given we've already had some iterations which bump the pg version number, I went ahead and updated prod and staging tarballs so we can unblock the PG12 deprecation upgrades 😬

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
pcnc: `retry` only retries until an exit code of `0` is returned - the retried `systemctl stop postgresql` completes successfully, but PG immediately restarts :(

Waiting a few seconds after cycling the service state and stopping it again seems to do the trick.
This is likewise just a temporary fix for PG12->PG15 upgrades

bmpandrade (approved):

PR #809, opened 2023-11-29
chore: bump version

delgado3d:
bumping because 15.1.0.135 release workflow failed partially


darora:
AMI seems to have been built - what was the failure?

darora (approved):

PR #813, opened 2023-11-30
chore: remove plv8 from exclusions

pcnc:
* keeps plv8 around when upgrading

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -11,9 +11,6 @@
 # them depending on regtypes referencing system OIDs or outdated library files.
 EXTENSIONS_TO_DISABLE=(
     "pg_graphql"
-    "plv8"
-    "plcoffee"
-    "plls"
 )
 
 PG14_EXTENSIONS_TO_DISABLE=(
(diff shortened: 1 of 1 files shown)


bmpandrade (approved):

bmpandrade:
`test-ami` was failing with 

```
==> amazon-ebssurrogate.source: Error deregistering existing AMI: UnauthorizedOperation: 
You are not authorized to perform this operation. 
User: arn:aws:sts::436098097459:assumed-role/internal-actions-runners-role/1701348762036699624 is 
not authorized to perform: ec2:DeregisterImage on resource: 
arn:aws:ec2:ap-southeast-1::image/ami-0e35fa03abcc1b9e7 because no 
identity-based policy allows the ec2:DeregisterImage action. 
```

Since there was another PRs with similar issues, that seem to already been fixed, I've triggered a re-run on the failed job.

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -60,6 +58,48 @@ fi
...
+    #  naming scheme, containing the version in their library's file name
+    #  e.g. wrappers-0.1.16.so, rather than wrappers.so
+    # pg_upgrade errors out when it doesn't find an equivalent file in the new PG version's
+    #  library directory, so we're making sure the new version has the expected (old version's)
+    #  file name.
+    # After the upgrade completes, the new version's library file is used.
+    # i.e. 
+    #  - old version: wrappers-0.1.16.so
+    #  - new version: wrappers-0.1.18.so
+    #  - workaround to make pg_upgrade happy: copy wrappers-0.1.18.so to wrappers-0.1.16.so
```
darora: this feels so wrong, and is going to be hard to debug post-hoc. I take it that symlinks don't work?

If they don't, maybe we can touch a hidden file indicating what we've done as part of the upgrade?

darora (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
pcnc: Symlinks should work, as they're regularly used by lib files. Switching to use symlinks

PR #743, opened 2023-09-27
WIP adding Envoy to the image

supaPatrick:
this is what I think the initial commit for envoy into the image would look like. Envoy would be turned off and not in the same slice as other services but this commit would allow me to test easier moving forward.

PR #817, opened 2023-12-01
feat: add Envoy proxy support

thebengeu:
Works as intended as far as I've tested with the all-in-one image and manually disabling Kong then enabling Envoy using the AMI.

- Envoy is disabled by default.
- The YAML config files are generated by https://github.com/supabase/infrastructure/blob/beng/envoy/worker/src/lib/config-utils/envoy.ts, better to review those there since there are comments.
- Referenced https://github.com/supabase/postgres/pull/703 and https://github.com/supabase/postgres/pull/743 a bit, but mostly rewritten.
- Used `/etc/envoy` to be consistent with the official Envoy image, and there are many config files instead of just one, so https://github.com/supabase/supabase-admin-api/pull/174 won't work, Admin API changes in https://github.com/supabase/supabase-admin-api/pull/185.
- Keeping both kong and envoy in the services slice. I'm not sure what the issue was expected to be, when restarting using the services.slice, only running services are restarted, so restart doesn't say, restart kong if kong is stopped and envoy is running.


pcnc (approved):

ansible/files/adminapi.sudoers.conf
```diff
@@ -1,3 +1,4 @@
+Cmnd_Alias ENVOY = /bin/systemctl start envoy.service, /bin/systemctl stop envoy.service, /bin/systemctl restart envoy.service, /bin/systemctl disable envoy.service, /bin/systemctl enable envoy.service, /bin/systemctl reload envoy.service
```
darora, line 1: You'll need to add a grant for adminapi to be able to use all these (refer to entries on lines 24-27)

ansible/files/envoy_config/lds.yaml
```diff
@@ -0,0 +1,334 @@
+resources:
```
darora, line 1: Do some or all of these config files get overwritten during the boot process? We don't currently include generated versions of managed files here atm

ansible/files/start-envoy.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+# Workaround using `tee` to get `/dev/stdout` access logging to work, see:
+# https://github.com/envoyproxy/envoy/issues/8297#issuecomment-620659781
+exec /opt/envoy --config-path /etc/envoy/envoy.yaml --restart-epoch "$RESTART_EPOCH" | tee
```
darora: Likely want https://github.com/envoyproxy/envoy/issues/8297#issuecomment-1548401607 if we're doing this (use bash, set -euo pipefail, presumably capturing stderr through tee is also needed?)

ansible/files/envoy_config/remove_apikey_query_parameter.lua
```diff
@@ -0,0 +1,5 @@
+function envoy_on_request(request_handle)
```
darora, line 1: Add comment describing why we're using this function. If possible, a test capturing the intended behaviour.

ansible/files/envoy_config/lds.yaml
thebengeu: This is just temporary for AMI testing purposes. `lds.yaml` needs to be generated, the rest are static but the config generation code for envoy.yaml and cds.yaml is in worker to keep the source of truth for the YAML files in one place and DRY.

Building and testing the AMI one more time before merging, then will move `lds.yaml` elsewhere to be all-in-one specific.

ansible/files/envoy_config/remove_apikey_query_parameter.lua
thebengeu: Added comment and tests.

ansible/files/envoy_config/lds.yaml
thebengeu: Actually, why not include the generated files only in this repo and in the AMI, as the source of truth, instead of having duplicate versions in `init-scripts`? Only `cds.yaml` would have to be overwritten to enable TLS for admin API, which is disabled in all-in-one.

darora: Ends up splitting up the source of truth, especially given at least some of the files (looks like lds.yaml in this case) needs to be updated in both init-scripts and the worker at the same time.

darora (approved):

ansible/files/envoy_config/lds.yaml
thebengeu: Per offline discussion, at least for now `lds.yaml` needs to be included in this repo so that external partners are able to use the all-in-one image as-is.

Ideally we'd consolidate all such config to be injected via remote payload, such that the init process for the AMI and Fly is more consistent. If necessary for external partners, we could publish instructions and the necessary config publicly.

PR #828, opened 2023-12-07
fix: remove unnecessary comment and bump `postgres-version`

thebengeu:
Previous `chore` commit didn't trigger release.


darora:
Oh this repo doesn't use the semantic release plugin that uses the commit messages 😬 . The version string is manually maintained in common.vars.pkr.hcl ; you need to adjust it manually if you want a new release

darora (approved):
Approving pre-emptively for when you do change the release version

PR #834, opened 2023-12-08
chore: enable the en_US.UTF-8 locale

darora:


pcnc (approved):

pcnc:
Also bumped PG version to avoid shipping anything without including this change

PR #849, opened 2023-12-20
feat: add columnar db

riderx:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ ARG groonga_release=12.0.8
 ARG pgroonga_release=2.4.0
 ARG wrappers_release=0.2.0
 ARG hypopg_release=1.3.1
+ARG columnar_release=1.1.0
 ARG pg_repack_release=1.4.8
 ARG pgvector_release=0.4.0
 ARG pg_tle_release=1.0.3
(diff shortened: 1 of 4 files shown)


olirice:
thank you for opening and the PR looks good

The bar for adding new extensions to the platform is very high as it's a big commitment from us to maintain them indefinitely if the current authors become unavailable. In this case we the extension is fairly new (< 18 months old) and the platform already supports a comparable extension through TimescaleDB's column store. For those reasons, I'm going to close this for now. If we see a big uptick in requests for hydra we can revisit in the future

riderx:
Thanks @olirice for the quick reply.

I understand your concern, to give some context, Hydra is a company running cloud DB like Supabase.
It's a fork of an older project, [citus](https://github.com/citusdata/citus) **7 years old**.

Hydra isolated the extension before citus took a direction of Distributed DB, I could have used columnar from citus.

In my point of view (as Supabase cloud client) TimescaleDB is not doing the job, the license prevents any interesting use of TimescaleDB.
https://docs.timescale.com/about/latest/timescaledb-editions/
No compression, no Continuous aggregates, no retention policy, no Actions and automation.

So by adding an alternative, i'm resolving my current issue in Supabase Cloud.

So I would be grateful if you could consider this a bit further.

Columnar is "Next generation of cstore_fdw", using latest best practices of Postgres.

This could open many use cases in analytics for cloud user who cannot use TimescaleDB

I'm open to any alternative I found hydra was more close in philosophy that citus to supabase.

A good alternative could be to use https://pgxman.com/ for extension since it's made for install extension and has hydra listed.

olirice:
thanks for the links, very interesting read

I think long term we are likely to revisit column storage in addition to timescale. Before we make any decisions we'll need to do an internal review of options that includes:
- features
- performance
- alignment with our products' direction
- supportability by our tech team
- supportability by our support team (we handle support for our extensions as well as the supabase platform)

I'll make sure we include hydra in that comparison when we circle back to tackle to it, but its not something we can add for right now

riderx:
@olirice okay, i will keep using clickhouse until i can put back my analytics into Supabase.
Thank for considering it !

PR #889, opened 2024-02-17
fix: search_path /$user remove backslash

taytzehao:

Bug fix

The current search_path has "/$user" rather than the default value

## What is the new behavior?

The search_path has "$user" with the slash removed


olirice:

PR #902, opened 2024-03-08
fix: use dedicated role for pg_upgrade

soedirgo:
[Context](https://supabase.slack.com/archives/C01D6TWFFFW/p1702959923630909)

The hypothesis is pg_upgrade changes all extensions' owner to whichever username was used for the upgrade. If that's the case, performing the pg_upgrade with a dedicated role and then `REASSIGN OWNED`-ing it should (🤞) solve the above issue.


soedirgo:
This approach doesn't work :/

PR #907, opened 2024-03-14
chore: remove pgbouncer from docker aio image [GEN-8039]

bmpandrade:
Removes pgbouncer from Docker AIO image.

Also adds some conditionals around healthcheck.sh to avoid running checks when gotrue/fail2ban are disabled.


tomashley (approved):
looks sane

olirice (approved):
lgtm

PR #925, opened 2024-03-28
Sam/sync versions

samrose:

This PR introduces a `/nix` folder in this repo, plus a `flake.nix` and `flake.lock` that facilitate using the Nix package management system to package supabase/postgres, and all of our extensions and wrappers. A user will need nix installed on their machine. As of 4/1/2024 the package set only builds on target machines (`x86_64-linux` and `aarch64-linux`), however work is under way to also support building and using directly on `aarch64-darwin` (macOs). As of 4/1/2024, versions of packages and extensions are synced from `/ansible/vars.yml` via a utility that can be run by executing `nix run .#sync-exts-versions` (you must have nix installed and be on the supported `x86_64-linux` and `aarch64-linux` for this command to work). The short term goal is to sync these versions as they are updated by our infrastructure and postgres teams, then to see the nix packaged versions build successfully in parallel over time, along with tests of the nix packaged versions passing. 

The supabase/postgres repo will continue to source it's dependencies from ansible for the short term, while we stabilize this nix build. 

This pr also updates postgres to 15.6

## Next steps
Forthcoming PR's will include: 

*  integrating the nix work into our ansible/packer builds, 
*  building natively on aarch64-darwin (macOs),
* ability to override and specify postgresql 15 version directly
*  more testing
*  eventually a git pre commit hook to perform the sync (needs to be able to build on macos first)
*  compare our built version of postgresql 15.6 to the version sourced from debian packages that is currently in user (to compare patches, etc to avoid surprises)

## Impact on users of this repo

As of this PR, there is no impact on users of the repo.


olirice (approved):

olirice (approved):

soedirgo (approved):

darora (approved):
Approving to unblock further work. Not carefully reviewed (I'm not the right person to review nix code in any case), which is OK here as it doesn't modify the existing build.

PR #943, opened 2024-04-11
Sam/pgrx 0 11 3 support

### LINKED ISSUE #942: Resolve nix build on aarch64-linux
# Chore

Native/non-cross-compiled builds are failing on aarch64-linux platform. We need to achieve native builds to create and publish nix binary caches for the platforms we need to support.

samrose:

I discovered that on some distros (not just in nix builds) that aarch64-linux builds of our psql_15 bundle were failing due to this issue

https://github.com/pgcentralfoundation/pgrx/issues/1429#issuecomment-1852730951

The problem was fixed in `rust-nindgen` and then ported to pgrx 0.11.3

We updated `pg_graphql`, `pg_jsonschema` and `supabase-wrappers` to use 0.11.3, tested directly  on x86_64-linux and aarch64-linux (including assuring that `nix flake check -L --all-systems` passes on both systems)

Updating to 0.11.3 required a bump of `nixpkgs-unstable` which is found in the changes at https://github.com/supabase/postgres/compare/sam/pgrx-0-11-3-support?expand=1&template=default.md#diff-216b2b7bfde9416c79d133bacb031e95702a20bdedb548c0b055c837aa4f6a9c  this is achieved by running `nix flake update nixpkgs`  , which is tied to the `nixpkgs-unstable` in our flake https://github.com/supabase/postgres/blob/f0a2e4153022c53e1d82e44f7290a5451c12075e/flake.nix#L5

pgrx 0.11.3 in nix required that we set a var `CARGO` (in this case setting it to the cargo supplied by nixpkgs see: https://github.com/supabase/postgres/compare/sam/pgrx-0-11-3-support?expand=1&template=default.md#diff-0742481ee07030a8964829237ee62f3a15a7796f0daea6a0361b2b67292eb9abR17 )

I needed to update the gh action workflows to make sure we are not just cross compiling, but actually building on the native machine/platform for each platform we want to support currently (for this pr x86_64-linux and aarch64-linux ).  

It's necessary for ci builds that happen on our self hosted runners to be run from within a docker container. 

So, this  PR converts the build for x86 and arm64 to run from within a docker container (using the open source release container of nixpkgs/nix-flakes).

We needed to run some base line tests upon successful completion of the build on each system, and so we're running `nix flake check -L --all-systems` at this stage, which runs some basic checks that the postgres server will start as expected, and that certain extensions will install successfully/without error.

We are building the docker image, but not yet deploying it at this stage (building to test that the image build succeeds with any given change. There will be a future refactored action to deploy the image on merge to `develop` branch

closes #942


samrose:
Will need a couple iterations/commits  to resolve gh actions issues

pashkinelfe (approved):

pcnc (approved):

PR #951, opened 2024-04-18
feat: salt state for supervisor gotrue/fail2ban

tomashley:


docker/all-in-one/salt-init/docker/etc/minion
```diff
@@ -0,0 +1,6 @@
+master: salt-master
```
bmpandrade: Since we know we will not use master at this point for Fly and since we are touching the /etc/minion anyways, wouldn't be preferable to just set `file_client: local`? And we can remove `--local` from salt-call runs. 
And avoid for some weird reasoning if a salt-call is called later, it doesn't end up 1st trying to look for the master, if we already know it doesn't exist.

Anyway, end result with `file_client: local` and `--local` is the same, just a personal pref :) 

edit: and `master_type: disable` as well

tomashley: Good ideas. 
Have added them all

tomashley:
Abandoned in preference of https://github.com/supabase/postgres/pull/997

PR #952, opened 2024-04-18
chore: run logrotate every 5 mins

darora:
Some log files get produced at a rate faster than can be handled under
our current 10min freq.


bmpandrade (approved):
👍 

look at checks as there is a ci failing with "ERROR: failed to solve: Canceled: context canceled", but shouldn't have nothing to do with this code change, as it's pretty straightforward

PR #920, opened 2024-03-22
chore: remove pgbouncer from AIO image except pg role [GEN-8039]

bmpandrade:
Same changes as previous PR #907 , except this will keep the .sql that creates the pgbouncer role. This will be kept for now until the full pgbouncer cleanup happens, since there were migrations failing due to the lack of pgbouncer role present, when creating new projects in Fly, during tests in Staging.

Before opening the PR, this changes were tested with `aio-15.1.1.31-no-pgbouncer-docker` image in staging, which went through, creating a complete project with no issues and no pgbouncer presence, except for the role, as mentioned before.

PR #961, opened 2024-05-01
fix: gotrue config persistence for AIO

pcnc:
* Fixes GoTrue losing its config on subsequent restarts
* Replaces gotrue's and fail2ban's commands to ensure external processes don't restart the service's process itself
* Adds flag to disable pgbouncer


docker/all-in-one/entrypoint.sh
```diff
@@ -80,10 +80,23 @@ function enable_lsn_checkpoint_push {
 }
 
 function disable_fail2ban {
+  sed -i "s/command=.*/command=sleep 1/" /etc/supervisor/services/fail2ban.conf
```
tomashley: there is a danger here that this sleep isn't long enough. 
By default if a program doesn't stay up for 1 second, it is considered a failure and it will restart
I would set this to 2, or set startsecs to 0 - but startsecs isn't in the config file rn

from the manual

```
startsecs

The total number of seconds which the program needs to stay running after a startup to consider the start successful (moving the process from the STARTING state to the RUNNING state). Set to 0 to indicate that the program needn’t stay running for any particular amount of time.

Note

Even if a process exits with an “expected” exit code (see exitcodes), the start will still be considered a failure if the process exits quicker than startsecs.

Default: 1

Required: No.

Introduced: 3.0
```

pcnc: Ooh, nice catch. Will set it to `5` just to make sure it doesn't have any side effects

tomashley (approved):

delgado3d:
may need to update testinfra tests

bmpandrade:
@pcnc , related to disabled PGbouncer, had this stale PR for quite a while now, which removes it completely from AIO: https://github.com/supabase/postgres/pull/920 

Do you prefer to merge yours and than update my PR with gotrue/fail2ban changes that went in?

pashkinelfe (approved):

PR #963, opened 2024-05-07
fix: remove spurious newline

darora:
Not immediately obvious why, but causes the version string extraction
to break for upgrade script publication

https://github.com/supabase/postgres/actions/runs/8900354674/job/24441719357


darora:
d'oh, right, it would result in an empty line as well. Will make the script more resilient as well

samrose (approved):

delgado3d (approved):

samrose:
@darora if you keep seeing the CI issue where the nix install runs out of space on the ci run, please let me know.

PR #964, opened 2024-05-07
fix: remove empty line in release version

delgado3d:


darora:
https://github.com/supabase/postgres/pull/963 also makes the workflow script more resilient

PR #978, opened 2024-05-22
chore: temporarily hardcode PostgREST version for aio image

dragarcia:
* should allow the aio image to be built
    * https://github.com/supabase/postgres/actions/runs/9187951408


darora (approved):
Would suggest also adding a note in the bars file where the versionis defined

PR #980, opened 2024-05-23
feat: set admin-api to 0.64.1 and bump AMI version to 15.1.1.56 [GEN-8523]

bmpandrade:


darora (approved):

pashkinelfe (approved):

bmpandrade:
Validated between 15.1.1.55 and custom ami with admin-api 0.64.1, and it solves the issue where Auth fails for the 1st request that hit the DB, after toggling SSL Enforcement.

[internal slack ref](https://supabase.slack.com/archives/C01D6TWFFFW/p1716760694950799?thread_ts=1716501774.821939&cid=C01D6TWFFFW)

PR #981, opened 2024-05-27
Update postgresql.conf to increase max_slot_wal_keep_size

encima:

Increases `max_slot_wal_keep_size` default from 1GB

Current behaviour is 1GB by default.

As this is a setting we do not expose, the current default is too low to successfully replicate using a tool like Airbyte (as the WAL will grow too large during the Hourly sync window).

## What is the new behavior?

Proposing to set this as 4GB by default but 2GB could be a conservative option.

## Additional context

Some relevant issues are noted in the (internal) [Notion document](https://www.notion.so/supabase/RFC-Billing-DB-space-based-on-allocated-disk-space-84a13bd070ac4eaba2c1ef24c35ea98d?pvs=4)

### DIFF
--- a/ansible/files/postgresql_config/postgresql.conf.j2
+++ b/ansible/files/postgresql_config/postgresql.conf.j2
@@ -300,7 +300,7 @@ max_wal_senders = 10		# max number of walsender processes
 max_replication_slots = 5	# max number of replication slots
 				# (change requires restart)
 #wal_keep_size = 0		# in megabytes; 0 disables
-max_slot_wal_keep_size = 1024   # in megabytes; -1 disables
+max_slot_wal_keep_size = 4096   # in megabytes; -1 disables
 #wal_sender_timeout = 60s	# in milliseconds; 0 disables
 #track_commit_timestamp = off	# collect timestamp of transaction commit
 				# (change requires restart)


thebengeu (commented):
I think 4GB would be fine, is there wider agreement on this change already?

thebengeu (approved):
Discussed offline, there's agreement on this increase.

Tang8330:
Just curious - have y'all thought about making this configurable per https://supabase.com/docs/guides/platform/custom-postgres-config instead of upping the threshold?

Tang8330:
For customers that have a large Supabase instance, it'd be nice to have this number be a % of the total database size as opposed to having it fixed to 1GB (or soon 4GB)

encima:
> Just curious - have y'all thought about making this configurable per https://supabase.com/docs/guides/platform/custom-postgres-config instead of upping the threshold?

Yes, this should certainly be done. This current setting is to allow replication to function on all instances now but should be configurable in the future (currently it cannot be due to some infra designs but it will be).

jgraettinger:
Hi :wave: - checking in on the status of this increase.

Our (https://estuary.dev/) users frequently encounter logical replication slot invalidations due to the current 1GB setting. It's far too small in real-world usage, even after users make best-effort mitigations to prevent large transactions.

It would of course be wonderful to see this be configurable, but increasing to 4GB would stop much of the bleeding.

encima:
@thebengeu is this OK to merge now that it has passed checks?

thebengeu:
@encima yes, good to merge. Note that this change only affects new projects, and we'll need to roll out the change to existing projects where desired.

PR #982, opened 2024-05-27
chore: update delegated init function in entrypoint

pcnc:
* delegated entrypoint function updates to go with https://github.com/supabase/infrastructure/pull/18282
  * unfortunately two calls are needed as redirecting from a POST to an S3 signed URL GET doesn't work
    * can't really bundle it into adminapi to bypass this as supervisor isn't started at this time
* hit an auto-format on entrypoint.sh to get rid of tabs/extraneous spacing

### DIFF
--- a/docker/all-in-one/entrypoint.sh
+++ b/docker/all-in-one/entrypoint.sh
@@ -211,18 +211,9 @@ function start_supervisor {
 
 DELEGATED_ARCHIVE_PATH=/data/delegated-init.tar.gz
 
-fetch_delegated_payload() {
-  DELEGATED_PAYLOAD_URL=$(curl -XPOST -sf "$SUPABASE_URL/$DELEGATED_INIT_LOCATION" -H 'Content-Type: application/json' -d '{"reportingToken":"'"$REPORTING_TOKEN"'"}')
+function fetch_and_execute_delegated_payload {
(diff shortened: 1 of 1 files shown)


docker/all-in-one/entrypoint.sh
```diff
@@ -209,34 +209,43 @@ function start_supervisor {
...
+
+fetch_delegated_payload() {
+  DELEGATED_PAYLOAD_URL=$(curl -XPOST -sf "$SUPABASE_URL/$DELEGATED_INIT_LOCATION" -H 'Content-Type: application/json' -d '{"reportingToken":"'"$REPORTING_TOKEN"'"}')
+
+  if [ -z "$DELEGATED_PAYLOAD_URL" ]; then
+    echo "Failed to get response, bailing"
+    return
+  fi
+
+  curl -s --time-cond $DELEGATED_ARCHIVE_PATH -o $DELEGATED_ARCHIVE_PATH "$DELEGATED_PAYLOAD_URL"
```
pcnc: `--time-cond` here will throw some warnings if the file doesn't already exist, but they don't have any side-effects besides informational; curl also specifies in a verbose manner that it's actively disabling the flag

pashkinelfe (approved):

thebengeu (approved):

tomashley (approved):

PR #983, opened 2024-05-29
fix: fix invalid keys in systemd service

darora:
Requisite does not impose ordering, so I've added After as well.


darora:
@tomashley fyi

delgado3d (approved):

PR #986, opened 2024-06-03
Sam/nix ami docker

samrose:
## What kind of change does this PR introduce?

Refactoring the base Dockerfile to source it's base image from our nix build, and integrate that package bundle into the Dockerfile, with minimal other changes.

PR #987, opened 2024-06-03
Sam/nix ami docker aio

samrose:

This new docekr aio Docker image does the following:

- source from the image created at #986 
- creates a base AOI Dockerfile + Dockerfile targeting fly.io and another targeting local dev


samrose:
deprecated in favor of https://github.com/supabase/postgres/pull/1037

PR #995, opened 2024-06-06
feat: add auth generated env file

hf:
Registers the optional `/etc/gotrue/gotrue.generated.env` file on the Auth (`gotrue`) service. This file can be used by the Supabase Admin API to generate a base configuration, which then can be overriden by the Supabase Platform via `/etc/gotrue/gotrue.env` or Supabase Staff when necessary in `/etc/gotrue/gotrue.overrides.env`.

### DIFF
--- a/ansible/files/gotrue.service.j2
+++ b/ansible/files/gotrue.service.j2
@@ -12,6 +12,7 @@ RestartSec=3
 MemoryAccounting=true
 MemoryMax=50%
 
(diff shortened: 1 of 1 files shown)


darora (commented):
Looks good pending name change suggested in https://github.com/supabase/postgres/pull/996/files#r1630532315

darora (approved):

PR #996, opened 2024-06-06
feat: add auth optimizations service

hf:
Adds an Auth optimizatins service that writes to the `/etc/gotrue/gotrue.generated.env` file.

### DIFF
--- a/ansible/files/gotrue-optimizations.service.j2
+++ b/ansible/files/gotrue-optimizations.service.j2
@@ -1,0 +1,11 @@
+[Unit]
+Description=GoTrue (Auth) optimizations
+
+[Service]
+Type=oneshot
+# we don't want failures from this command to cause PG startup to fail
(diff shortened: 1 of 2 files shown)


ansible/files/gotrue-optimizations.service.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=GoTrue (Auth) optimizations
+
+[Service]
+Type=oneshot
+# we don't want failures from this command to cause PG startup to fail
+ExecStart=/bin/bash -c "/opt/supabase-admin-api optimize auth --destination-config-file-path /etc/gotrue/gotrue.base.env ; exit 0"
```
darora: Let's follow existing convention and using "generated" in the file name rather than "base"

hf: Done.

darora (approved):

PR #997, opened 2024-06-07
feat: delegated-init support for Fly projects

tomashley:

Feature: Installs salt in the all-in-one Docker container used in Fly.io projects. Enables a delegated entry point for the container.

## What is the new behavior?

## Additional context

Add any other context or screenshots.%

### DIFF
--- a/docker/all-in-one/entrypoint.sh
+++ b/docker/all-in-one/entrypoint.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -eou pipefail
 
+START_TIME=$(date +%s%N)
+
 PG_CONF=/etc/postgresql/postgresql.conf
 SUPERVISOR_CONF=/etc/supervisor/supervisord.conf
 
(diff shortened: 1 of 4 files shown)


pcnc (approved):

docker/all-in-one/entrypoint.sh
```diff
@@ -206,19 +209,42 @@ function start_supervisor {
...
+  curl -s --time-cond $DELEGATED_ARCHIVE_PATH -o $DELEGATED_ARCHIVE_PATH "$DELEGATED_INIT_LOCATION"
+
+  if [ ! -f $DELEGATED_ARCHIVE_PATH ]; then
+    echo "No delegated payload found, bailing"
+    return
+  fi
+
+  # only extract a valid archive
+  if [[ $(tar -tzf $DELEGATED_ARCHIVE_PATH) ]]; then
+    # TODO: (tom) Only extract newer tar archives
```
pcnc: Could we save the filemtime of the last extracted tarball in a persistent file and compare if current tarball filemtime is different?

tomashley: Have solved the time issue by adding a touch to the entryscript at build time
We could only be changing salt files, but this file will always definitively be in the archive.

PR #1011, opened 2024-06-24
chore: upgrade to latest auth version (v2.154.2)

hf:
Bumps Auth to v2.154.2 which fixes an issue with the `auth.uid()` and related functions.


delgado3d:
PR name is incorrect

delgado3d (approved):

PR #1018, opened 2024-07-02
feat: add `safeupdate` to `shared_preload_libraries`

soedirgo:

Feature

pg_safeupdate can't be used with the `postgres` role since setting `[shared/session]_preload_libraries` requires superuser.

## What is the new behavior?

Allow using pg_safeupdate with the `postgres` role by setting `safeupdate.enabled = on`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -614,10 +614,6 @@ RUN --mount=type=cache,target=/ccache,from=public.ecr.aws/supabase/postgres:ccac
 RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --nodoc
 
 ####################
-# 21-auto_explain.yml
-####################
-
-####################
 # 22-pg_jsonschema.yml
 ####################
 FROM base as pg_jsonschema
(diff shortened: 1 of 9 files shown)


ansible/tasks/postgres-extensions/09-pg-safeupdate.yml
```diff
@@ -28,3 +28,10 @@
   file:
     state: absent
     path: /tmp/pg-safeupdate-{{ pg_safeupdate_release }}
+
+- name: pg-safeupdate - set safeupdate.enabled
+  become: yes
+  lineinfile:
+    path: /etc/postgresql/postgresql.conf
+    state: present
+    line: safeupdate.enabled = off
```
darora: Why here instead of including in the base template?

soedirgo: We do this for other extensions (`cron.database_name`, `auto_explain.log_min_duration`, etc.) so just did the same here

soedirgo: I don't see a problem with including it in the base template though... going to smoke test that

soedirgo: Seems to work fine - updated the PR

steve-chavez (approved):

darora (approved):

PR #1021, opened 2024-07-02
chore: additional fallbacks for delegated-entry

pcnc:
* ensures the default entrypoint path runs if delegated entry execution fails

### DIFF
--- a/docker/all-in-one/entrypoint.sh
+++ b/docker/all-in-one/entrypoint.sh
@@ -218,12 +218,17 @@ function fetch_and_execute_delegated_payload {
 
   if [ ! -f $DELEGATED_ARCHIVE_PATH ]; then
     echo "No delegated payload found, bailing"
-    return
+    return 1
   fi
 
   # only extract a valid archive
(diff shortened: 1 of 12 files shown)


docker/all-in-one/entrypoint.sh
```diff
@@ -358,6 +367,12 @@ run_prelaunch_hooks
 
 if [ -n "${DELEGATED_INIT_LOCATION:-}" ]; then
   fetch_and_execute_delegated_payload
+  DELEGATED_INIT_RETURN_CODE=$?
+  if [ "$DELEGATED_INIT_RETURN_CODE" -ne 0 ]; then
```
thebengeu: How about exiting if return code is 0 and falling back to the code below, moved out of the `else`? To be DRY, and the execution time may still be relevant?

thebengeu (approved):

pashkinelfe (approved):

docker/all-in-one/entrypoint.sh
pcnc: Oh yes 👀 that'd be way better. Giving it a spin

PR #1025, opened 2024-07-03
Sam/nix and conventional consolidate

samrose:

This is a PR against already reviewed changes in  #1012 to consolidate ansible configurations, and use vars to switch which tasks are called, and avoid de-duplication of config code for different AMI outcomes. 

Goals of this Sub-PR:

-  Dedupe ansible ami building code
- All existing PR-triggered tests passing including testinfra tests for both versions, osquery permission checks, etc


pashkinelfe (approved):

ansible/tasks/internal/optimizations.yml
```diff
@@ -1,39 +1,29 @@
-- name: ensure services are stopped
-  community.general.snap:
```
darora, line 2: is this removal no longer needed?

ansible/files/postgresql_config/postgresql.service.j2
```diff
@@ -19,6 +19,7 @@ TimeoutStartSec=86400
 Restart=always
 RestartSec=5
 OOMScoreAdjust=-1000
+EnvironmentFile=-/etc/environment.d/postgresql.env
```
darora, line 22: No particular concerns, but what gets included in this file and what's populating it? Or is it there just in case it's needed?

darora (approved):

ansible/tasks/internal/optimizations.yml
samrose: @div yes, this was no longer present at this point in the build

ansible/files/postgresql_config/postgresql.service.j2
samrose: @darora when nix installs the postgresql package, nix doesn't perform a lot of the operations that the debian package performs. So here I am adding the following values so that postgres will know about locales

this file only contains

LOCALE_ARCHIVE=/usr/lib/locale/locale-archive
LANG="en_US.UTF-8"
LANGUAGE="en_US.UTF-8"
LC_ALL="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"

PR #1026, opened 2024-07-04
feat: patch supautils sha

J0:
Currently it seems like the [SHA256sum is inaccurate](https://github.com/supabase/postgres/actions/runs/9789146863/job/27028480566#step:6:1191) which is leading to a build error

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -85,7 +85,7 @@ wal2json_release: "2_5"
 wal2json_release_checksum: sha256:b516653575541cf221b99cf3f8be9b6821f6dbcfc125675c85f35090f824f00e
 
 supautils_release: "2.2.1"
-supautils_release_checksum: sha256:1a2d2b8fe604d38921ed9cf3a0d56dd142a274035d0dca17ad21cdc81ddd9569
+supautils_release_checksum: sha256:f1f33371390322ac830645b8b0b8e249cb8ca10b19fdeae917f383014ed01b5d
 
 pljava_release: master
 pljava_release_checksum: sha256:e99b1c52f7b57f64c8986fe6ea4a6cc09d78e779c1643db060d0ac66c93be8b6


soedirgo (approved):

darora:
What's changed? The version was changed [a month ago](https://github.com/supabase/postgres/pull/989/files), and we've had releases since then.

J0:
Not much has changed - I realized after opening the PR that this doesn't block the release. Figured we might as well patch though. Also fine to close.

darora:
That's actually bad as well, since a sha mismatch _should_ break the build and cause a release failure?

samrose (changes requested):
@darora @J0 

To address the checksum issue that Div raised in this PR

I think you are going to need something like this

```diff
diff --git a/Dockerfile b/Dockerfile
index 88f5575..b703703 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -824,7 +824,9 @@ RUN checkinstall -D --install=no --fstrans=no --backup=no --pakdir=/tmp --nodoc
 FROM base as supautils
 # Download package archive
 ARG supautils_release
-ADD "https://github.com/supabase/supautils/releases/download/v${supautils_release}/supautils-v${supautils_release}-pg${postgresql_major}-${TARGETARCH}-linux-gnu.deb" \
+ARG supautils_release_checksum
+ADD --checksum=${supautils_release_checksum} \
+    "https://github.com/supabase/supautils/releases/download/v${supautils_release}/supautils-v${supautils_release}-pg${postgresql_major}-${TARGETARCH}-linux-gnu.deb" \
     /tmp/supautils.deb
 
 ####################
```

J0:
> That's actually bad as well, since a sha mismatch should break the build and cause a release failure?

Yes, a sha mismatch should break the build and cause a release failure. I wasn't too sure if the resulting sha mismatch was intentional or something that had gone unnoticed for a long while though.

Let's close this PR as it sounds like the proper fix here would be to patch the Dockerfile rather than the Shasum

PR #1039, opened 2024-07-10
chore: generate pg_upgrade logs on successful run as well

pcnc:
* also fixes logs perms for Vector to consume

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -72,7 +72,8 @@ cleanup() {
     if [ -d "${MOUNT_POINT}/pgdata/pg_upgrade_output.d/" ]; then
         echo "Copying pg_upgrade output to /var/log"
         cp -R "${MOUNT_POINT}/pgdata/pg_upgrade_output.d/" /var/log/ || true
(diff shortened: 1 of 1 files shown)


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -72,6 +72,7 @@ cleanup() {
     if [ -d "${MOUNT_POINT}/pgdata/pg_upgrade_output.d/" ]; then
         echo "Copying pg_upgrade output to /var/log"
         cp -R "${MOUNT_POINT}/pgdata/pg_upgrade_output.d/" /var/log/ || true
+        chmod -R 777 /var/log/pg_upgrade_output.d/
```
Crispy1975: Does this need permissions to be this open?

Crispy1975 (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
pcnc: No, this was a subsequent commit meant for another branch :saddog: 

But ideally Vector would have access to it. Theoretically Vector is part of the postgres group, hence group read access should be enough - for some reason it doesn't really discover files. Tbd!

PR #1041, opened 2024-07-11
feat: install salt-minion on AMI

tomashley:

Feature: Installs salt-minion package


ansible/playbook.yml
```diff
@@ -136,6 +136,13 @@
       tags:
         - aws-only
 
+    # Install salt minion
+    # Only for AWS images
+    - name: install salt minion
```
darora: better spot to import these might be in ansible/tasks/setup-supabase-internal.yml

darora (approved):
@samrose fyi

ansible/playbook.yml
tomashley: Moved tasks

PR #1069, opened 2024-07-23
Upgrade pg_net and allow greater user-control over its parameters

TheOtherBrian1:

Modified supautils configs

pg_net variables and the log_lock_waits variable cannot be configured by users.

## What is the new behavior?

Allows users to configure pg_net variables (other than database):
- pg_net.ttl
- pg_net.batch_size

Gives users permission to log locks that last for more than 1s
- log_lock_waits

## Additional context


steve-chavez (approved):
LGTM

steve-chavez:
Hm, wait. Is this config applicable now that we use the Nix build?

olirice:
@samrose It doesn't look like the settings in supautils.conf are being applied (maybe only a local thing) but 
```
postgres=# set pg_net.ttl = '1s'
ERROR:  parameter "pg_net.ttl" cannot be changed now
```

- is that the correct location for supautils config?
- does it apply locally?

samrose:
> @samrose It doesn't look like the settings in supautils.conf are being applied (maybe only a local thing) but
> 
> ```
> postgres=# set pg_net.ttl = '1s'
> ERROR:  parameter "pg_net.ttl" cannot be changed now
> ```
> 
>  * is that the correct location for supautils config?
> * does it apply locally?

I think you are right that those settings have not been wired in yet on local testing 

I’ll be back in front of computer in a few and will get it working in local testing context

samrose:
We did recently merge in a pr to develop that correctly used all of our configurations and migrations 

https://github.com/supabase/postgres/pull/1130

soedirgo (approved):
Tested on staging

soedirgo:
cc @supabase/backend for a +1

dragarcia (approved):

PR #1072, opened 2024-07-24
feat: docker base psql15.6 img and release

samrose:
docker base psql15.6 img and releaseon merge to develop. nix creates a base version of the image which will be pushed to docker.io on each PR push, ovewriting the same tag. Dockerfile-156 will then use that tag as it's base to build the finale base supabase/postgres image for 15.6 (built from nix)


darora (approved):
Approving to unblock, given produced artifacts look OK

Issue #1078, opened 2024-07-31
Alternative strategy for Improved CI Build and Deployment times


DMills27:
# Chore

## Describe the chore

The current setup uses Github Actions to build and deploy from this repository. However, it suffers from issues like needing to rerun the entire pipeline to debug or test changes, leading to unnecessary wait times instead of caching unchanged results. Additionally, the pipelines themselves aren't necessarily the most optimised since they run on [vCPUs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#:~:text=GitHub%2Dhosted%20runner%20is%20a%20new%20virtual%20machine%20(VM)%20hosted%20by%20GitHub%20with%20the%20runner%20application%20and%20other%20tools%20preinstalled).

I suggest [Garnix](https://garnix.io/) as an alternative for (certain aspects of) the CI process. Garnix is a Nix based CI build tool. In order to use it one simply needs to have a `flake.nix` file in the repository under consideration. In making this suggestion I'm not arguing for overhauling the existing architecture but simply offering an alternative to the current build process that will work in a more performant manner while not intruding on what already exists, aimed mainly at those who wish to self-host. This comes with a number of benefits, such as:

- Blazingly fast build times through [higher levels of parallelism](https://gvolpe.com/blog/garnix/#:~:text=Garnix%20has%20more%20powerful%20machines%20than%20Github%2C%20and%20that%20in%20turn%2C%20enables%20higher%20parallelism.%20Additionally%2C%20rather%20than%20vCPUs%2C%20we%20use%20CPU%20shares) as well as making use of CPU shares, rather than vCPUs.
- Sharing the same cache among multiple targeted repositories. 
- Making it much easier to deploy sophisticated integration and E2E NixOS tests akin to [this one](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/mysql/mysql-replication.nix) which starts a primary MySQL instance and synchronises two secondary instances. It then simulates downtime by stopping one secondary instance and adding data to the primary, then restarts and re-synchronises the secondary instance. Tests like these are useful for testing the robustness of database migrations.
- Allowing one to seamlessly [debug failed builds](https://garnix.io/docs/deployments#debugging-failed-builds) since Garnix provisions a server upon each build, which one can SSH into and figure out what went wrong. Disclaimer: This is a paid feature though.

[...]

PR #1079, opened 2024-07-31
Adding Garnix Cache

DMills27:
Adding the Garnix cache to the `flake.nix` to enable faster CI build times, if a users opts to use Garnix, as described in [this issue](https://github.com/supabase/postgres/issues/1078)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,11 @@
 {
   description = "Prototype tooling for deploying PostgreSQL";
 
+  nixConfig = {
+    extra-substituters = ["https://cache.garnix.io"];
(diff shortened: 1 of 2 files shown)


olirice:
hi @DMills27 
thanks for opening

We'd like to keep this repo clear of any tooling and configuration that isn't used by our release pipeline. The nix overhaul is in the wake of steady complexity creep in our previous build process so we're taking a (probably overly) strict stance on it to avoid facing the same issue again in a few years.

Given how light weight your patch is I'd suggest maintaining a fork for your use and periodically rebasing

PR #1084, opened 2024-08-02
Sam/docker int test

samrose:

This PR is introducing a simple integration test after the build of the image, and prior to the push of the image.

Before it is merged, it needs the `-dev` suffix removed from the docker image tag


samrose:
deprecated closing

PR #1077, opened 2024-07-31
chore: bump gotrue version to v2.157.1

kangmingtay:
## What kind of change does this PR introduce?
* See [changelog](https://github.com/supabase/auth/compare/v2.156.0...v2.157.1) - this version of gotrue introduces the changes made to support asymmetric JWTs for projects 

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

PR #1086, opened 2024-08-05
fix: update auth to v2.158.1

J0:
Prior art: https://github.com/supabase/postgres/pull/1077

Changelog: https://github.com/supabase/auth/compare/v2.156.0...v2.158.1


olirice (approved):
Approved it, but (as you mentioned) the versions will need to be bumped again after the pg_graphql pr merges

delgado3d (approved):

PR #1104, opened 2024-08-09
fix: wrappers nix-based pg_upgrade

pcnc:
* The lookup path for nix-based binaries isn't necessarily the obvious one
* This ensures that a wrappers lib file named as expected to by PG exists on all existing paths


darora (approved):
Hard to review this well. Need to move the upgrade scripts to something better tested. 

Approving based on your validation runs.

PR #1105, opened 2024-08-09
Sam/pgroonga deps

samrose:

a comment on a PR https://github.com/supabase/postgres/pull/462#issuecomment-2270662752 reported that in our 15.6 release, `TokenMecab` was no longer working on the `pgroonga` extension.

This PR restores that Tokenizer capability to the proonga extension, and adds the following test below, which will run on every PR going forward. 

```sql
-- File: 0005-test_pgroonga_revised.sql

begin;
    -- Plan for 3 tests: extension, table, and index
    select plan(3);
    
    -- Create the PGroonga extension
    create extension if not exists pgroonga;
    
    -- -- Test 1: Check if PGroonga extension exists
    select has_extension('pgroonga', 'The pgroonga extension should exist.');
    
    -- Create the table
    create table notes(
        id integer primary key,
        content text
    );
    
    -- Test 2: Check if the table was created
    SELECT has_table('public', 'notes', 'The notes table should exist.');    
    -- Create the PGroonga index
    CREATE INDEX pgroonga_content_index
            ON notes
         USING pgroonga (content)
          WITH (tokenizer='TokenMecab');
    
    -- -- Test 3: Check if the index was created
    SELECT has_index('public', 'notes', 'pgroonga_content_index', 'The pgroonga_content_index should exist.');
    
    -- -- Cleanup (this won't affect the test results as they've already been checked)
    DROP INDEX IF EXISTS pgroonga_content_index;
    DROP TABLE IF EXISTS notes;
    
    -- Finish the test plan
    select * from finish();
rollback;
```

### DIFF
--- a/nix/ext/pgroonga.nix
+++ b/nix/ext/pgroonga.nix
@@ -11,7 +11,7 @@ stdenv.mkDerivation rec {
   };
   nativeBuildInputs = [ pkg-config makeWrapper ];
   buildInputs = [ postgresql msgpack-c supabase-groonga mecab ];
-  
+  propagatedBuildInputs = [ supabase-groonga ];
   configureFlags = [
     "--with-mecab=${mecab}"
     "--enable-mecab"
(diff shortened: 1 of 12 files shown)


nix/ext/pgroonga.nix
```diff
@@ -1,30 +1,63 @@
...
+
+    # Modify the main PGroonga SQL file to include MeCab plugin registration
+    for sql_file in $out/share/postgresql/extension/pgroonga-*.sql; do
+      echo "SELECT pgroonga_command('plugin_register ${supabase-groonga}/lib/groonga/plugins/tokenizers/mecab.so');" >> $sql_file
+    done
+
+    cat << EOF > $out/share/postgresql/extension/pgroonga_set_paths.sql
+    DO \$\$
+    BEGIN
+      SET pgroonga.log_path TO current_setting('data_directory') || '/groonga.log';
```
darora: Is there an existing log file? Is something set up to handle rotation for it?

darora (approved):

nix/ext/pgroonga.nix
samrose: @darora good point, I don't think there is an existing logfile in production.

I actually turned this on for debugging purposes, and will definitely remove it before merge

samrose: If we decided we wanted that on, there may be a better way to do it post-build, but likely we don't need that logging is my guess.

samrose: @div I remove this and the other sql patching

olirice (approved):

olirice:
@samrose lgtm, feel free to merge once the conflicts are resolved

PR #1109, opened 2024-08-11
fix: it is required to run this at the end of AMI build

samrose:
to remove deps in the store not used by the installed pkg

We have size constraints and need to make sure any extra files introduced in the build process are removed before the AMI is deployed.


ansible/playbook.yml
```diff
@@ -206,3 +206,9 @@
       shell: |
         sudo -u ubuntu bash -c ". /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix profile remove osquery"
       when: stage2_nix
+
+    - name: nix collect garbage
+      become: yes
+      shell: |
+        sudo -u ubuntu bash -c ". /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix-collect-garbage -d"
+      when: stage2_nix
```
pcnc: Missing new line here

pcnc (approved):

ansible/playbook.yml
samrose: Thank you fixed with https://github.com/supabase/postgres/pull/1109/commits/5ad1be53eff0a814a383f2a299a1c3321d3ff82a

PR #1125, opened 2024-08-15
feat: change bootstrap user to `supabase_admin` upon upgrade

soedirgo:
https://www.notion.so/supabase/Change-bootstrap-user-to-supabase_admin-during-pg_upgrade-fa83a2da23994591af8d075edebed6de?pvs=4

This unblocks PG16, which disallows the bootstrap user to be downgraded to a non-superuser.


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -375,17 +382,26 @@ function initiate_upgrade {
...
+        su -c "$PGBINNEW/initdb -L $PGSHARENEW -D $PGDATANEW/ --username=supabase_admin" -s "$SHELL" postgres
     fi
 
+    # This line avoids the need to supply the supabase_admin password on the old
+    # instance, since pg_upgrade connects to the db as supabase_admin using unix
+    # sockets, which is gated behind scram-sha-256 per pg_hba.conf.j2. The new
+    # instance is unaffected.
+    echo "local all supabase_admin trust
+$(cat /etc/postgresql/pg_hba.conf)" > /etc/postgresql/pg_hba.conf
+    run_sql -c "select pg_reload_conf();"
```
pcnc: In case of an upgrade failure + retry, this will likely generate duplicate entries - we should validate that the rule doesn't exist before adding it to the file.

soedirgo: Updated 👍

pcnc (approved):
Approving - to be merged on Sept 11th in UTC morning

common-nix.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "15.6.1.118"
+postgres-version = "15.6.1.119"
```
soedirgo:
```suggestion
postgres-version = "15.6.1.120"
```

PR #1128, opened 2024-08-15
fix: generate sql migrations for tag versions

samrose:

this package needs to be responsible going forward for generating the range of sql files for migrations


samrose:
Turns out this was not needed

PR #1130, opened 2024-08-15
nix run .#start server 15 local dev full config

samrose:

This PR fixes our `nix run .#start-server 15` function to use configurations found in our ansible files as a basis for configuring postgresql to run locally

Once you run `nix run .#start-server 15` you can then run `nix run .#start-client-and-migrate 15` and you should have a working environment that matches very closely with our AMI release configuration (however it is using the versions in your current branch of your current repo for the .#psql_15/bin bundle).

The logging config in our AMI is commented out so that logs will stream to stdout/stderr for local development

This has been tested on aarch64-linux and aarch64-darwin + x86_64-linux


logfile
olirice, line 1: Does this logfile need to be committed?

olirice (approved):

logfile
pcnc, line 1: Approved - also curious about this logfile 👀

pcnc (approved):

logfile
samrose: thank you removed

samrose: Thanks I removed that

PR #1152, opened 2024-08-21
feat: enable webhooks by default

sweatybridge:
SQL migrated over from cli / api https://github.com/supabase/cli/blob/develop/internal/db/start/templates/schema.sql#L26
~With improved retry support from @mansueli: https://github.com/supabase/infrastructure/pull/16831~

Solves an immediate pain point with enabling webhooks. More context: https://supabase.slack.com/archives/C07E5GFAHTM/p1724229124279079

Upgrade path for existing projects can be done via ansible rollout as the migration is idempotent.

### DIFF
--- a/ansible/tasks/setup-postgres.yml
+++ b/ansible/tasks/setup-postgres.yml
@@ -239,7 +239,7 @@
   become_user: postgres
   shell: |
     source /var/lib/postgresql/.bashrc
-    /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data start
+    /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data start -o "-c shared_preload_libraries='pg_net'"
   environment:
     LANG: en_US.UTF-8
     LANGUAGE: en_US.UTF-8
--- a/ebssurrogate/files/unit-tests/unit-test-01.sql
+++ b/ebssurrogate/files/unit-tests/unit-test-01.sql
@@ -12,6 +12,7 @@ SELECT extensions_are(
     'pg_graphql',
     'pgcrypto',
     'pgjwt',
+    'pg_net',
     'uuid-ossp',
     'supabase_vault'
      ]
--- a/migrations/db/migrations/20240822021428_enable_webhooks_by_default.sql
+++ b/migrations/db/migrations/20240822021428_enable_webhooks_by_default.sql
@@ -20,8 +20,7 @@ CREATE TABLE IF NOT EXISTS supabase_functions.migrations (
 -- Initial supabase_functions migration
 INSERT INTO supabase_functions.migrations (version) VALUES
   ('initial'),
-  ('20210809183423_update_grants'),
-  ('20240125163000_add_retry_to_http_request')
+  ('20210809183423_update_grants')
 ON CONFLICT DO NOTHING;
 
 -- supabase_functions.hooks definition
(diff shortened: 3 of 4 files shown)


soedirgo:
Please omit the retry changes for now, the Postgres team doesn't have the bandwidth to fix it if it goes south.

@steve-chavez would it be an issue to enable Database Webhooks by default once we package it as an extension?

steve-chavez:
> @steve-chavez would it be an issue to enable Database Webhooks by default once we package it as an extension?

I'm a bit confused about this change. So this means that the `supabase_functions` schema will be created by default and not with a manual step on the dashboard?

If so, is certainly an improvement, but ideally we package `supabase_functions` as an extension and not a "migration". This is so we can control the SQL for debugging and upgrading. So far this has been impossible to do for me.

So could we move it to `nix/ext/supabase_functions`, make it a extension (v0.1) and make it depend on pg_net?

This would be a huge improvement.

steve-chavez:
> So could we move it to nix/ext/supabase_functions, make it a extension (v0.1) and make it depend on pg_net?

If the above can be done without a breaking change later, I can help with it after I finish https://github.com/supabase/pg_net/pull/139.

sweatybridge:
Yes, this removes the manual step from enabling on dashboard.

Enabling webhooks is also currently irreversible since we don't support disabling from dashboard. That's why I think it's ok to have it as a migration since we have to deal with the upgrade path for existing projects anyway.

But I agree a separate extension is better. I believe the `supabase_functions` schema is currently locked down to supabase_admin so it's safe to switch without breaking.

This webhook also creates the `supabase_functions_admin` role. Do we want to create it in the extension too?

mansueli:
+1 for moving into an extension

Having it as an extension will also make it easier to release updates later on (like the retry feature)

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

steve-chavez:
> I believe the supabase_functions schema is currently locked down to supabase_admin so it's safe to switch without breaking.

Ok cool. Then this should be good to go.

> This webhook also creates the supabase_functions_admin role. Do we want to create it in the extension too?

Haven't thought much about this, maybe the role creation and the privileges should remain as migration. I don't recall seeing a pg extension that creates roles (since they're global objects). But is something for later.

steve-chavez (approved):

soedirgo:
> pg extension that creates roles

pgsodium does this, but yeah agree that creating global objects in an extension is a bit awkward

ansible/tasks/setup-postgres.yml
```diff
@@ -239,7 +239,7 @@
   become_user: postgres
   shell: |
     source /var/lib/postgresql/.bashrc
-    /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data start
+    /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data start -o "-c shared_preload_libraries='pg_net'"
```
darora, line 242: Not sure why this is needed? Isn't pg_net already included?

darora (commented):
Isn't pg_net still seeing segfaults? Do we really want to enable it by default?

steve-chavez:
> Isn't pg_net still seeing segfaults? Do we really want to enable it by default?

No more segfaults since https://github.com/supabase/pg_net/releases/tag/v0.9.3, those were fixed on https://github.com/supabase/pg_net/pull/136.

ansible/tasks/setup-postgres.yml
sweatybridge: > psql:/tmp/migrations/db/migrations/20240822021428_enable_webhooks_by_default.sql:4: ERROR:  pg_net is not in shared_preload_libraries

pg_net is not included in shared_preload_libraries for this code path, hence resulting in an error https://github.com/supabase/postgres/actions/runs/10495775396/job/29074946651

sweatybridge: Are we happy to merge this one?

steve-chavez:
Before the main concern with this is that `pg_net` backround worker would work even if the user wasn't using it.

Nowadays that's no longer a concern since https://github.com/supabase/pg_net/pull/199, `pg_net` worker will just sleep if there's no activity on its queue.

samrose:
closing, re-open if needed

PR #1173, opened 2024-08-23
Sam/migration testing

samrose:

Migration testing that aligns with the workflow [here](https://github.com/supabase/postgres/blob/develop/.github/workflows/test-pg-upgrade.yml) but able to run locally against 2 versions of pg (currently just 15 -> 15)

This test is run with `nix run .#migration-test <git-tag-only> gitsha-or-gittag> pg_upgrade`

We'll deprecate the "major version" as it is too vague to meaningful for the use cases for this tool.

This PR is WIP/Draft at this point, but prior to merging, we'll create specific tags for building and testing previous versions, and retain those in the git repo, and implement the support for gitsha or tag.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -62,6 +62,16 @@
 
         sfcgal = pkgs.callPackage ./nix/ext/sfcgal/sfcgal.nix { };
         pg_regress = pkgs.callPackage ./nix/ext/pg_regress.nix { };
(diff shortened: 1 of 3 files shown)


samrose:
closing in favor of https://github.com/supabase/postgres/pull/1183

PR #1174, opened 2024-08-26
fix: only keep 1mil pg_cron job run details during pg_upgrade

pcnc:


soedirgo (approved):

darora (approved):
We should document this at https://supabase.com/docs/guides/platform/migrating-and-upgrading-projects . 

_Maybe_ also create a changelog entry?

PR #1177, opened 2024-09-02
feat: configure Envoy to compress additional response types

darora:
Adds a number of Postgrest response types for compression by Envoy.

### DIFF
--- a/ansible/files/envoy_config/lds.yaml
+++ b/ansible/files/envoy_config/lds.yaml
@@ -103,7 +103,78 @@ resources:
                             request_handle
                               :headers()
                               :replace(":path", path:gsub("&=[^&]*", ""):gsub("?=[^&]*$", ""):gsub("?=[^&]*&", "?"))
+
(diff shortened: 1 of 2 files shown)


dragarcia (approved):

istarkov:
I can confirm that it works now, given that these changes have been applied to our Supabase project. 

However, I noticed that all other responses previously used Brotli compression by default, but now they are using gzip. 
[Support Ticket ID: 3125882163 / Discussion](https://github.com/orgs/supabase/discussions/28757)

_It’s not a big deal, but Brotli is more commonly used than gzip in our infrastructure._
_Anyway, we’ve excluded the use of .single() in our code, so we’ll be able to compare the difference between gzip and Brotli using our stats, if it exists and is statistically significant._

darora:
We're looking into updating out Cloudflare configuration to prefer using Brotli. In the meantime, we'll be merging this in to have the docker images updated.

PR #1179, opened 2024-09-02
feat: bump auth to v2.160.0

hf:
Bumps Auth to v2.160.0.


common.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "15.1.1.90"
```
pcnc: Update `nix-common.vars.pkr.hcl` instead please

pcnc (changes requested):

J0 (dismissed):

common.vars.pkr.hcl
```diff
@@ -1 +1 @@
...
+postgres-version = "15.1.1.91"
```
darora: No longer in use - update the common-nix file instead

darora (approved):

pcnc (approved):

PR #1187, opened 2024-09-06
fix: continue running migrations when auth.uid/auth.role/auth.email are missing in the database

pcnc:


migrations/db/migrations/20211124212715_update-auth-owner.sql
```diff
@@ -1,8 +1,12 @@
 -- migrate:up
 
 -- update owner for auth.uid, auth.role and auth.email functions
-ALTER FUNCTION auth.uid owner to supabase_auth_admin;
-ALTER FUNCTION auth.role owner to supabase_auth_admin;
-ALTER FUNCTION auth.email owner to supabase_auth_admin;
-
+DO $$
+BEGIN
+    ALTER FUNCTION auth.uid owner to supabase_auth_admin;
```
darora, line 6: Should we have each statement in its own block? So that even if e.g. `auth.uid` is missing, the other two still get reassigned?

darora (approved):

migrations/db/migrations/20211124212715_update-auth-owner.sql
pcnc: Done! Will give it a test on staging, then merge

PR #1189, opened 2024-09-09
fix: run nix-gc before pg_upgrade

pcnc:


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -277,16 +277,23 @@ function initiate_upgrade {
         NIX_FLAKE_VERSION=$(cat "$PG_UPGRADE_BIN_DIR/nix_flake_version")
 
         if [ "$IS_NIX_BASED_SYSTEM" = "false" ]; then
-            echo "1.1. Nix is not installed; installing."
-
-            curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install --no-confirm \
-            --extra-conf "substituters = https://cache.nixos.org https://nix-postgres-artifacts.s3.amazonaws.com" \
-            --extra-conf "trusted-public-keys = nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI=% cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="
+            if [ ! -f /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh ]; then
```
thebengeu, line 280: Nit: `&&` with above into one condition?

thebengeu (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
pcnc: That would unfortunately never run :[ 

The `nix` binary is _generally_ located in `/root/.nix-profile/bin` - with the path provided by sourcing `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`

In truth, if `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` doesn't exist, neither will `/root/.nix-profile/bin/nix`, which makes `! command -v nix > /dev/null` kind of superfluous.

Will remove the secondary check once we figure out the avenue of hosting the installer ourselves, as I'm not sure how stable the `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` path is at the moment

darora (approved):

PR #1190, opened 2024-09-10
Sam/pg 15 8

samrose:

- bumping the git revision of nixpkgs to the latest rev or `nixpkgs-unstable`
- Bringing packaging of postgresql and a dependency into our codebase, so that we can control the build, dependencies, and minor version changes independent of what is happening in nixpkgs
- for running on darwin, needed some env vars for PGPORT during the build phase
- introducing postgresql_16, but not yet incorporating it into the installables available from the flake
- advanced postgresql minor version to "15.8"

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -40,6 +40,7 @@ jobs:
           role-to-assume: ${{ secrets.DEV_AWS_ROLE }}
           aws-region: "us-east-1"
           output-credentials: true
+          role-duration-seconds: 7200
       - name: write secret key
         # use python so we don't interpolate the secret into the workflow logs, in case of bugs
         run: |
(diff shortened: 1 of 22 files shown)


olirice:
sorry in advance for the hassle but could we do this in 3 separate PRs?
- PR1 - Bringing packaging of postgresql and a dependency into our codebase
- PR2 - advanced postgresql minor version to "15.8"
- PR3 - introducing postgresql_16, but not yet incorporating it into the installables available from the flake

at this phase we're frequently using your PRs to identify where to look to handle these types of tasks and having them separate would make that easier so e.g. any on the team would know how to add pg17

samrose:
Yes of course seems well reasoned to me. I’ll close this and follow up with those PR

PR #1191, opened 2024-09-11
feat(pg_upgrade): use self-hosted nix installer if available; move files out of /tmp

pcnc:
* Saves us from headaches due to external outages, or mismatched installer issues
* Fixes vacuumdb not being run as superuser post-bootstrap user switch


samrose (approved):
Seems like a reasonable approach to me. 

The one caveat is that there can be security issues with specific versions of nix, which tend to be fixed and then distributed with the installer by DS.

So, we should definitely have a process to keep out self-hosted installer up to date on a weekly basis or so, outside of the operation it is being used for.

pcnc:
Moving into draft as it still needs some testing

thebengeu (approved):

PR #1195, opened 2024-09-11
feat: bringing postgres package into our package set

samrose:

This PR introduces the following changes 

1. Bring postgresql package into our codebase instead of using upstream package
2. small cleanup of the output of tests to reflect what is currently working when running `nix flake check -L` 

The version of postgresql for 15 remains unchanged here, and no other major version is yet introduced. This change also does not update the version of nixpkgs used. Those changes will come in subsequent pull requests.

### DIFF
--- a/nix/postgresql/generic.nix
+++ b/nix/postgresql/generic.nix
@@ -1,0 +1,309 @@
+let
+
+  generic =
+      # adapted from the nixpkgs postgresql package
+      # dependencies
+      { stdenv, lib, fetchurl, fetchpatch, makeWrapper
+      , glibc, zlib, readline, openssl, icu, lz4, zstd, systemd, libossp_uuid
+      , pkg-config, libxml2, tzdata, libkrb5, substituteAll, darwin
+      , linux-pam
+
+      # This is important to obtain a version of `libpq` that does not depend on systemd.
+      , systemdSupport ? lib.meta.availableOn stdenv.hostPlatform systemd && !stdenv.hostPlatform.isStatic
+      , enableSystemd ? null
+      , gssSupport ? with stdenv.hostPlatform; !isWindows && !isStatic
+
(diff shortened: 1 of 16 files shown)


nix/postgresql/generic.nix
```diff
@@ -0,0 +1,309 @@
...
+    env.NIX_CFLAGS_COMPILE = lib.optionalString (olderThan "13") "-I${libxml2.dev}/include/libxml2";
+
+    configureFlags = [
+      "--with-openssl"
+      "--with-libxml"
+      "--with-icu"
+      "--sysconfdir=/etc"
+      "--libdir=$(lib)/lib"
+      "--with-system-tzdata=${tzdata}/share/zoneinfo"
+      "--enable-debug"
```
pashkinelfe, line 102: I'd drop --enable-debug and use CFLAGS=-g3 instead like in https://github.com/supabase/postgres/commit/fdaee4b59f351ca6e50fa4cdfb0bb655dae7b3db

As per  
https://wiki.postgresql.org/wiki/Developer_FAQ  --enable-debug will also set less speed-optimal optimization i.e. -Og CFLAGS (and also -fno-omit-frame-pointer) that could decrease performance compared to default -02. I think for production system -g3 is better

olirice (approved):

nix/postgresql/generic.nix
samrose: @pashkinelfe ok I will try that thanks!

darora (approved):

nix/postgresql/generic.nix
samrose: @pashkinelfe I am going to do this for sure, but going to isolate it to another PR so that we can keep the rest of this work moving forward, and so that I can assure that we build a separateDebugInfo with this outcome too.

PR #1200, opened 2024-09-12
chore: create an isolated output for postgres, to expose debug symbols

samrose:
This is a small PR to expose only the underlying postgresql_15 package. It turns out that postgresql in nixpkgs are already built with debug symbols enabled, but in a separate directory which is considered the best practice.

To access these symbols you can run

```bash
nix build .#postgresql_15.debug && nix derivation show .#postgresql_15.debug | jq -r '.[].env.debug'
/nix/store/pxx1mdvxdq932km87a8d8q4nqvadpspy-postgresql-15.6-debug
```

Then use `/nix/store/<yourpath>-postgresql-15.6-debug`  in your debugger and these will be the corresponding debug symbols for the current version of postgresql_15 

this same method will work going forward with our own packaging of postgresql as well.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -273,6 +273,7 @@
           sfcgal = sfcgal;
           pg_regress = pg_regress;
           pg_prove = pkgs.perlPackages.TAPParserSourceHandlerpgTAP;
+          postgresql_15 = pkgs.postgresql_15;
           # Start a version of the server.
           start-server =
             let
--- a/flake.nix
+++ b/flake.nix
@@ -274,6 +274,29 @@
           pg_regress = pg_regress;
           pg_prove = pkgs.perlPackages.TAPParserSourceHandlerpgTAP;
           postgresql_15 = pkgs.postgresql_15;
+
+          postgresql_15_src = pkgs.stdenv.mkDerivation {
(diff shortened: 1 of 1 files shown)


pashkinelfe (approved):

darora (approved):

steve-chavez:
With this is also possible to do `nix build .#postgresql_15_src`, and then get the source code:
```
ls -la result
lrwxrwxrwx 1 user user 66 Sep 12 16:43 result -> /nix/store/bgf6a5l4hpsf19c3h4jfx2vykxmysy58-postgresql-15-src-15.6

ls /nix/store/bgf6a5l4hpsf19c3h4jfx2vykxmysy58-postgresql-15-src-15.6/src/
/nix/store/bgf6a5l4hpsf19c3h4jfx2vykxmysy58-postgresql-15-src-15.6/src/:
backend  common      fe_utils  interfaces  Makefile.global.in  Makefile.shlib  pl    template  timezone  tutorial
bin      DEVELOPERS  include   Makefile    makefiles           nls-global.mk   port  test      tools
```
I haven't tried it with GDB yet, but it's a start.

steve-chavez (approved):

PR #1220, opened 2024-09-18
fix: ami runner runs out of space with larger builds

samrose:

this builder ephemeral ec2 instance had been operating on the default size of 8GB. This is typically enough, but from time to time, builds may increase as we advance items like minor version of postgres, etc. this can require a rebuild of all extensions that use postgresql as a buildInput/nativeBuildInput. So, we'll increase the diskspace to assure builds can complete.


samrose:
closing in favor of https://github.com/supabase/postgres/pull/1231

PR #1226, opened 2024-09-23
Sam/plv8 version update

samrose:

This PR updates plv8 to 3.1.10 which is the last version of the project that still builds plcoffee, plls, plv8 and also is has support for pg16 (although this still needs to be tested and we may need to make contributions to plv8 or patches on our end to round out support after testing)

Notes: we need to move to plv8 3.2.x to support using v8 sourced from nodejs 20_x will be in a future PR. 

This PR has been tested on `aarch64-darwin x86_64-linux aarch-64-linux`


olirice (approved):

darora (approved):
@pcnc do we need any changes to upgrades to account for a plv8 version update? Or will users automatically be prompted to drop their plv8 objects anyway

PR #1232, opened 2024-09-24
Update timescale derivation to disallow non-apache configs

olirice:
Updates the timescaledb extension derivation such that it always builds the Apache


samrose (approved):
Thanks, looks great!

pcnc (approved):

darora:
We really need to start bumping the version file with these changes

PR #1267, opened 2024-10-08
fix: reformat ec2 cleanup commands

samrose:

This kind of failure of testinfra run was causing instances to remain running instead of being cleaned up

https://github.com/supabase/postgres/actions/runs/11178616589/job/31076546301#step:12:6

reformatted command to avoid the error


thebengeu (approved):

darora (approved):

samrose:
Added -r option to xargs to prevent it from running if there's no input (i.e., no instances to terminate).

PR #1266, opened 2024-10-07
Sam/15 16 nix package, build and test of 15.8 and 16.3

samrose:
## What kind of change does this PR introduce?

This makes all the necessary changes to package, build and test in the nix/pg_regress context our psql bundles with 15.8 and 16.3

We attempt to change no versions, and as much as possible to keep 1 pg_regress test suite, and one set of extensions and wrappers. 

There will be a follow up PR to address the packer/ansible/gh action work for integration tests and releasing images.

It includes the changes to packer/ansible/gh actions to allow for a matrix build of pg versions on release.

## Significant changes

This PR moves the source of truth for data about postgres major and semver back to ansible/vars.yml, which are then read at time of release in the ami-release gh action. The _**common-vars.nix.yml file is no longer under source control**_, but instead is generated duing the ci run for ami release or testinfra runs. The version is now controlled in this allows us to avoid making too many changes to our release and test tooling for AMI)

To bump the versions of postgres it is proposed to update these values that are now in `anisble/vars.yml`

These versions cover both release and testinfra

```yml
postgres_major:
  - "15"
  - "16"

# Full version strings for each major version
postgres_release:
  postgres15: "15.8.1.003-staging-5"
  postgres16: "16.3.1.000-staging-5"
```

PR #1268, opened 2024-10-09
pg 15 and 16 packer/ansible/ghactions

samrose:

This PR is a child PR of https://github.com/supabase/postgres/pull/1266

It includes the changes to packer/ansible/gh actions to allow for a matrix build of pg versions on release.

## Significant changes

This PR moves the source of truth for data about postgres major and semver back to ansible/vars.yml, which are then read at time of release in the ami-release gh action. The _**common-vars.nix.yml file is no longer under source control**_, but instead is generated duing the ci run for ami release or testinfra runs. The version is now controlled in this allows us to avoid making too many changes to our release and test tooling for AMI)

To bump the versions of postgres it is proposed to update these values that are now in `anisble/vars.yml`

These versions cover both release and testinfra

```yml
postgres_major:
  - "15"
  - "16"

postgres_release:
  postgres15: "15.8.1.003-staging-5"
  postgres16: "16.3.1.000-staging-5"
```


.github/workflows/ami-release-nix.yml
```diff
@@ -3,17 +3,37 @@ name: Release AMI Nix
 on:
   push:
     branches:
-      - develop
-      - release/*
+      #- develop
+      #- release/*
+      - sam/15-16-ghactions
```
soedirgo: Needs reverting?

samrose: This is a pr to another branch where that is commented out. My intention is to merge this to that branch, once that branch pr is approved I would revert this change to trigger release

samrose: @soedirgo okto merge to parent pr?

soedirgo: Yeah parent PR looks OK, but need a +1 from infra

PR #1272, opened 2024-10-15
Allow user defined role for start-client

olirice:
Add the option to pass a `-u, --user` flag to `nix run .#start-client` to set which role to start the connection with.

By default it continues to use `postgres` but you can now also use `supabase_admin` in case you're debugging a permissions issue or otherwise need security to get out of the way while working


samrose (approved):

pcnc (approved):

samrose:
This one is likely blocked by https://github.com/supabase/pg_net/pull/156 as far as successful build on macos goes

samrose (approved):

PR #1280, opened 2024-10-17
Update supautils.conf.j2 | allow users to log temp files

TheOtherBrian1:

Allow users to log temp files 

Users cannot reconfigure the `log_temp_files` variable

## What is the new behavior?

Users can now log temp files, which is necessary for reconfiguring `work_mem`

## Additional context

I was at [pgconf_nyc](https://2024.pgconf.nyc/tickets/) a few weeks ago and I spoke to Christophe Pettus, a Postgres Expert at [PGX](https://pgexperts.com/). He told me that its best practice to log all temp_file creation as it is one of the best ways to know when to modify `work_mem`. He made a [blog post](https://thebuild.com/blog/2023/03/13/everything-you-know-about-setting-work_mem-is-wrong/) that expounds on his reasoning.

It just so happens that I came across a [ticket](https://app.hubspot.com/live-messages/19953346/inbox/8209868365#email) creating a bunch of GB size temp files and it caused their disk to expand.

Because we now charge for disk provisioned vs. disk used, the "usage" warnings are making this problem much more noticeable to users. Because they're complaining more, there's a need to give them tools to self-manage


ansible/files/postgresql_config/supautils.conf.j2
```diff
@@ -7,6 +7,6 @@ supautils.privileged_extensions = 'address_standardizer, address_standardizer_da
 supautils.privileged_extensions_custom_scripts_path = '/etc/postgresql-custom/extension-custom-scripts'
 supautils.privileged_extensions_superuser = 'supabase_admin'
 supautils.privileged_role = 'postgres'
-supautils.privileged_role_allowed_configs = 'auto_explain.log_min_duration, auto_explain.log_nested_statements, log_lock_waits, log_min_messages, pg_net.batch_size, pg_net.ttl, pgaudit.log, pgaudit.log_catalog, pgaudit.log_client, pgaudit.log_level, pgaudit.log_relation, pgaudit.log_rows, pgaudit.log_statement, pgaudit.log_statement_once, pgaudit.role, pgrst.*, plan_filter.*, safeupdate.enabled, session_replication_role, track_io_timing'
+supautils.privileged_role_allowed_configs = 'auto_explain.log_min_duration, auto_explain.log_nested_statements, log_lock_waits, log_min_messages, log_temp_files, pg_net.batch_size, pg_net.ttl, pgaudit.log, pgaudit.log_catalog, pgaudit.log_client, pgaudit.log_level, pgaudit.log_relation, pgaudit.log_rows, pgaudit.log_statement, pgaudit.log_statement_once, pgaudit.role, pgrst.*, plan_filter.*, safeupdate.enabled, session_replication_role, track_io_timing'
```
darora, line 11: We really need to push these to be handled via https://supabase.com/docs/guides/platform/custom-postgres-config , rather than supautils

If you update them using supautils, they won't get replicated to replicas (and in the future, failover primaries).

darora (approved):

ansible/files/postgresql_config/supautils.conf.j2
TheOtherBrian1: Some of our settings `work_mem` and `max_parallel_workers` can be set through both the CLI and just generic SQL. For the time being, this is a fast way to make configs accessible. 

> " they won't get replicated to replicas (and in the future, failover primaries)."

Is this true only for `supautils` enabled settings or for any change outside of the CLI? Either or, a note should be added to the [customizing postgres docs](https://supabase.com/docs/guides/database/custom-postgres-config)

TheOtherBrian1: I changed pg_audit's settings, which is handled by supautils. It captured data from both the primary and replica. Has there been any recent updates?

PR #1310, opened 2024-11-05
chore: adding support for x86_64-darwin

samrose:

Minimum basis for supporting x86_64-darwin build

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -42,7 +42,7 @@ jobs:
           role-to-assume: ${{ secrets.DEV_AWS_ROLE }}
           aws-region: "us-east-1"
           output-credentials: true
-          role-duration-seconds: 8200
+          role-duration-seconds: 7500
       - name: write secret key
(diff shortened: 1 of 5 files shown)


darora (approved):
Do we need to build the artifacts as part of the CI builds? Having support for being able to do it sounds good though

samrose:
> Do we need to build the artifacts as part of the CI builds? Having support for being able to do it sounds good though

It does help us to run these builds/checks. Usually a true full build only happens when something has changed in the source, otherwise it is passed over. Then, when some source is changed, the build for that part will happen, the results will be cached into our binary cache, and so then 90% of the rest of the future ci runs will be able to deterministically refer to and use the cached version. In the case of this build, adding a new system did trigger a full build, but that is also what we needed to test, and achieve.

So it is typical practice in nix based systems to run both builds and checks like this in ci. It should actually trigger a full build only in the small minority of cases. And in those cases, you almost always want to test the build, and cache it if it succeeds.

PR #1313, opened 2024-11-06
fix: use `sudo` on `apt-get` commands

soedirgo:
https://supabase.slack.com/archives/C01D6TWFFFW/p1730877098588249?thread_ts=1730876824.155519&cid=C01D6TWFFFW

### DIFF
--- a/scripts/nix-provision.sh
+++ b/scripts/nix-provision.sh
@@ -39,8 +39,8 @@ EOF
 }
 
 function cleanup_packages {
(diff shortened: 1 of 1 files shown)


delgado3d (approved):

delgado3d:
Bump version? So we have a release?

soedirgo:
No need I think - just need to bump when someone needs to cut a release

PR #1315, opened 2024-11-08
Update pg net to 0.13.0 on pg 15.6 branch

olirice:
Update pg net to 0.13.0 on pg 15.6 branch. change is already on develop


delgado3d (approved):
bump version in common-nix.vars.pkr.hcl for ami release

PR #1325, opened 2024-11-19
feat: separate envoy lds configs for self-hosting and supabase use-cases

hf:
Separates Envoy's `lds.yaml` file for two different use cases: self-hosting and for Supabase internal use.

For self-hosting:

- File remains to be called `lds.yaml`
- Includes only self-hosing concepts such as fixed API keys
- Control which `lds.yaml` file is used with the `envoy_lds` Docker build arg

For Supabase:

- File is called `lds.supabase.yaml`
- Is replaced into `/etc/envoy/lds.yaml` with an Ansible task part of `ansible/tasks/setup-supabase-internal.yml`
- For Fly, replaced using the `envoy_lds` Docker build arg
- In follow up PR it will contain an origin protection key which is not something available in the Docker image for self-hosting
- Future: once origin protection keys are rolled out and new API keys are in place, Envoy will no longer check any API keys

### DIFF
--- a/ansible/files/envoy_config/lds.supabase.yaml
+++ b/ansible/files/envoy_config/lds.supabase.yaml
@@ -1,0 +1,436 @@
+resources:
+  - '@type': type.googleapis.com/envoy.config.listener.v3.Listener
+    name: http_listener
+    address:
+      socket_address:
+        address: '::'
+        port_value: 80
+        ipv4_compat: true
+    filter_chains:
+      - filters: &ref_1
+          - name: envoy.filters.network.http_connection_manager
+            typed_config:
+              '@type': >-
+                type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+              access_log:
+                - name: envoy.access_loggers.stdout
+                  filter:
+                    status_code_filter:
+                      comparison:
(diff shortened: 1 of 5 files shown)


ansible/files/envoy_config/lds.supabase.yaml
```diff
@@ -0,0 +1,436 @@
+resources:
```
hf, line 1: A verbatim copy of `lds.yaml`

hf:
Will bump versions after this is approved.

ansible/tasks/internal/setup-envoy.yml
hf, line 1: Remove, not needed.

pcnc (approved):

ansible/vars.yml
```diff
@@ -11,8 +11,8 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgres15: "15.8.1.010"
-  postgres16: "16.3.1.016"
+  postgres15: "15.8.1.011"
+  postgres16: "16.3.1.017"
```
hf, line 16: @samrose @supabase/postgres OK to merge with these?

samrose (approved):
Ok to merge

PR #1326, opened 2024-11-19
feat: separate envoy lds configs for self-hosting and supabase use-cases [15.6]

hf:
Like #1325 but for 15.6.


hf:
Will bump version after it's approved.

PR #1329, opened 2024-11-20
feat: upgrade to auth v2.164.0

J0:
Upgrade Auth version to v2.164.0

### DIFF
--- a/common.vars.pkr.hcl
+++ b/common.vars.pkr.hcl
@@ -1,1 +1,1 @@
-postgres-version = "15.1.1.96"
+postgres-version = "15.1.1.95"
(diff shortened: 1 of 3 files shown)


hf:
~Seems like something is off here?~

common.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "15.1.1.94"
+postgres-version = "15.1.1.95"
```
hf: Can you check `ansible/vars.yaml` not sure if it's on this branch.

J0: Could you check again? It's showing up on my end

image

samrose (changes requested):
@J0 @hf in this branch if you are targeting 15.6 you will actually need to bump the version in https://github.com/supabase/postgres/blob/release/15.6/common-nix.vars.pkr.hcl

samrose (approved):

PR #1343, opened 2024-11-28
feat: control shared_buffers value if oriole prior to start

samrose:

oriole needs different shared_buffer treatment/lower value, so for now we'll detect and set oriole to 32MB shared_buffers.

Once we have multiple sizes available, we'll check instance size and stagger it.

### DIFF
--- a/ansible/files/postgres_prestart.sh.j2
+++ b/ansible/files/postgres_prestart.sh.j2
@@ -19,31 +19,44 @@ get_shared_buffers() {
 update_orioledb_buffers() {
    local pg_conf="/etc/postgresql/postgresql.conf"
    local value="$1"
+   
+   # Update orioledb.main_buffers
    if grep -q "^orioledb.main_buffers = " "$pg_conf"; then
        sed -i "s/^orioledb.main_buffers = .*/orioledb.main_buffers = $value/" "$pg_conf"
    else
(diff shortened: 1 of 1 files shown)


ansible/files/postgres_prestart.sh.j2
```diff
@@ -19,31 +19,44 @@ get_shared_buffers() {
...
-   local has_orioledb=$(check_orioledb_enabled)
-   if [ "$has_orioledb" -lt 1 ]; then
-       return 0
-   fi
-   local shared_buffers_value=$(get_shared_buffers)
-   if [ ! -z "$shared_buffers_value" ]; then
-       update_orioledb_buffers "$shared_buffers_value"
+   
+   # Update shared_buffers during alpha release to 32MB
+   # TODO will need to add logic here for different ec2 sizes
```
darora, line 31: Let's not duplicate that logic here. The right place to do it is in the internal binary that already generates these for our current setup.

darora (approved):
Approving to unblock testing, but this logic should not be here in this repo, it should be in admin-api

ansible/files/postgres_prestart.sh.j2
samrose: @darora ok I will look into how to add that appropriately in admin-api closing this

PR #1360, opened 2024-12-06
feat: build and cache debug and src on this branch

samrose:

Needed to build and cache debug and src on the 15.6 branch

### DIFF
--- a/docker/nix/build_nix.sh
+++ b/docker/nix/build_nix.sh
@@ -5,6 +5,13 @@ nix --version
 if [ -d "/workspace" ]; then
     cd /workspace
 fi
+SYSTEM=$(nix-instantiate --eval -E builtins.currentSystem | tr -d '"')
 nix build .#psql_15/bin -o psql_15
 nix flake check -L 
 nix copy --to s3://nix-postgres-artifacts?secret-key=nix-secret-key ./psql_15
+if [ "$SYSTEM" = "aarch64-linux" ]; then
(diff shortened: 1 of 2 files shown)


docker/nix/build_nix.sh
```diff
@@ -5,6 +5,13 @@ nix --version
...
     cd /workspace
 fi
+SYSTEM=$(nix-instantiate --eval -E builtins.currentSystem | tr -d '"')
 nix build .#psql_15/bin -o psql_15
 nix flake check -L 
 nix copy --to s3://nix-postgres-artifacts?secret-key=nix-secret-key ./psql_15
+if [ "$SYSTEM" = "aarch64-linux" ]; then
+    nix build .#postgresql_15_debug -o ./postgresql_15_debug
+    nix build .#postgresql_15_src -o ./postgresql_15_src
+    nix copy --to s3://nix-postgres-artifacts?secret-key=nix-secret-key ./postgresql_15_debug-debug
```
tgallacher, line 15: I'm sure the psql drv will add the extra `-debug` suffix, but just to double check as I can't see where it might do this: Do you expect the additional (second) "debug" suffix?

samrose: nix unfortunately introduces that extra suffix, so it is correct, although not great.

samrose: I'll follow up soon to see if I can correct it as the source, but for now I just want to make sure this is available to our team.

tgallacher (approved):

PR #1362, opened 2024-12-09
fix: pgmq ownership

soedirgo:
Automatically assigns ownership of the pgmq extension and associated entities to `postgres` from `supabase_admin`.

This is already the case by default. However, when projects are paused and restored, they are restored with supabase_admin, which alters their ownership and makes the extension inoperable.

### DIFF
--- a/ansible/files/postgresql_extension_custom_scripts/pgmq/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/pgmq/after-create.sql
@@ -1,0 +1,19 @@
+do $$
+declare
+  extoid oid := (select oid from pg_extension where extname = 'pgmq');
+  r record;
+begin
+  set local search_path = '';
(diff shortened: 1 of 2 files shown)


samrose:
@soedirgo is there a way for us to keep adding to our tests we run [here](https://github.com/supabase/postgres/blob/develop/ansible/tasks/test-image.yml#L118), to test for these kinds of corrections? There is probably some point in that file where the appropriate configurations are run, that could then let us verify that permissions etc are the way they need to be. We could start out by testing this, and then eventually test many other permissions if they are not already tested.

Crispy1975 (approved):

soedirgo:
@samrose yeah I'll see if I can add that test; it should be something simple like:
- create extension pgmq
- check if pgmq tables, functions, etc. are owned by `postgres`

PR #1372, opened 2024-12-13
publish docker images/image updates for major versions

samrose:

- Add a Dockerfile that covers orioledb-17
- update publish to dockerhub to do a matrix build on any major version with corresponding `Dockerfile-<version>`, skipping if Dockerfile doesn't exist


Dockerfile-orioledb-17
```diff
@@ -0,0 +1,234 @@
...
+
+
+WORKDIR /
+
+
+RUN mkdir -p /usr/lib/postgresql/bin \
+    /usr/lib/postgresql/share/postgresql \
+    /usr/share/postgresql \
+    /var/lib/postgresql \
+    && chown -R postgres:postgres /usr/lib/postgresql \
```
avallete, line 80: **question**

Should we also have the `chown -R postgres:postgres /var/lib/postgresql` here like in the 15.X versions ?

avallete (approved):
LGTM

delgado3d (approved):

.github/workflows/mirror.yml
```diff
@@ -36,7 +43,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - uses: akhilerm/tag-push-action@v2.1.0
         with:
-          src: docker.io/supabase/postgres:${{ inputs.version }}
+          src: docker.io/supabase/postgres:${{ inputs.version }}_${{ inputs.arch }}
           dst: |
-            public.ecr.aws/supabase/postgres:${{ inputs.version }}
-            ghcr.io/supabase/postgres:${{ inputs.version }}
+            public.ecr.aws/supabase/postgres:${{ inputs.version }}_${{ inputs.arch }}
+            ghcr.io/supabase/postgres:${{ inputs.version }}_${{ inputs.arch }}
```
sweatybridge: We should be mirroring the merged manifest instead of per arch image.

.github/workflows/dockerhub-release-matrix.yml
```diff
@@ -0,0 +1,197 @@
...
+
+            let matrix = {include: $processed_versions}
+            let matrix_json = ($matrix | to json -r)
+            $"matrix=($matrix_json)" | save --append $env.GITHUB_OUTPUT
+          '
+
+  publish:
+    needs: get_publish_version
+    strategy:
+      matrix: ${{ fromJson(needs.get_publish_version.outputs.matrix) }}
```
sweatybridge: We don't need a matrix here because the merge manifest job should create a single tag that can be passed down to mirror workflow.

.github/workflows/dockerhub-release-matrix.yml
```diff
@@ -0,0 +1,197 @@
...
+          tags: ${{ steps.image.outputs.pg_version }}_${{ matrix.arch }}
+          platforms: linux/${{ matrix.arch }}
+          cache-from: type=gha,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
+          file: ${{ matrix.postgres.dockerfile }}
+  merge_manifest:
+    needs: [prepare, build, build_release_image]
+    strategy:
+      matrix:
+        include: ${{ fromJson(needs.prepare.outputs.matrix_config).include }}
```
sweatybridge, line 127: Ditto

samrose: In this case it's a matrix on just the major version and not on the architecture.

Dockerfile-orioledb-17
samrose: fixed thanks!

.github/workflows/dockerhub-release-matrix.yml
samrose: I ended up dropping that and doing it in a completely different way, but also made sure that after merge_manifest onward we are only dealing with the single tag produced from merge_manifest

.github/workflows/mirror.yml
samrose: fixed that thanks!

PR #1383, opened 2024-12-17
fix: jump update auth version to 2.165.1 from 2.163.2

hf:
When the `release/15.6` branch was retired in favor of develop, the variables in `ansible/vars.yml` weren't ported back, which caused a very old Auth version 2.163.2 to be installed.

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -9,9 +9,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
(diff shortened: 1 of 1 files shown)


hf:
What file do I bump for a new Postgres version?

pcnc:
Approved. You can bump the version number in vars.yaml. Bump up every major version, it’s fine

pcnc (approved):

PR #1384, opened 2024-12-19
Ensure that any `pgmq` sequences owned by `supabase_admin` are transferred to `postgres`

olirice:
Ensure that any `pgmq` sequences owned by `supabase_admin` are transferred to `postgres`

Tables are being correctly associated with `postgres` on restore but any sequence created before-pause are currently shifting ownership to supabase_admin

The permissions change is narrowly scoped to:
- sequences
- in the pgmq schema
- currently owned by supabase_admin


pcnc (approved):

olirice:
Realized why this is happening to sequences. going to solve it in the on-pause hook instead

PR #1389, opened 2024-12-24
fix: upgrade ami to contain auth 2.167.0

J0:
- Need to check if `common.vars.pkr.hcl` and if we should bump `postgres_release` in the vars.yml

Update: Done (see comments)

### DIFF
--- a/common.vars.pkr.hcl
+++ b/common.vars.pkr.hcl
@@ -1,1 +1,1 @@
-postgres-version = "15.1.1.95"
+postgres-version = "15.1.1.96"
--- a/common.vars.pkr.hcl
+++ b/common.vars.pkr.hcl
@@ -1,1 +1,1 @@
-postgres-version = "15.1.1.96"
+postgres-version = "15.1.1.95"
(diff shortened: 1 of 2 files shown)


common.vars.pkr.hcl
```diff
@@ -1 +1 @@
-postgres-version = "15.1.1.95"
```
samrose: On this one, we moved the postgres version to 

https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#L11

And you will want to bump the minor version on all three of those 

the common.vars.pkr.hcl file will be removed soon/deprecated. Instead, we now generate it at build time, and use the vars in https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#L11 as the "source of truth" for versions of postgres.

samrose (changes requested):

common.vars.pkr.hcl
```diff
@@ -1 +1 @@
...
+postgres-version = "15.1.1.96"
```
hf: This feels wrong? It was in another file I think.

samrose: @hf yes you got it. We changed it and now these versions live at https://github.com/supabase/postgres/pull/1389#discussion_r1896669812

samrose (approved):

pcnc (approved):

PR #1394, opened 2025-01-03
pg_mooncake test

olirice:
This PR is for experimental purposes only

Currently failing on needing rust > 1.81+ but current overlay is limited to 1.80


jmackie:
Hi 👋 Are there any plans to pick this back up? Or is there anything I can do to help? 🙏

olirice:
Yes we do intend to pick it back up. Got distracted by other shiny things
Please know that this is for experimental purposes only and is not an indication that pg_mooncake will land on the platform in any way. We have had equivalent branches for other extensions in the past

PR #1399, opened 2025-01-07
Sam/oriole17 flake check

samrose:

This PR consolidates `start-client` flake app into `start-server`, and then uses `start-server` in the `makeCheckHarness` nix flake check, and in that flake check, brings in the rest of the migrations and config to match our AMI set up as closely as possible. start-client now just wraps `psql` for the version of postgres we're running. `start-server` inherits the `--migration-file` capability, and has an optional `--skip-migrations` 

In addition, this PR refactors start-server to start/stop postgres with https://github.com/DarthSim/overmind so that managing the running service in foreground, background, etc is simpler. The script handles all interaction with overmind, so the `nix run ` commands are unchanged other than the new flags.

This PR will remain in DRAFT mode until the flake checks pass.


samrose:
This pr deprecates pg 16 for now as well

nix/tests/util/pgsodium_getkey.sh
```diff
@@ -1,10 +1,8 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 
-KEY_FILE="${1:-/tmp/pgsodium.key}"
-
-if [[ ! -f "${KEY_FILE}" ]]; then
-    head -c 32 /dev/urandom | od -A n -t x1 | tr -d ' \n' > "${KEY_FILE}"
+if [[ ! -f "$KEY_FILE" ]]; then
```
pcnc, line 7: This is missing - is it intended?
`KEY_FILE="${1:-/tmp/pgsodium.key}"`

pcnc (approved):

PR #1402, opened 2025-01-08
Isolate failing tests on pg17 and OrioleDB

olirice:
Separates tests that have different outputs on out pg15 and pg17 (or OrioleDB) so they can be more easily excluded from the  PG 17 build check.

Any tests that need to be scoped to a single build are prefixed with `z_<pg major version>` for example `z_17` for tests that should only run on pg17 build and `z_15` for tests that should only run on pg15


ansible/files/postgresql_extension_custom_scripts/pgmq/after-create.sql
```diff
@@ -2,18 +2,41 @@ do $$
 declare
   extoid oid := (select oid from pg_extension where extname = 'pgmq');
   r record;
+  cls pg_class%rowtype;
```
olirice, line 5: @soedirgo could you pls review this file's update

samrose (approved):

ansible/files/postgresql_extension_custom_scripts/pgmq/after-create.sql
soedirgo: Looks ok

PR #1403, opened 2025-01-09
feat: update envoy lds with origin protection keys

hf:
Updates the `lds.supabase.yaml` file to include origin protection keys by default. The string `supabase_origin_protection_key` is replaced with the exact OPK during [project initialization](https://github.com/supabase/infrastructure/blob/develop/init-scripts/project/00-init.sh#L115).

See also:
- https://github.com/supabase/postgres/pull/1325

### DIFF
--- a/ansible/files/envoy_config/lds.supabase.yaml
+++ b/ansible/files/envoy_config/lds.supabase.yaml
@@ -82,6 +82,25 @@ resources:
                                         name: ':path'
                                         string_match:
                                           contains: apikey=supabase_admin_key
+                        origin_protection_key_missing:
+                          permissions:
+                            - any: true
+                          principals:
+                            - not_id:
+                                header:
+                                  name: sb-opk
+                                  present_match: true
+                        origin_protection_key_not_valid:
+                          permissions:
+                            - any: true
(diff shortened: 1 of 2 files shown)


hf:
This change should go out to all Postgres versions that can be selected by customers. How do I do that?

ansible/files/envoy_config/lds.supabase.yaml
```diff
@@ -82,6 +82,25 @@ resources:
...
+                                header:
+                                  name: sb-opk
+                                  present_match: true
+                        origin_protection_key_not_valid:
+                          permissions:
+                            - any: true
+                          principals:
+                            - not_id:
+                                or_ids:
+                                  ids:
```
pcnc, line 99: Any reason for not doing a direct comparison here rather than an array check?

pcnc (approved):

ansible/files/envoy_config/lds.supabase.yaml
hf: There can be multiple keys here, coming from the `trusted_keys` column, to enable easy rotation.

PR #1417, opened 2025-01-23
Enabled log_statement = 'ddl' by default.

KevinBrolly:
Postgres `log_statement` config allows logging of DDL statements by default. This can be very useful in some debugging scenarios, and also helpful for assisting users when tables have been accidentally `DROP`'ed or altered.

### DIFF
--- a/docker/all-in-one/etc/postgresql-custom/postgresql-platform-defaults.conf
+++ b/docker/all-in-one/etc/postgresql-custom/postgresql-platform-defaults.conf
@@ -5,3 +5,4 @@ log_connections = on
 statement_timeout = 120000
 jit = off
 pgaudit.log = 'ddl'
+log_statement = 'ddl'


docker/all-in-one/etc/postgresql-custom/postgresql-platform-defaults.conf
```diff
@@ -5,3 +5,4 @@ log_connections = on
 statement_timeout = 120000
 jit = off
 pgaudit.log = 'ddl'
+log_statement = 'ddl'
```
darora, line 8: This file is only included in container images, so not used for AWS projects

KevinBrolly: @darora So I should edit it in the ansible conf, e.g. here (https://github.com/supabase/postgres/blob/develop/ansible/files/postgresql_config/postgresql.conf.j2#L543) so that it becomes the default for AWS projects?

PR #1420, opened 2025-01-27
chore: introduce tooling for pg 17.4 non-orioledb

samrose:


pcnc (approved):

migrations/db/init-scripts/00000000000000-initial-schema.sql
```diff
@@ -18,7 +18,15 @@ grant pg_read_all_data to supabase_read_only_user;
...
 create extension if not exists pgcrypto         with schema extensions;
-create extension if not exists pgjwt            with schema extensions;
+do $$
+begin 
+    if exists (select 1 from pg_available_extensions where name = 'pgjwt') then
+        if not exists (select 1 from pg_extension where extname = 'pgjwt') then
+            create extension if not exists pgjwt with schema "extensions" cascade;
+        end if;
+    end if;
+end $$;
```
steve-chavez, line 28: Could we wait for merging https://github.com/supabase/postgres/pull/1489 before this one? I'm moving this `migrations/db/init-scripts/00000000000000-initial-schema.sql` file there.

samrose: @steve-chavez sure. I am also trying to test and merge https://github.com/supabase/postgres/actions/runs/14040188774 into this one via rebase once i am done testing it too.

I'll focus on helping you get #1489 tests passing, and then we can rebase into this one hopefully by tomorrow, and restest with these changes (and test them on 17.4 version too)

samrose: This was actually resolved by @steve-chavez in https://github.com/supabase/postgres/pull/1493 (thanks @steve-chavez !) once that is merged we'll rebase and get those changes into this branch, go through more testing and get this into staging

samrose: Oh yes, and I think that #1493 needs to now be rebased into https://github.com/supabase/postgres/pull/1489 and then once #1489 is merged into develop, then we'll be good to go

migrations/tests/extensions/10-timescaledb.sql
```diff
@@ -1,7 +1,7 @@
 begin;
 do $_$
 begin
-  if not exists (select 1 from pg_extension where extname = 'orioledb') then
+  if current_setting('server_version_num')::integer >= 150000 and current_setting('server_version_num')::integer < 160000 then
```
steve-chavez, line 4: I had to exclude orioledb before since it gave errors when running with timescale. This conditional implies that orioledb will now work normally starting from pg 15?

samrose: I was not able to get the tests to run and pass without that change. Instead, postgres tried incorrectly to create extension timescaledb and plv8 for those extensions on orioledb-17 and on pg 17

timescale and plv8 are only supported on pg15

This condition tried to limit it to pg15, which is the only major version where timescale and plv8 are supported. If. you have a different way to achieve that feel free to suggest it of course.

samrose: (logic is if equal or greater than 15 and less than 16)

samrose: This is for creating extension timescaledb and not oriole fwiw

samrose: This branch contains both pg 17 and orioledb-17 and so it's better to just limit this logic to pg 15, as that is the only major version that supports these extensions that this file is trying to create extension on (same with plv8)

steve-chavez: Ok cool, got it. Thanks for elaborating.

flake.nix
```diff
@@ -435,6 +436,26 @@
...
+              mkdir -p $out
+              cp -r . $out
+            '';
+            meta = with pkgs.lib; {
+              description = "PostgreSQL 17 source files";
+              homepage = "https://www.postgresql.org/";
+              license = licenses.postgresql;
+              platforms = platforms.all;
+            };
+          };
```
steve-chavez, line 458: `flake.nix` has a lot already and it's getting hard to maintain, could we move this derivation to the `nix` directory?

samrose: @steve-chavez sure go for it

samrose: We can definitely refactor this, but going to be a follow up PR

migrations/db/init-scripts/00000000000000-initial-schema.sql
samrose: Update, we're going to move forward with this, and we'll work on re-introducing https://github.com/supabase/postgres/pull/1489 ASAP and integrating it into this, but we don't need to hold up the release of this, for that work.

PR #1422, opened 2025-01-29
fix: custom scripts were not configured to run correctly

samrose:

custom extension scripts were not configd to run correct in the `start-server` context, this pr fixes that


soedirgo (approved):

olirice:
awesome, knew something wasn't quite right but wasn't sure what. This will make those scripts much easier to test!

pcnc (approved):

PR #1444, opened 2025-02-15
chore: schemas needed updating

samrose:

Needed to run these commands any time migrations change

`nix run .#dbmate-tool -- --version 15 --flake-url "."`

`nix run .#dbmate-tool -- --version orioledb-17 --flake-url "."`

Will produce updates to schema per major version, which test.yml workflow can then test for diff


samrose:
deprecated

PR #1433, opened 2025-02-05
Move orioledb to extensions schema by default

olirice:
Moves the orioledb extension to the extensions schema if:
- orioledb is enabled in the `public` schema
- the `extensions` schema exists

This is important because orioledb exposes some entities that we don't necessarily want exposed over APIs by default

PR #1448, opened 2025-02-17
move orioledb to extensions schema

samrose:

Moves the orioledb extension to the extensions schema if:

orioledb is enabled in the public schema
the extensions schema exists
This is important because orioledb exposes some entities that we don't necessarily want exposed over APIs by default

Recommending we use this PR instead of https://github.com/supabase/postgres/pull/1433 as it is more up to date CI, less hassle toward reconciling git changes


darora (approved):
Does this work as expected with restores and unpauses?

samrose:
@darora thanks for the approval, I should have noted that we are going to hold off on merging this until I test restore, unpause. That should be done by end of day tomorrow, and then will set an actual vars.yml release version, and pr for rollout. Oliver stated that this does not require rollout to existing instances. But our policy now is to always test pause/unpause and restore, so will be doing that for sure.

samrose:
tested and confirmed no failures on pause/restore. Per @olirice this doesn't actually need to apply the sql migration on an existing project. But the migration succeeds on new project.

PR #1461, opened 2025-02-24
feat: add test for security definer functions

staaldraad:

test

## What is the new behavior?

Adds security / regression checks to ensure new security definer functions aren't introduced without first being vetted.

## Additional context

Checks all extensions installed, if a new extension is added, it should be present in [prime.sql](https://github.com/supabase/postgres/blob/develop/nix/tests/prime.sql). If a new security definer function is introduced, it must be vetted and if safe, added to the output  in security.out

Fixes SEC-204

### DIFF
--- a/nix/tests/expected/security.out
+++ b/nix/tests/expected/security.out
@@ -1,0 +1,30 @@
+-- get a list of security definer functions owned by supabase_admin
+-- this list should be vetted to ensure the functions are safe to use as security definer
+select
+    p.proname
+from pg_catalog.pg_proc p
+    left join pg_catalog.pg_namespace n ON n.oid = p.pronamespace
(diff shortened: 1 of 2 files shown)


darora (approved):
lgtm but should also get a +1 from @supabase/postgres

samrose:
@staaldraad I am re-running the failed testinfra test, as it appears to have just been a connectivity issue on that. Once it passes, I'll add an approval on this one. Thank you!

samrose:
@staaldraad all checks passing!

samrose (approved):

imor:
Maybe also prefix the function names with schema name to make it absolutely clear which function are security definers.

PR #1462, opened 2025-02-25
Update supautils.conf.j2 | add log_error_verbosity

TheOtherBrian1:

Supautils update

Users cannot change the `log_error_verbosity` setting

## What is the new behavior?

Allows them access through supautils

## Additional context

[User](https://app.hubspot.com/live-messages/19953346/inbox/8790930795#email) requested permission to update the value


ansible/files/postgresql_config/supautils.conf.j2
```diff
@@ -9,6 +9,6 @@ supautils.privileged_extensions = 'address_standardizer, address_standardizer_da
...
 supautils.privileged_extensions_superuser = 'supabase_admin'
 supautils.privileged_role = 'postgres'
-supautils.privileged_role_allowed_configs = 'auto_explain.*, log_lock_waits, log_min_duration_statement, log_min_messages, log_statement, log_temp_files, pg_net.batch_size, pg_net.ttl, pg_stat_statements.*, pgaudit.log, pgaudit.log_catalog, pgaudit.log_client, pgaudit.log_level, pgaudit.log_relation, pgaudit.log_rows, pgaudit.log_statement, pgaudit.log_statement_once, pgaudit.role, pgrst.*, plan_filter.*, safeupdate.enabled, session_replication_role, track_io_timing, wal_compression'
+supautils.privileged_role_allowed_configs = 'auto_explain.*, log_error_verbosity, log_lock_waits, log_min_duration_statement, log_min_messages, log_statement, log_temp_files, pg_net.batch_size, pg_net.ttl, pg_stat_statements.*, pgaudit.log, pgaudit.log_catalog, pgaudit.log_client, pgaudit.log_level, pgaudit.log_relation, pgaudit.log_rows, pgaudit.log_statement, pgaudit.log_statement_once, pgaudit.role, pgrst.*, plan_filter.*, safeupdate.enabled, session_replication_role, track_io_timing, wal_compression'
```
darora, line 12: Why would we want to allow this only for the `postgres` role?

TheOtherBrian1: Ideally, we'd give users free reign to change it for all roles.

I suggested changes to supautils because it is trivial to change and, at least in the short term, gives users some control

darora: Let's do it the right way instead, which also doesn't block on backporting supautils to existing projects.

PR #1473, opened 2025-03-12
chore: rm xmrig src by setting attribute to throw

samrose:

xmrig is apparently an official nix package https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/nixos/modules/services/misc/xmrig.nix

It is normal when installing a nix package, for nix to download the whole nix language package definition source code for all of the packages available in nixpkgs. However, only the executable binaries that you explicitly install are actually present.

Never the less, this pr sets the attribute for xmrig package to "throw", and runs garbage collect after install to remove any unrelated source , although the executable was NEVER installed or running at any time.

### DIFF
--- a/Dockerfile-15
+++ b/Dockerfile-15
@@ -67,6 +67,7 @@ WORKDIR /nixpg
 
 RUN nix profile install .#psql_15/bin 
 
+RUN nix store gc
 
 
 WORKDIR /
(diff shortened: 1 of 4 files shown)


darora (approved):

steve-chavez (approved):

samrose:
We're going to test this in CLI before merging and releasing, as the changes could impact cli/local dev users. Running nix garbage collect should have no impact, but we'll test that now to assure cli users have no issue.

PR #1484, opened 2025-03-18
fix: protect plpgsql

soedirgo:


Crispy1975 (approved):

samrose:
@staaldraad @doublethink can you take a look at this to make sure you don't see any security concern with these changes? Thank you!

PR #1485, opened 2025-03-19
Sam/wal g 2 and 3

samrose:

Keep the old binary around in AMIs for now, and link v3 to /usr/local/bin/wal-g-v3 v2 still links to /usr/local/bin/wal-g

### DIFF
--- a/ansible/tasks/setup-wal-g.yml
+++ b/ansible/tasks/setup-wal-g.yml
@@ -36,7 +36,7 @@
 - name: Create symlink for wal-g from Nix profile to /usr/local/bin
   ansible.builtin.file:
     src: /home/wal-g/.nix-profile/bin/wal-g-2
-    dest: /usr/local/bin/wal-g-v2
+    dest: /usr/local/bin/wal-g
     state: link
     force: yes  # This will replace existing file/symlink if it exists
   become: yes   # Need sudo to write to /usr/local/bin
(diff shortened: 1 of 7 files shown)


pcnc (approved):

ansible/tasks/setup-wal-g.yml
```diff
@@ -86,20 +21,45 @@
...
+  shell: |
+    sudo -u wal-g bash -c ". /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix profile install github:supabase/postgres/{{ git_commit_sha }}#wal-g-3"
   when: stage2_nix
 
 - name: Create symlink for wal-g from Nix profile to /usr/local/bin
   ansible.builtin.file:
-    src: /home/wal-g/.nix-profile/bin/wal-g
-    dest: /usr/local/bin/wal-g
+    src: /home/wal-g/.nix-profile/bin/wal-g-2
+    dest: /usr/local/bin/wal-g-v2
```
pcnc: We'd want to keep the old version backwards compatible, unfortunately

samrose: @pcnc so there you mean old version needs to be `/usr/local/bin/wal-g`

samrose: @pcnc maybe it is just that v2 needs to be installed at /usr/local/bin/wal-g but v3 installed at /usr/local/bin/wal-g-v3  but the default that can be run under command  wal-g in new ami is the v3 version is that the gist of what is needed

samrose: I @pcnc I think I got this resolved in the way you needed

samrose:
@pcnc ran through tests again here to verify, so merging this to hand off to you

PR #1489, opened 2025-03-21
fix: evtrigs ownership

steve-chavez:
Fixes #1437 (comment).

Moves migrations/db/init-scripts to migrations/db/migrations.

PR #1493, opened 2025-03-24
chore: move ansible migration tests to nix flake check 

steve-chavez:
## Problem

Currently the ansible migration tests are only done on CI, and this takes about 52 mins, which is terrible for development speed.

This is also needed for passing CI failures on https://github.com/supabase/postgres/pull/1489. Also related to https://github.com/supabase/postgres/pull/1492.

## Solution

Migration tests are now run on `nix flake check`, which can be used locally and on CI.

- tests for some extensions are adjusted to run conditionally on non-orioledb
  + preserves behavior from the ansible tests
- `ebssurrogate/files/unit-tests/unit-test-01.sql` is deleted
  + the same behavior is tested already on `nix/test/prime.sql`

Also cleanup `ansible/tasks/test-image.yml`:

- remove pg_prove since it's no longer used
- remove migration tests since they're now tested with nix flake check
- removed unit test run since unit-test-01.sql is no longer used (see above)


ansible/tasks/test-image.yml
```diff
@@ -74,53 +68,6 @@
...
-      register: files_to_remove
-      when: is_psql_oriole
-
-    - name: Delete matched extension files
-      ansible.builtin.file:
-        path: "{{ item.path }}"
-        state: absent
-      loop: "{{ files_to_remove.files }}"
-      when: is_psql_oriole
-      become: yes
```
steve-chavez, line 108: These exceptions from orioledb are now done properly using SQL on each test by checking if the orioledb extension is loaded. (there's no file removal needed)

migrations/tests/extensions/01-postgis.sql
```diff
@@ -1,43 +1,38 @@
...
-grant execute on all routines in schema tiger, tiger_data to authenticated;
-
--- authenticated role should have access now
-set local role authenticated;
-select tiger.pprint_addy(tiger.pagc_normalize_address('710 E Ben White Blvd, Austin, TX 78704'));
-ROLLBACK;
-
-BEGIN;
--- address standardizer creates a table in extensions schema, owned by supabase_admin
-create extension if not exists address_standardizer_data_us with schema extensions;
```
steve-chavez, line 39: This `create extension` was removed because it's already included on

https://github.com/supabase/postgres/blob/9eaaa417a1f72dad691268be938168b6eb5200c4/nix/tests/prime.sql#L5

but the logic that tests the postgres role is able to select `us_lex` is still preserved on the following lines.

samrose (approved):

samrose (approved):

PR #1499, opened 2025-03-26
revert and cherry pick commits 

samrose:
Reverting back to this state, and dropping migration commits for now. Will re-map those in in a follow up PR. Will push one more commit to this one to drop upload to prod on workflow for migrations


samrose:
closed in favor of https://github.com/supabase/postgres/pull/1500

PR #1512, opened 2025-04-02
fix: change qemu build to oriole

darora:
The array we grab the version from was re-ordered as PG17 non-oriole was introduced. A more robust mechanism can be added later but this array is rarely adjusted, so this is likely good enough for now.

Also moves up some fixes from the init-scripts, and removes Kong from the qemu artifact.

### DIFF
--- a/ebssurrogate/scripts/qemu-bootstrap-nix.sh
+++ b/ebssurrogate/scripts/qemu-bootstrap-nix.sh
@@ -106,7 +106,7 @@ EOF
 
 function clean_legacy_things {
     # removes things that are bundled for legacy reasons, but we can start without for our newer artifacts
-    apt-get unmark zlib1g* # TODO (darora): need to make sure that there aren't other things that still need this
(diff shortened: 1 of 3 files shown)


ebssurrogate/scripts/qemu-bootstrap-nix.sh
```diff
@@ -135,7 +135,12 @@ function clean_system {
 	mkdir /var/log/sysstat
 
 	chown -R postgres:postgres /var/log/wal-g
-	chmod -R 0300 /var/log/wal-g
+        # moving up fixes from init scripts
+	chmod -R 0310 /var/log/wal-g
+        chmod 0340 /var/log/wal-g/pitr.log
+
+        chmod 0600 /etc/vector/vector.yaml
+        chown vector:vector /etc/vector/vector.yaml
```
tgallacher: formatting misalignment?

tgallacher (approved):

PR #1519, opened 2025-04-03
fix: skip installing ndisc issue resolver in qemu mode

darora:
The issue was caused by very frequent RAs in ec2 when operating in
IPv6 mode. The qemu artifact gets configured with a static IP instead,
so this service should no longer be needed.

Additionally, fixes the build process to correctly define qemu_mode
for both ansible calls.

Also ports over a change from our init code for adminapi config loading.

### DIFF
--- a/ansible/files/adminapi.service.j2
+++ b/ansible/files/adminapi.service.j2
@@ -8,6 +8,9 @@ User=adminapi
 Restart=always
 RestartSec=3
 Environment="AWS_USE_DUALSTACK_ENDPOINT=true"
+{% if qemu_mode is defined and qemu_mode %}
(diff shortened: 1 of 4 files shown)


ansible/files/adminapi.service.j2
```diff
@@ -8,6 +8,9 @@ User=adminapi
 Restart=always
 RestartSec=3
 Environment="AWS_USE_DUALSTACK_ENDPOINT=true"
+{% if qemu_mode is defined and qemu_mode is True %}
```
darora: I _think_ this should work?

thebengeu (approved):

PR #1539, opened 2025-04-11
fix: grant Vault privs to service_role

soedirgo:
service_role used to be able to manage secrets in Vault <=0.2.8 because it had privileges to pgsodium functions; we do the grants for Vault functions instead

### DIFF
--- a/ansible/files/postgresql_extension_custom_scripts/supabase_vault/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/supabase_vault/after-create.sql
@@ -1,5 +1,5 @@
 grant usage on schema vault to postgres with grant option;
-grant select, delete on vault.secrets, vault.decrypted_secrets to postgres with grant option;
+grant select, delete, truncate, references on vault.secrets, vault.decrypted_secrets to postgres with grant option;
 grant execute on function vault.create_secret, vault.update_secret, vault._crypto_aead_det_decrypt to postgres with grant option;
 
 -- service_role used to be able to manage secrets in Vault <=0.2.8 because it had privileges to pgsodium functions
--- a/ansible/files/postgresql_extension_custom_scripts/supabase_vault/after-create.sql
+++ b/ansible/files/postgresql_extension_custom_scripts/supabase_vault/after-create.sql
@@ -1,3 +1,8 @@
 grant usage on schema vault to postgres with grant option;
 grant select, delete on vault.secrets, vault.decrypted_secrets to postgres with grant option;
 grant execute on function vault.create_secret, vault.update_secret, vault._crypto_aead_det_decrypt to postgres with grant option;
+
+-- service_role used to be able to manage secrets in Vault <=0.2.8 because it had privileges to pgsodium functions
+grant usage on schema vault to service_role;
+grant select, delete on vault.secrets, vault.decrypted_secrets to service_role;
+grant execute on function vault.create_secret, vault.update_secret, vault._crypto_aead_det_decrypt to service_role;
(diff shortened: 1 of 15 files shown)


ansible/files/postgresql_extension_custom_scripts/supabase_vault/after-create.sql
```diff
@@ -1,3 +1,8 @@
 grant usage on schema vault to postgres with grant option;
 grant select, delete on vault.secrets, vault.decrypted_secrets to postgres with grant option;
 grant execute on function vault.create_secret, vault.update_secret, vault._crypto_aead_det_decrypt to postgres with grant option;
+
+-- service_role used to be able to manage secrets in Vault <=0.2.8 because it had privileges to pgsodium functions
+grant usage on schema vault to service_role;
+grant select, delete on vault.secrets, vault.decrypted_secrets to service_role;
+grant execute on function vault.create_secret, vault.update_secret, vault._crypto_aead_det_decrypt to service_role;
```
steve-chavez, line 8: The first line is tested but the following two are not.

We need tests for this to ensure a regression doesn't happen.

I suggest adding them on `nix/tests/sql/vault.sql`

steve-chavez (changes requested):
We need more tests as explained above.

samrose:
If it helps, I had a moment and created some proposed tests that I think cover this PR, as a PR to this one here 

The tests are found in https://github.com/supabase/postgres/pull/1540

Feel free to reject that if it doesn't do what you need/you are implementing your own

soedirgo:
@supabase/postgres tests added, please +1. Aiming to release this early next week

steve-chavez (changes requested):
Tests need changing as mentioned on https://github.com/supabase/postgres/pull/1540#discussion_r2040678809

nix/tests/expected/vault.out
```diff
@@ -52,8 +58,9 @@ ORDER BY object_name, grantee, privilege_type;
  vault  | secrets                   | supabase_admin | TRUNCATE
  vault  | secrets                   | supabase_admin | UPDATE
  vault  | update_secret             | postgres       | EXECUTE
+ vault  | update_secret             | service_role   | EXECUTE
```
steve-chavez, line 65: Effects of the changes now can be clearly visualized.

steve-chavez (approved):
PR was rebased on top of https://github.com/supabase/postgres/pull/1544, test now look good.

The `chore: bump versions` commit didn't make the rebase as some other commit already used the same versions on https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#L11-L14.

samrose:
> PR was rebased on top of #1544, test now look good.
> 
> The `chore: bump versions` commit didn't make the rebase as some other commit already used the same versions on https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#L11-L14.

I think in this case we just need to advance those versions one more than 

https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#L11

..unless this PR has not been tested in local infra for relevant scenarios. If not, we should add a suffix to the version and test, then remove that and merge.

steve-chavez:
> I think in this case we just need to advance those versions one more than

Did that, although not sure if correct.

samrose:
> > I think in this case we just need to advance those versions one more than
> 
> Did that, although not sure if correct.

version bump looks correct, but also we need to make sure we have tested these changed before we merge into develop. If you're not able to do so, either myself or @soedirgo can for the short term.

soedirgo:
Tested on local infra (old: 15.8.1.073, new: 15.8.1.074-vault-2):
- pause on old AMI, restore to new AMI
- pause on new AMI, restore to new AMI
- upgrade from old AMI to new AMI

PR #1452, opened 2025-02-19
feat: reland vault w/o pgsodium

soedirgo:
Reland https://github.com/supabase/postgres/pull/1431; the migration has been merged in https://github.com/supabase/postgres/pull/1442

Tested flows on local infra:
1. pause & restore w/o vault
	- create a new project
	- `drop extension supabase_vault cascade;`
	- pause project
	- restore project
	- `select * from vault.decrypted_secrets`
1. pause & restore w/ existing secrets
	- create a new project
	- create a new secret in Settings > Vault
	- pause project
	- restore project
	- `select * from vault.decrypted_secrets`
1. pause & restore w/ existing secrets w/o pgsodium
	- follow steps in 2
	- `drop extension pgsodium`
	- pause project
	- restore project
	- `select * from vault.decrypted_secrets`
1. pg_upgrade
	- create a new project on an earlier AMI version (e.g. 15.8.1.044)
	- create a new secret in Settings > Vault
	- upgrade project in Settings > Infrastructure
	- `select * from vault.decrypted_secrets`

PR #1540, opened 2025-04-11
chore: proposed additions to vault regress tests

samrose:

proposed regress tests for covering scenarios in the PR #1539

### DIFF
--- a/nix/tests/expected/vault.out
+++ b/nix/tests/expected/vault.out
@@ -1,4 +1,4 @@
-SET ROLE service_role;
+SET ROLE postgres;
 SELECT EXISTS (
   SELECT 1 FROM vault.create_secret('my_s3kre3t')
 ) AS can_create_secret;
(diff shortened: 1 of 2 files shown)


nix/tests/expected/vault.out
```diff
@@ -1,42 +1,101 @@
...
-from
-  vault.create_secret(
+SET ROLE service_role;
+SELECT EXISTS (
+  SELECT 1 FROM vault.create_secret('my_s3kre3t')
+) AS can_create_secret;
+ can_create_secret 
+-------------------
+ t
+(1 row)
```
steve-chavez, line 8: @samrose These types of tests (similar to pgtap style) are hard to understand IMO and often times incomplete. 

What we need instead are queries on the catalogs like:

https://github.com/supabase/postgres/blob/ae8cdaf6c41a656a262c97c45d19b2fbce82c1ee/nix/tests/expected/roles.out#L186-L191

This way when a change in privileges happens, it's very clear on the resulting diffs. Which helps identifying problems and also evolving the code.

soedirgo: Let's call this approach [testing against public APIs](https://abseil.io/resources/swe-book/html/ch12.html#test_via_public_apis) for the sake of argument

I disagree that this is harder to understand - it's a lot clearer what's being tested here, especially for people who don't work on Vault and understand its internal workings (this one is to assert that you can create a secret, that one is to assert that you can do it with a specific role, etc.)

Snapshotting the system catalogs makes it easier to achieve completeness (you'd need to be a lot more verbose to do the same by testing against public APIs) but it doesn't guarantee completeness either. E.g. it wouldn't have caught the issue being addressed in https://github.com/supabase/postgres/pull/1539, because it arose due to us using a new function introduced in the new Vault for creating secrets.

You'd see a diff, sure - it'll say new functions are introduced, but I can tell you that my reaction to that diff would've been "new functions, that's expected" - I wouldn't see the connection between that and `service_role`.

But these tests would've caught it.

soedirgo: You also pay for that ease of achieving completeness with [brittleness](https://abseil.io/resources/swe-book/html/ch12.html#test_via_public_apis#preventing_brittle_tests). Now every time e.g. a function or a privilege grant is added you get a test failure even if it doesn't introduce any bugs.

steve-chavez: > You also pay for that ease of achieving completeness with [brittleness](https://abseil.io/resources/swe-book/html/ch12.html#test_via_public_apis#preventing_brittle_tests). Now every time e.g. a function or a privilege grant is added you get a test failure even if it doesn't introduce any bugs.

This is why you limit the queried columns from the catalogs. You can already see the clear diff on https://github.com/supabase/postgres/pull/1539/files#diff-db10041c0fa09ebfd2e8573a68c98cce8afba5db52e1d908e9d0760a3558cf88R578, that's the main goal. No brittleness there.

steve-chavez: Maybe this particular test does make sense, but I see others below having the result of `
ERROR:  permission denied for function _crypto_aead_det_noncegen`. Those are better in terms of pg_catalog queries.

If this test is proving that `ERROR: permission defnied..` doesn't happen, then it should be changed.

soedirgo: That is actually a case of brittleness, because `service_role` not directly having `USAGE` privilege is not it itself a bug - it previously inherited that privilege from `pgsodium_keyiduser` IIRC.<br><br>What would make it less brittle is if it tests that `service_role` has the privilege (whether direct or indirect). What would be even less brittle is if we test for the use case we care about (i.e. `service_role` not being able to create secrets)

soedirgo: A clearer example of the brittleness is that it's causing test failures [here](https://github.com/supabase/postgres/pull/1452/files#diff-48903d7fcddff22a03dfce62dcc706abfce5502ac1fe4008533dba1d6e6cb3a6) even though it doesn't signal any bug

soedirgo: > If this test is proving that ERROR: permission defnied.. doesn't happen, then it should be changed.

Yeah that should be changed. It should just check that the statement fails to run, the specific function where it fails should be irrelevant

steve-chavez: > I disagree that this is harder to understand - it's a lot clearer what's being tested here,
> I can tell you that my reaction to that diff

There's a lot of personal preference in the above argument, but what's more important is that all contributors here agree on a "convention", at least for "privilege testing". Otherwise this aspect will be hard to maintain. Note that there are already various tests for permissions, like: 

https://github.com/supabase/postgres/blob/36b18b42fa4f8fcb0ce19f8193221c9538a25c7c/nix/tests/expected/roles.out#L186-L200

It's a matter of deciding if we want to keep doing those, or we want to do this type of testing a la pgtap, globally. I for one, would prefer to keep a consistency with previous work, which to me is the clearest, at least for "privilege testing".

steve-chavez: I might be able to help with these tests new week. Then we can compare both approaches better.

steve-chavez (changes requested):
Tests need changing as discussed above.

nix/tests/expected/vault.out
soedirgo: I don't mind following that convention for privilege testing and role memberships. These tests are not specifically testing for privileges though, they're testing interfaces (but for different roles). E.g. the test should fail if `service_role` tries to create a secret, but the function throws an error during execution.

soedirgo: Actually snapshotting catalogs is still harder to read especially for inherited privileges - using something like `has_schema_privilege` accounts for inherited privileges and it's a lot clearer what it's testing for.

Anyway let's not block the AMI release because of testing approaches. That can be done is a separate PR - managing secrets through `service_role` is broken until this is released.

soedirgo (approved):

nix/tests/expected/vault.out
steve-chavez: > Anyway let's not block the AMI release because of testing approaches.

Hard disagree here. Maintenance is important and tech debt accrues, making changes much harder later. Case in point: the recent event triggers delayed release.

Also don't see why this merging had to be rushed over the weekend when I offered to improve things here

steve-chavez: Improved tests were done on https://github.com/supabase/postgres/pull/1544

steve-chavez:
Changes here were superseded by https://github.com/supabase/postgres/pull/1544.

PR #1549, opened 2025-04-15
feat: multi-ext-versions-pgrcron

samrose:

Incremental repackage of pg extensions to include all known installed branches, starting with pg cron. This release will include 1.3, 1.4, 1.6 

Older versions only support pg 15, and so are only covered with pg 15, and WILL NOT be installed along with pg 17+ or oriole releases. This PR includes a patch for 1.3.1 version due to incompat with 15.8

We will test to see that this still on restore, across all versions.


nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,121 @@
...
+          --replace '#include "postgres.h"' '#include "postgres.h"
+      #include "commands/async.h"
+      #include "miscadmin.h"'
 
-  buildInputs = [ postgresql ];
+        # Update function calls to use PostgreSQL 15 APIs
+        substituteInPlace src/pg_cron.c \
+          --replace 'ProcessCompletedNotifies();' '/* ProcessCompletedNotifies removed */' \
+          --replace 'pg_analyze_and_rewrite(parsetree, sql, NULL, 0,NULL);' 'pg_analyze_and_rewrite_fixedparams(parsetree, sql, NULL, 0, NULL);'
+      '';
```
steve-chavez: @samrose Could we have the patches as a `.patch` file? We already do this for plv8 ([ref](https://github.com/supabase/postgres/blob/develop/nix/ext/0001-build-Allow-using-V8-from-system.patch)).

Changes are easier to see that way and applying the patch with the Nix `patches` attribute is a standard practice.

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,121 @@
...
+    if lib.versionAtLeast pg.version "17"
+    then { "1.6.4" = allVersions."1.6.4"; }
+    else allVersions;
+
+  allVersionsForPg = lib.mapAttrs mkPgCron (getVersions postgresql);
+
+in
+stdenv.mkDerivation {
+  pname = "pg_cron-all";
+  version = "multi";
```
steve-chavez: Q: Could this `version` give us trouble later on since it looks like it's not meant to be modified? Say if we modify the individual pg cron versions, could the hardcoded `multi` version cause a "permanent stale cache"?

samrose: @steve-chavez it's a good call, I will create an increment-able version there, that we can advance when/if we change the the package at all.

steve-chavez: @samrose Not sure if possible, but maybe this version can be a concatenation of all the individual pg_cron versions. That way it doesn't have to be updated manually.

nix/ext/pg_cron.nix
samrose: resolved

nix/ext/pg_cron.nix
samrose: @steve-chavez hmm yeah it probably is possible. Ok!

samrose: Done!

steve-chavez (approved):

Crispy1975 (approved):

samrose:
@Crispy1975 I was going to wait to merge this until you had a chance to test it out in staging. Just wanted to add this as a status on this PR.

samrose:
We needed to handle the lib that pg_cron worker uses on start of postgres server cluster.

So, we're going to add a file pg_extensions.json and read that, and set the alias before postgres starts to account for the worker.

steve-chavez:
> We needed to handle the lib that pg_cron worker 

@samrose Could you elaborate on that? Which lib does it need and why do we require a `pg_extension.json` file? Can't the dependency by fully managed by a Nix expression?

samrose:
> > We needed to handle the lib that pg_cron worker
> 
> @samrose Could you elaborate on that? Which lib does it need and why do we require a `pg_extension.json` file? Can't the dependency by fully managed by a Nix expression?

It already is. The nix dependency packaging part is done.

However, the AMI and Nix package won't know which version is being restored to. So, when a restore is triggered, the version that was active at backup time will be recorded into the infrastructure middleware db, and then it will be written as a datapoint to a file on restore. Then, the postgres prestart script will read that version number, and re-alias the installed package to the expected version, because postgres starts pg_cron worker on startup.

steve-chavez:
> However, the AMI and Nix package won't know which version is being restored to. So, when a restore is triggered, the version that was active at backup time will be recorded into the infrastructure middleware db, and then it will be written as a datapoint to a file on restore. 

Will the above be relevant even when we transition to physical backups?

samrose:
> > We needed to handle the lib that pg_cron worker
> 
> @samrose Could you elaborate on that? Which lib does it need and why do we require a `pg_extension.json` file? Can't the dependency by fully managed by a Nix expression?

We package all of known installed versions of pg_cron, then when a project is restored from a backup file, a specific version was in use when the backup happened. So, on restore, we can switch alias to that nix packaged version, and it will restore without error AND the pg_cron worker that starts up when postgres starts up will use the right version.

samrose:
> > However, the AMI and Nix package won't know which version is being restored to. So, when a restore is triggered, the version that was active at backup time will be recorded into the infrastructure middleware db, and then it will be written as a datapoint to a file on restore.
> 
> Will the above be relevant even when we transition to physical backups?

yes, for sure @Crispy1975 is working with me on this now and can confirm. The same issue may affect timescale too

Extensions that do not start a worker at postgres startup likely won't have this issue and will just be able to find and use the packaged version it needs (we'll find out when we test it).

THIS pr is for pg_cron. We are packaging and releasing these multiversion one extension at a time.

samrose:
> THIS pr is for pg_cron. We are packaging and releasing these multiversion one extension at a time.

and pg_cron definitely needs this extra handling, due to worker starting at pg server startup

samrose:
all of this multi version work will be solely for physical backup approach fwiw. so, anything we do in multiversion extension pkg will be due to initial testing and confirming we need to do it for physical backups.

steve-chavez:
> yes, for sure @Crispy1975 is working with me on this now and can confirm. The same issue may affect timescale too

Got it. Thanks for clarifying.

> Extensions that do not start a worker at postgres startup likely won't have this issue and will just be able to find and use the packaged version it needs (we'll find out when we test it).

Let me backup this a bit. So the root problem is that the restore doesn't capture the C library version right? (that's where the code for background worker is)

And this is because it just does `CREATE EXTENSION pg_cron <version>`? (and this only contains the SQL code)

samrose:
> > yes, for sure @Crispy1975 is working with me on this now and can confirm. The same issue may affect timescale too
> 
> Got it. Thanks for clarifying.
> 
> > Extensions that do not start a worker at postgres startup likely won't have this issue and will just be able to find and use the packaged version it needs (we'll find out when we test it).
> 
> Let me backup this a bit. So the root problem is that the restore doesn't capture the C library version right? (that's where the code for background worker is)
> 
> And this is because it just does `CREATE EXTENSION pg_cron <version>`?

the restore captures it, but the restore can't read prior to preforming the restore because it is a binary file. So, our solution is to simultaneously record the version that was installed at backup time, and then make that data available as a file when the restore is run, the prestart can read it, and can alias pg_cron to the right lib before postgres starts. Then, restore *should* take care of the rest (we are testing this theory now).

testinfra/test_ami_nix.py
```diff
@@ -321,14 +329,103 @@ def is_healthy(host, instance_ip, ssh_identity_file) -> bool:
...
+                        logger.error(f"systemd status: {systemd_status.stdout}")
+                        logger.error(f"systemd error: {systemd_status.stderr}")
+                        
+                        # Get detailed systemd status
+                        logger.error("Detailed systemd status:")
+                        host.run("sudo systemctl status postgresql -l --no-pager")
+                        
+                        # Check init script logs
+                        logger.error("Init script logs:")
+                        host.run("sudo journalctl -u cloud-init --no-pager")
```
steve-chavez: Q: Can this logic be tested locally? Or does it rely on a component outside of this repo?

samrose: @steve-chavez it's a high priority for me to be able to test this locally, but as of this moment it cannot. I don't like this state of affairs and so will be finding some way to change it ASAP. For today, I am stuck doing it this way.

samrose: (it relies on creating an AWS ami, and running an ec2 instance. In the RFC I am working on, it will recognize if it is trying to be run in aws or in a local ubuntu vm, and switch context and run)

samrose: It will also have a way to create a base AMI nightly, boot a machine based off of the latest nightly available, and make the changes that need to be made only. This will drastically speed up all AMI building operations

samrose: @steve-chavez over here  https://github.com/supabase/postgres/pull/1547 I came very close to being able to run this locally, although it really in that case is just still running on aws which does not solve the problem fully. I'd rather be able to run fully local, or in ci. But will need to do some work to get this our repo to achieve that. But it's going to happen relatively soon.

steve-chavez: @samrose All good. Thanks for elaborating.

samrose: @steve-chavez I am actually creating a PR to this PR, to try and have an immediate way to run this locally.

samrose: It's too much work to piggy back on this. But we're going to get into it soon

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,147 @@
...
+    };
+    "1.6.4" = {
+      rev = "v1.6.4";
+      hash = "sha256-t1DpFkPiSfdoGG2NgNT7g1lkvSooZoRoUrix6cBID40=";
+    };
+    "1.5.2" = {
+      rev = "v1.5.2";
+      hash = "sha256-+quVWbKJy6wXpL/zwTk5FF7sYwHA7I97WhWmPO/HSZ4=";
+    };
+  };
```
RossComputerGuy, line 22: Instead of specifying the rev's and inheriting them, wouldn't it be better to just do `tag = "v${version}"` in the `src`?

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,147 @@
...
 
   meta = with lib; {
-    description = "Run Cron jobs through PostgreSQL";
-    homepage    = "https://github.com/citusdata/pg_cron";
-    changelog   = "https://github.com/citusdata/pg_cron/raw/v${version}/CHANGELOG.md";
-    platforms   = postgresql.meta.platforms;
-    license     = licenses.postgresql;
+    description = "Run Cron jobs through PostgreSQL (multi-version compatible)";
+    homepage = "https://github.com/citusdata/pg_cron";
+    platforms = postgresql.meta.platforms;
```
RossComputerGuy, line 168:
```suggestion
    inherit (postgresql.meta) platforms;
```

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,147 @@
...
 
+  getVersions = pg:
+    if lib.versionAtLeast pg.version "17"
+    then { "1.6.4" = allVersions."1.6.4"; }
+    else allVersions;
+
+  allVersionsForPg = lib.mapAttrs mkPgCron (getVersions postgresql);
+
+in
+stdenv.mkDerivation {
```
RossComputerGuy, line 75: This looks like a light derivation so maybe `runCommand` would be simpler to use? You can still pass in `buildInputs` and `version`. Then at least `dontUnpack`, `dontConfigure`, `dontBuild` wouldn't be needed.

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,147 @@
...
+    if [ ! -f "$LIB_DIR/pg_cron-$VERSION$EXT" ]; then
+      echo "Error: Version $VERSION not found"
+      exit 1
+    fi
+
+    # Update library symlink
+    ln -sfnv "pg_cron-$VERSION$EXT" "$LIB_DIR/pg_cron$EXT"
+
+    echo "Successfully switched pg_cron to version $VERSION"
+    EOF
```
RossComputerGuy: Are other extensions going to do the same thing? If so, could there be a "generic" wrapper that could have `substituteAll` be called on?

nix/ext/pg_cron.nix
```diff
@@ -1,31 +1,147 @@
...
+  dontUnpack = true;
+  dontConfigure = true;
+  dontBuild = true;
+
   installPhase = ''
-    mkdir -p $out/{lib,share/postgresql/extension}
+    mkdir -p $out/{lib,share/postgresql/extension,bin}
+    
+    # Install all versions
+    for drv in ${lib.concatStringsSep " " (lib.attrValues allVersionsForPg)}; do
```
RossComputerGuy, line 89: Might be able to just use `$buildInputs` instead of the `concatStringsStep` stuff.

nix/ext/pg_cron.nix
samrose: There may be one other extension that does something similar but I am not sure yet if it will be generic enough to re-use

samrose:
deprecated in favor of #1737

PR #1550, opened 2025-04-16
fix: update wrappers server options post-upgrade

soedirgo:
Wrappers were previously using `vault.secrets.key_id`, which will no longer work with new Vault; we migrate it to use `vault.secrets.id` instead.

Tested on local infra with the same steps as [this](https://github.com/supabase/infrastructure/pull/22555), swapping out the pause & restore step with upgrading the project via Project Settings > Infrastructure

TODO:
- [x] merge https://github.com/supabase/supabase/pull/35075

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
@@ -42,8 +42,34 @@ function execute_wrappers_patch {
     option_rec RECORD;
     vault_secrets RECORD;
   BEGIN
-    IF EXISTS (SELECT FROM pg_available_extension_versions WHERE name = 'wrappers' AND version = '0.4.6')
-      AND EXISTS (SELECT FROM pg_extension WHERE extname = 'wrappers')
+    IF EXISTS (SELECT FROM pg_extension WHERE extname = 'wrappers' AND extversion NOT IN (
+      '0.1.0',
+      '0.1.1',
+      '0.1.4',
+      '0.1.5',
+      '0.1.6',
+      '0.1.7',
+      '0.1.8',
+      '0.1.9',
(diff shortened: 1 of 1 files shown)


ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -32,6 +32,45 @@ function execute_extension_upgrade_patches {
...
+    # If upgrading to pgsodium-less Vault, Wrappers need to be updated so that
+    # foreign servers use `vault.secrets.id` instead of `vault.secrets.key_id`
+    UPDATE_WRAPPERS_SERVER_OPTIONS_QUERY=$(cat <<EOF
+  DO \$\$
+  DECLARE
+    server_rec RECORD;
+    option_rec RECORD;
+    vault_secrets RECORD;
+  BEGIN
+    IF EXISTS (SELECT FROM pg_available_extension_versions WHERE name = 'wrappers' AND version = '0.4.6')
```
pcnc: Is 0.4.6 going to be the only impacted version, or do we need to account for future versions as well?

pcnc (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
soedirgo: Need to account for future versions; was gonna say this should already check for that since all future versions should have previous versions in `pg_available_extension_versions` but that's not the case 🤦

soedirgo: Updated the PR and re-testing

soedirgo:
~Tested again and confirmed to be working~ not, the tested script is on https://github.com/supabase/postgres/pull/1552

PR #1554, opened 2025-04-17
fix: update pgbouncer.get_auth (INFRA-1530)

kangmingtay:
* Updates the severity level of from warning to debug: INFRA-1530
* Updates `pgbouncer.get_auth` definition

## What is the new behavior?

## Additional context


steve-chavez (changes requested):
@samrose @kangmingtay We need pgbouncer tests to ensure a regression doesn't happen.

They should be similar to the ones [auth.out](https://github.com/supabase/postgres/blob/61ff753243078fe2b14e5b8f184f4a00c80342ee/nix/tests/expected/auth.out#L1) has. Namely:

- tests that cover the pgbouncer schema: tables with owners, functions with owners
- tests that cover the privileges roles have on the pgbouncer db objects

Just noted there's a test for the `get_auth` function:
https://github.com/supabase/postgres/blob/61ff753243078fe2b14e5b8f184f4a00c80342ee/nix/tests/expected/security.out#L14

But it doesn't cover the `owner`, that needs to be added.

steve-chavez:
How urgent is this? I might be able to help with the tests here if there's time.

migrations/schema-15.sql
```diff
@@ -1003,3 +1009,46 @@ CREATE EVENT TRIGGER pgrst_drop_watch ON sql_drop
...
+    ('00000000000002'),
+    ('00000000000003'),
+    ('10000000000000'),
+    ('20211115181400'),
+    ('20211118015519'),
+    ('20211122051245'),
+    ('20211124212715'),
+    ('20211130151719'),
+    ('20220118070449'),
+    ('20220126121436'),
```
steve-chavez: Q: Why is this needed? What service keeps track of these migration rows?

Also, it's duplicated across several files. Is there a way to DRY this?

kangmingtay: hmm these were actually generated from `nix run .#dbmate-tool -- --version all`

steve-chavez: Right, this was clarified on https://github.com/supabase/postgres/pull/1555#discussion_r2049738715.

migrations/db/migrations/20250417190610_update_pgbouncer_get_auth.sql
```diff
@@ -0,0 +1,24 @@
...
+            then null 
+            else rolpassword::text 
+        end 
+    from pg_authid 
+    where rolname=$1 and rolcanlogin;
+end;
+$$;
+
+-- from migrations/db/migrations/20250312095419_pgbouncer_ownership.sql
+grant execute on function pgbouncer.get_auth(p_usename text) to postgres;
```
steve-chavez, line 22: @kangmingtay We just need one more test to prove this is working correctly :pray: . Just copy/paste this query from vault and change it to use the `pgbouncer` schema:

https://github.com/supabase/postgres/blob/61ff753243078fe2b14e5b8f184f4a00c80342ee/nix/tests/expected/vault.out#L1-L27

This should show all the privileges on the pgbouncer schema objects.

nix/tests/sql/pgbouncer.sql
```diff
@@ -0,0 +1,54 @@
...
+  pg_namespace n on p.pronamespace = n.oid
+join
+  pg_roles r on p.proowner = r.oid
+where
+  n.nspname = 'pgbouncer'
+order by
+  p.proname;
+
+-- Tests role privileges on the pgbouncer objects
+-- INSERT and UPDATE privileges should not be present on the pgbouncer tables for postgres and service_role, only SELECT and DELETE
```
steve-chavez: This comment shouldn't be needed, it made sense in the vault test but not here. Better to delete it as it could cause confusion later.
```suggestion
```

steve-chavez (approved):

steve-chavez (approved):
Great work! :100: 

Those testinfra CI jobs are flaky, I suggest just rerunning them manually and see if they pass.

PR #1555, opened 2025-04-17
docs: schema.sql per maj version update

samrose:

docs needed coverage on updating schema.sql per major version based on feedback from @kangmingtay who tried to use the docs. They lacked this information.

### DIFF
--- a/migrations/README.md
+++ b/migrations/README.md
@@ -130,7 +130,22 @@ After making changes to migrations, you should update the schema.sql files for e
 nix run .#dbmate-tool -- --version all
 ```
 
-This will create a schema.sql file for each major version of PostgreSQL and OrioleDB (the files are named like `schema-<ver>`, `schema-oriole-<ver>`). Commit these changes to your repository and push to your branch. The test.yml workflow will verify these changes against the test matrix.
+This will create automatically  schema.sql file for each major version of PostgreSQL and OrioleDB (the files are named like `schema-<ver>`, `schema-oriole-<ver>`). Commit these changes to your repository and push to your branch. The workflow in `.github/workflows/test.yml`  will re-run this command in CI, and perform a git diff to verify the idempotency of the migrations, and that the latest changes have been committed.
+
 ## Testing
 
-Migrations are tested in CI to ensure they do not raise an exception against previously released `supabase/postgres` docker images. The full version matrix is at [test.yml](./.github/workflows/test.yml) in the `supabase-version` variable.
+In addition to ci test mentioned above, you can test migrations locallay by running the following test for each major version of postgres one at a time.
+
+Examples:
+
+```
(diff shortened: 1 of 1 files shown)


migrations/README.md
```diff
@@ -120,7 +120,17 @@ Then, populate the migration at `./db/migrations/xxxxxxxxx_<some_message>` and m
...
 
+After making changes to migrations, you should update the schema.sql files for each major version of PostgreSQL:
+
+```shell
+# First, stop any running PostgreSQL servers
+# Then from the root of supabase/postgres run:
+nix run .#dbmate-tool -- --version all
+```
+
+This will create a schema.sql file for each major version of PostgreSQL. Commit these changes to your repository and push to your branch. The test.yml workflow will verify these changes against the test matrix.
```
steve-chavez: This will also create the `migrations/schema-orioledb-17.sql`?

steve-chavez: Btw why is this needed? I had a related question on https://github.com/supabase/postgres/pull/1554/files#r2049582929.

samrose: Absolutely it will create a schema sql for 15, 17 and orioles 17 because they are different across major versions

samrose: It will append the version on the end

steve-chavez: Ok cool, then we should put it here for completion:

```suggestion
This will create a schema.sql file for each major version of PostgreSQL and OrioleDB (the files are named like `schema-<ver>`, `schema-oriole-<ver>`). Commit these changes to your repository and push to your branch. The test.yml workflow will verify these changes against the test matrix.
```

steve-chavez: Follow up question: why do we need these `schema-<ver>` files? The changes are redundant with the files included in `migrations/` right? Is it for visibility?

samrose: @steve it’s re run in CI compared to the existing file in repo and detects if migrations are covered and committed to this repo

steve-chavez: @samrose So they're for ensuring migrations are idempotent?

How about clarifying those are autogenerated here in the docs? 

We need to clarify that otherwise it's not clear and seems they're duplicated code, which leads to questions: https://github.com/supabase/postgres/pull/1554#discussion_r2049582929.

samrose: @steve-chavez ok sure. Yeah this is a common practice when dbmate tool is used. But sure we can document it the way you describe so it’s clear

samrose: @steve-chavez have a look  and let me know if these changes are good enough

https://github.com/supabase/postgres/pull/1555/files

steve-chavez: Yes, that looks good now. Thanks for clarifying.

steve-chavez (approved):

kangmingtay (approved):

PR #1557, opened 2025-04-18
feat: keep pgjwt present but deactivated for orioledb only to avoid r…

samrose:
This PR will also need to address this change in https://github.com/supabase/infrastructure/blob/develop/init-scripts/project/00-initial-schema.sql#L20 as well.

At this time, the scope of this change is to fix the following issues only. If we need to expand the scope, it needs to be in a different PR.

1. Remove pgjwt from the `dbExtensions17` filter in `flake.nix` to ensure it's not included in PostgreSQL 17 builds
2. Keep pgjwt in `orioleFilteredExtensions` so it's included in the orioledb-17 package
3. Keep pgjwt in the `supautils.privileged_extensions` list in `supautils.conf.j2` since it's a privileged extension
4. Ensure pgjwt is not in the `shared_preload_libraries` list in `postgresql.conf.j2` to prevent it from being loaded by default
5. Keep the existing sed commands in `Dockerfile-orioledb-17` and `ansible/tasks/stage2-setup-postgres.yml` that remove pgjwt from supautils.conf

This setup will:
- Include pgjwt in the orioledb-17 package
- Make it available as a privileged extension
- Not load it by default in the AMI or local testing in orioledb-17
- Retain it's deprecation in pg 17.4
- still enabled in pg 15.x

So the only change needed was to allow pgjwt to be built and delivered with oriole-17 package + adjusmtment of init and test sql.


migrations/schema-15.sql
```diff
@@ -1003,3 +1003,45 @@ CREATE EVENT TRIGGER pgrst_drop_watch ON sql_drop
...
+    ('00000000000002'),
+    ('00000000000003'),
+    ('10000000000000'),
+    ('20211115181400'),
+    ('20211118015519'),
+    ('20211122051245'),
+    ('20211124212715'),
+    ('20211130151719'),
+    ('20220118070449'),
+    ('20220126121436'),
```
steve-chavez: @samrose Q: Where do this migrations rows come from? And why are they necessary?

steve-chavez: Also why so many rows from a simple change?

samrose: @steve-chavez this PR is marked as a draft and so is not ready for review

you may recall this question being discussed here yesterday and answered 

https://github.com/supabase/postgres/pull/1555#discussion_r2049738715

These files are strictly automated output, and should never be edited or changed. It's common to see duplication and large amounts of output from automated file generation.

They are automated output from the https://github.com/amacneil/dbmate migration tool, which is the current tool in use for managing and updating database migrations in `supabase/postgres` 

Those particular rows are the "id" of the migrations applied as tracked by the dbmate tool

https://github.com/amacneil/dbmate?tab=readme-ov-file#schema-file

samrose: Here is where you pointed that this was answered yesterday https://github.com/supabase/postgres/pull/1554#discussion_r2049908882

samrose: I recommend looking at the documentation for dbmate linked in the repo to understand more about this tool, why it produces these files and how we use it etc https://github.com/amacneil/dbmate?tab=readme-ov-file#schema-file

steve-chavez: @samrose Sorry, although to be fair yesterday I asked about the generated files and not the `schema_migrations` rows.

I'll read the documentation of that tool.

---

That being said, there's always cognitive load from these tools. This is one of the reasons I was proposing to rely on a native pg extension instead, then we can forget about these details (if we manage to pull it through).

samrose: @steve-chavez you are right your question was different thanks for pointing that out. 

I am all in on your RFC in the meantime we have to use this approach until we can change it. Thank you for reviewing this too of course

steve-chavez (approved):

samrose:
...testing in progress

samrose:
Pause/restore test from `17.0.1.053-orioledb`  to this version `17.0.1.067-orioledb-pgjwt-2` worked


image

PR #1567, opened 2025-04-22
chore: add test for pgbouncer.get_auth

kangmingtay:
* refer to INFRA-1630
* This PR is for testing a release candidate image in local infra and staging of the changes that are in this PR: #1554

### DIFF
--- a/nix/tests/expected/pgbouncer.out
+++ b/nix/tests/expected/pgbouncer.out
@@ -58,7 +58,7 @@ SELECT
 FROM pg_proc p 
 JOIN schema_obj s ON s.oid = p.pronamespace 
 CROSS JOIN LATERAL aclexplode(p.proacl) AS acl 
-ORDER BY object_name, grantee, privilege_type; 
+ORDER BY object_name, grantee, privilege_type;
   schema   | object_name |    grantee     | privilege_type 
 -----------+-------------+----------------+----------------
  pgbouncer | get_auth    | pgbouncer      | EXECUTE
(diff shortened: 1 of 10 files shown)


nix/tests/expected/pgbouncer.out
```diff
@@ -58,11 +58,30 @@ SELECT
...
 
+-- Ensure that pgbouncer.get_auth() function does not return an expired password
+create role test_expired_user_password with login password 'expired_password' valid until '2000-01-01 00:00:00+00';
+create role test_valid_user_password with login password 'valid_password' valid until '2100-01-01 00:00:00+00';
+-- Update the pg_authid catalog directly to replace with a known SCRAM hash
+update pg_authid set rolpassword = 'SCRAM-SHA-256$4096:testsaltbase64$storedkeybase64$serverkeybase64' where rolname = 'test_valid_user_password';
+select pgbouncer.get_auth('test_expired_user_password');
+           get_auth            
+-------------------------------
+ (test_expired_user_password,)
```
steve-chavez, line 77: I realize the NULL is here but it's not that visible/readable and it'd be better to show it explicitly.

@kangmingtay How about adding `\pset null '<null>` at the start of the `pgbouncer.sql`? Then this result will show like:

```
           get_auth            
-------------------------------
 (test_expired_user_password,<null>)
```

kangmingtay: hmm it doesn't seem to work when i run `nix build .#checks.aarch64-darwin.psql_15 -L`, the nulls still show up as empty values rather than displaying the placeholder specified

kangmingtay: i think that's only a viable solution if we're using psql to run the .sql files but it seems like we're using pgtap which isn't affected by pset

steve-chavez: I see, for these we actually use pg_regress, it might have some trouble with psql options.

I've opened an issue to solve this later https://github.com/supabase/postgres/issues/1568.

steve-chavez (approved):

nix/tests/expected/pgbouncer.out
samrose: @steve-chavez the `.out` files in pg_regress tests are generated files. So, we should basically never edit those. Instead we're editing the sql in `nix/tests/sql`

But we don't want to have to maintain changes to files that are generated by the testing suite. Because, their core purpose (not just our use, but in general for anyone who uses pg_regress) is to be present to compare against when pg_regress test is run. So I'd recommend in this case, or in the case of any generated file that is not meant to be edited, to not add any edits to it. Because it will result in us needing to maintain a patch to retain the edit.

Maybe the upstream project is interested in a contribution toward making their generated output more readable?

steve-chavez: @samrose I don't understand what you mean, we always edit the .out files otherwise how do we pass tests? See https://github.com/supabase/postgres/pull/1539/files#diff-d87d64657f2572ca01422c6c6c673e0024929f13f453a91150543fd84d61573eR31-R36

We already do that and we've been doing it for a longer time on supautils.

> Maybe the upstream project is interested in a contribution toward making their generated output more readable?

That shouldn't be needed at all, we're just talking about adding a `\pset null '<null>` to a .sql file.

samrose: @steve-chavez I should have looked again sorry. I thought you were asking to manually edit the expected files that are the automated output of the tests (which is not recommended). But you asked to edit the sql which of course is fine. Sorry for my misunderstanding. 

> How about adding \pset null '<null> at the start of the pgbouncer.sql

PR #1576, opened 2025-04-30
feat: include release of ami to supported regions

samrose:

To speed up local infra testing, we will publish ami to all of the supported regions available for local testing/staging


samrose:
turns out it is already published to all regions, so we just need to look up by name to get id

PR #1581, opened 2025-05-01
feat: multi-ext-versions-pgvector

RossComputerGuy:
Similar motivation to #1549, leaving as a draft until we can figure out how to streamline some of this more. Feel free to review and take a look.


hunleyd (changes requested):
Conflicts and failing CI

PR #1582, opened 2025-05-01
feat: multi-ext-versions-pgtap

RossComputerGuy:
Serve pgtap as a multi-version extension.


RossComputerGuy:
Did a `nix flake check` and things failed but I can't seem to figure out why

PR #1590, opened 2025-05-09
feat: update rust to 1.84.0 and related needed changes

samrose:

needed to run `nix flake update rust-overlay` to bring on the code that would make rust 1.84.0 available to the nix build (this command updated just that rust-overlay input for the `flake.lock` file.)

Then, due to port conflicts on extension that rely on rust for pgrx builds, updating the input triggered a rebuild of other pg extensions (although did not change their dependencies, but it will always rebuild if inputs change version in any way). This created port conflicts for the internal tests the extensions run. So I applied a port handling to pg_graphql as well to fix that issue.

### DIFF
--- a/nix/ext/pg_graphql.nix
+++ b/nix/ext/pg_graphql.nix
@@ -27,7 +27,9 @@ buildPgrxExtension_0_12_9 rec {
   # Setting RUSTFLAGS in env to ensure it's available for all phases
   env = lib.optionalAttrs stdenv.isDarwin {
     POSTGRES_LIB = "${postgresql}/lib";
-    PGPORT = if (lib.versions.major postgresql.version) == "17" then "5440" else "5439";
+    PGPORT = toString (5430 + 
+      (if builtins.match ".*_.*" postgresql.version != null then 1 else 0) +  # +1 for OrioleDB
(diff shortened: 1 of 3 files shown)


nix/ext/pg_graphql.nix
```diff
@@ -27,7 +27,9 @@ buildPgrxExtension_0_12_9 rec {
   # Setting RUSTFLAGS in env to ensure it's available for all phases
   env = lib.optionalAttrs stdenv.isDarwin {
     POSTGRES_LIB = "${postgresql}/lib";
-    PGPORT = if (lib.versions.major postgresql.version) == "17" then "5440" else "5439";
+    PGPORT = toString (5430 + 
+      (if builtins.match ".*_.*" postgresql.version != null then 1 else 0) +  # +1 for OrioleDB
+      ((builtins.fromJSON (builtins.substring 0 2 postgresql.version)) - 15) * 2);  # +2 for each major version
```
steve-chavez, line 32: Hm, this is confusing. Why does pg_graphql needs to know anything about the PGPORT?

samrose: @steve-chavez its part of the pgrx build process and so needs to be handled for every pgrx extension at build time

PR #1592, opened 2025-05-12
feat: do not enable pgjwt by default in pg 15

samrose:

Going forward we will not enable pgjwt by default in postgres 15.x

### DIFF
--- a/ansible/tasks/stage2-setup-postgres.yml
+++ b/ansible/tasks/stage2-setup-postgres.yml
@@ -13,6 +13,10 @@
     - name: Check if psql_version is psql_17
       set_fact:
         is_psql_17: "{{ psql_version in ['psql_17'] }}"
+
+    - name: Check if psql_version is psql_15
+      set_fact:
+        is_psql_15: "{{ psql_version in ['psql_15'] }}"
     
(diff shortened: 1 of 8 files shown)


ansible/tasks/stage2-setup-postgres.yml
```diff
@@ -30,6 +34,14 @@
       when: is_psql_oriole or is_psql_17 and stage2_nix
       become: yes
 
+    - name: Remove pgjwt from supautils.conf if 15 build
+      ansible.builtin.command:
+        cmd: >
+          sed -i 's/ pgjwt,//g' 
+          /etc/postgresql-custom/supautils.conf
+      when: is_psql_15 and stage2_nix
+      become: yes
```
steve-chavez: Q: Can this be tried locally?

If not, we're planning to remove ansible later right?

samrose: @steve-chavez yes we are planning to remove ansible later. This can be tried locally, and will be part of our local testing env I'll add a change for that in this PR.

I am starting with the auth project to migrate package and config steps out of ansible. Then likely postgres will be next, fwiw.

delgado3d (approved):

PR #1604, opened 2025-05-21
fix: sync supabase roles with $POSTGRES_PASSWORD

jgoux:

Fixes DEVWF-427

In the self-hosting use case, when the `$POSTGRES_PASSWORD` is changed, the change isn't synced to all the supabase roles when the container is restarted.

## What is the new behavior?

The `$POSTGRES_PASSWORD` change is properly propagated to the supabase roles.


soedirgo (approved):

darora (approved):

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,83 @@
...
+	docker_temp_server_start "$@"
+
+	# alter the supabase_admin password
+	docker_process_sql <<-'EOSQL'
+		\set pgpass `echo "$POSTGRES_PASSWORD"`
+		ALTER USER supabase_admin WITH PASSWORD :'pgpass';
+	EOSQL
+
+	# execute the roles SQL file using docker_process_sql
+	docker_process_sql -f /docker-entrypoint-initdb.d/init-scripts/99-roles.sql
```
sweatybridge, line 20: @jgoux let's make this path an env var and only run pg_sync_password if this path is defined.

jgoux: Ok I'll make the change 👍

PR #1605, opened 2025-05-23
feat: update orioledb to beta11

samrose:

Update orioledb to beta11


samrose:
deprecate in favor of https://github.com/supabase/postgres/pull/1682

PR #1607, opened 2025-05-23
feat: support a pg 14 version of postgres bundle

samrose:

Just creating a pg 14.8 bundle to have an option to deploy an image on this version, where we need to work users to migrate forward from this version. We will not support any new instance on pg 14.x for what it's worth.


brainrake:
The pause/restore backfill work that motivated a nix-built PG 14 bundle has since concluded, and this has been conflicting since last July. Closing.

PR #1629, opened 2025-06-05
feat: multiple versions for the pgaudit extension

jfroche:
Build multiple versions of the pgaudit extension on different PostgreSQL versions.

### DIFF
--- a/nix/ext/pgaudit.nix
+++ b/nix/ext/pgaudit.nix
@@ -1,47 +1,107 @@
-{ lib, stdenv, fetchFromGitHub, libkrb5, openssl, postgresql }:
+{
+  pkgs,
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  libkrb5,
+  openssl,
+  postgresql,
+}:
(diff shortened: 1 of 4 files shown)


nix/ext/pgaudit.nix
```diff
@@ -1,47 +1,125 @@
...
+
+        mkdir -p $out/{lib,share/postgresql/extension}
+
+        # Install shared library with version suffix
+        mv ${pname}${postgresql.dlSuffix} $out/lib/${pname}-${version}${postgresql.dlSuffix}
+
+        # Install SQL files
+        sed -i '1s/^/DROP EVENT TRIGGER IF EXISTS pgaudit_ddl_command_end; \n/' *.sql
+        sed -i '1s/^/DROP EVENT TRIGGER IF EXISTS pgaudit_sql_drop; \n/' *.sql
+        sed -i 's/CREATE FUNCTION/CREATE OR REPLACE FUNCTION/' *.sql
```
steve-chavez, line 71: These seem quite invasive, why do we need them?

If they're really needed, it seems it would be more appropriate to patch the `pgaudit` migrations via `patches`?

jfroche:
Superseded by https://github.com/supabase/postgres/pull/1758

PR #1635, opened 2025-06-10
feat: allow checkout of fork repo on workflow approval

samrose:

When someone outside of the supabase organization submits a PR, and if their running of workflows is approved by a supabase maintainer after thorough review, then the gh action workflow needs to be able to checkout the code from the submitting repo.

This PR addresses the issue for Nix CI build, and may also include changes after testing for other workflows.


.github/workflows/ami-release-nix-single.yml
```diff
@@ -26,7 +26,11 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.inputs.branch }}
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          fetch-depth: 0
+          fetch-tags: true
+          token: ${{ secrets.GITHUB_TOKEN }}
```
steve-chavez: This looks duplicated in several places, which is too bad for maintenance. 

Github supports reusable workflows, can we deduplicate this using those? AFAICT it's only a matter of having another YAML file (say `shared-checkout`) which is then called in several places.

samrose: @steve-chavez you are right. Let me see what I can do.

samrose: Done

samrose:
closed/deprecated in favor of https://github.com/supabase/postgres/pull/1644

PR #1640, opened 2025-06-12
feat: enable build of multiple versions of the index_advisor extension on different PostgreSQL versions

jfroche:


samrose:
This all looks good. Doing some testing, then will approve, merge and get deployed. Thanks @jfroche !

jfroche:
replaced by #1671

PR #1651, opened 2025-06-18
feat: multiple versions for the pgjwt extension

jfroche:
Build multiple versions of the pgjwt extension on different PostgreSQL versions.

### DIFF
--- a/nix/ext/tests/pgjwt.nix
+++ b/nix/ext/tests/pgjwt.nix
@@ -1,0 +1,148 @@
+{ self, pkgs }:
+let
+  pname = "pgjwt";
+  inherit (pkgs) lib;
+  installedExtension =
+    postgresMajorVersion: self.packages.${pkgs.system}."psql_${postgresMajorVersion}/exts/${pname}-all";
+  versions = postgresqlMajorVersion: (installedExtension postgresqlMajorVersion).versions;
+  postgresqlWithExtension =
+    postgresql:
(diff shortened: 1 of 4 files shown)


nix/ext/tests/pgjwt.nix
```diff
@@ -0,0 +1,148 @@
...
+          assert installed_version == firstVersion, f"Expected ${pname} version {firstVersion}, but found {installed_version}"
+          for version in versions[pg_version][1:]:
+            run_sql(f"""ALTER EXTENSION ${pname} UPDATE TO '{version}';""")
+            installed_version = run_sql(r"""SELECT extversion FROM pg_extension WHERE extname = '${pname}';""")
+            assert installed_version == version, f"Expected ${pname} version {version}, but found {installed_version}"
+
+      start_all()
+
+      server.wait_for_unit("multi-user.target")
+      server.wait_for_unit("postgresql.service")
```
steve-chavez, line 130: Q: Can I try these tests locally? Do they run automatically with `nix flake check`? Also, is there a way to run only the pgjwt tests?

samrose: @steve-chavez that is the ultimate goal in part. We may need to add support for darwin, but they should run now on a physical linux machine

jfroche:
Superseded by https://github.com/supabase/postgres/pull/1751

PR #1652, opened 2025-06-18
ci: Use larger runner for other workflows and wait for the nix workflow to succeed

jfroche:
Use larger runner for the test and other workflows.
Wait for the nix workflow to succeed before running the test and testinfra-ami-build workflow

### DIFF
--- a/.github/workflows/ami-release-nix-single.yml
+++ b/.github/workflows/ami-release-nix-single.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: arm-runner
+    runs-on: large-linux-arm
     timeout-minutes: 150
 
     steps:
--- a/.github/workflows/ami-release-nix-single.yml
+++ b/.github/workflows/ami-release-nix-single.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Generate common-nix.vars.pkr.hcl
         run: |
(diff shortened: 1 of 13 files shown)


jfroche:
Note that we don't see the builds for the workflows depending on `workflow_run` as the `workflow_run` need to be present on the default branch (documentation: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run)

Manual workflow run:
- [Test Database](https://github.com/supabase/postgres/actions/runs/15743245799)
-

samrose (changes requested):
@jfroche I think we just need to apply these to 

- https://github.com/supabase/postgres/blob/develop/.github/workflows/ami-release-nix.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/dockerhub-release-matrix.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/qemu-image-build.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/ami-release-nix-single.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/publish-migrations-prod.yml
- https://github.com/supabase/postgres/blob/develop/.github/workflows/publish-migrations-staging.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/publish-nix-pgupgrade-bin-flake-version.yml 
- https://github.com/supabase/postgres/blob/develop/.github/workflows/publish-nix-pgupgrade-scripts.yml

as well, unless you see a reason we shouldn't?

samrose (approved):

PR #1653, opened 2025-06-19
chore: add `treefmt` config for nix files

yvan-sraka:
* add a ci job that run `treefmt`
* make it `nix fmt` default formatter
* add dumb simple `pre-commit` and `editorconfig`

### DIFF
--- a/.github/workflows/treefmt.yml
+++ b/.github/workflows/treefmt.yml
@@ -1,0 +1,15 @@
+name: treefmt
+on:
+    pull_request:
+    push:
+        branches: develop
+jobs:
+    formatter:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: cachix/install-nix-action@v31
+              with:
+                  nix_path: nixpkgs=channel:nixos-unstable
+            - name: treefmt
+              run: nix-shell -p treefmt nixpkgs-fmt --run "treefmt --ci"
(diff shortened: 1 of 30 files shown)


.github/workflows/treefmt.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+    formatter:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - uses: cachix/install-nix-action@v31
+              with:
+                  nix_path: nixpkgs=channel:nixos-unstable
+            - name: treefmt
+              run: nix-shell -p treefmt nixpkgs-fmt --run "treefmt --ci"
```
yvan-sraka, line 15: IIRC `nix fmt` doesn't have a "check" mode, so here I call `treefmt --ci` directly.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+repos:
+  - repo: local
+    hooks:
```
yvan-sraka, line 3: Maybe this is a bit overkill, but I find it a convenient way to set up a git commit hook, and so catch you forgot to run `nix fmt` (if your editor isn't doing it for you) before CI!

samrose (changes requested):
@yvan-sraka this looks great. Can you please add a bit in the nix/docs section to give some clues to people who are not as deeply embedded in Nix tooling etc how to use what was added here (maybe they need to add a plugin for their editor, anything else to know once this is merged and added)

yvan-sraka:
This has been mostly _(maybe it just miss an `editorconfig` and a `treefmt` one for user not using the nix devshell?)_ already tackled by https://github.com/supabase/postgres/pull/1681

PR #1660, opened 2025-06-20
feat: multiple versions for the pgsodium extension

jfroche:
Build multiple versions of the pgsodium extension on different PostgreSQL versions.


hunleyd (dismissed):
rebase to fix conflicts?

samrose (approved):

PR #1661, opened 2025-06-20
feat: multiple versions for the vault extension

jfroche:
Build multiple versions of the vault extension on different PostgreSQL versions.


hunleyd (changes requested):
rebase to fix conflicts?

hunleyd (approved):

samrose (approved):

PR #1664, opened 2025-06-23
feat: support multiple versions of the http extension

jfroche:
Build multiple versions of the http extension on different PostgreSQL versions. Add test for the extensions and their upgrade on PostgreSQL 15 and 17.


hunleyd (dismissed):
rebase to fix conflicts?

samrose (approved):

PR #1666, opened 2025-06-23
fix: allow postgres to grant storage privs to other roles

soedirgo:
Temporary fix until STORAGE-211 is completed


migrations/db/migrations/20250623125453_tmp_grant_storage_tables_to_postgres_with_grant_option.sql
```diff
@@ -0,0 +1,6 @@
+-- migrate:up
+-- TODO: remove this migration once STORAGE-211 is completed
+-- DRI: bobbie
+grant all on storage.buckets, storage.objects to postgres with grant option;
```
steve-chavez, line 4: I suggest an ALTER DEFAULT PRIVILEGES instead, to not worry about future objects created on the `storage` schema:

```sql
alter default privileges for user supabase_storage_admin in schema storage grant all
    on tables to postgres with grant option;
```

soedirgo: We don't want all Storage tables to have these grants, only these ones.

steve-chavez (approved):

PR #1668, opened 2025-06-23
feat: support multiple versions of the pgmq extension

jfroche:
Build multiple versions of the pgmq extension on different PostgreSQL versions.


nix/ext/versions.json
za-arthur, line 1: Would it make sense to add new releases since `1.5.1`?
https://github.com/pgmq/pgmq/releases

hunleyd (dismissed):
rebase to fix conflicts?

samrose (approved):

PR #1671, opened 2025-06-24
feat: support multiple versions of the index_advisor extension

jfroche:
Build multiple versions of the index_advisor extension on different PostgreSQL versions.

### DIFF
--- a/nix/ext/versions.json
+++ b/nix/ext/versions.json
@@ -10,5 +10,14 @@
       "pgrx": "0.14.3",
       "rust": "1.87.0"
     }
+  },
+  "index_advisor": {
+    "0.2.0": {
+      "postgresql": [
+        "15",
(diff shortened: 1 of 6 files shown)


nix/ext/versions.json
```diff
@@ -10,5 +10,15 @@
       "pgrx": "0.14.3",
       "rust": "1.87.0"
     }
+  },
+  "index_advisor": {
+    "0.2.0": {
+      "postgresql": [
+        "15",
+        "17",
+        "orioledb-17"
```
samrose, line 19: I added orioledb-17

samrose: ..other changes are needed to do this properly. Coming in a follow up PR

samrose (approved):

PR #1673, opened 2025-06-25
Sam/revert wandelf

samrose:
## What kind of change does this PR introduce?

just a temporary revert of these features until we rename it.

PR #1674, opened 2025-06-25
fix: more gandalf pathname

soedirgo:
Not quite there yet https://github.com/supabase/postgres/actions/runs/15882639321/job/44786792775#step:14:64


samrose:
@soedirgo may we close this one as https://github.com/supabase/postgres/pull/1673 reverts it. Infra team will change the name of the project and then create a new PR

PR #1679, opened 2025-06-27
feat: add the supabase admin agent to the ami build

Crispy1975:
This PR adds the new `supabase-admin-agent` to the AMI build process. This binary helps to manage various aspects of a running Supabase project.

### DIFF
--- a/ansible/files/permission_check.py
+++ b/ansible/files/permission_check.py
@@ -94,10 +94,16 @@ expected_results = {
     "systemd-coredump": [
         {"groupname": "systemd-coredump", "username": "systemd-coredump"}
     ],
+    "supabase-admin-agent": [
+        {"groupname": "supabase-admin-agent", "username": "supabase-admin-agent"},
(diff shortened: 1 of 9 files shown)


ansible/files/permission_check.py
```diff
@@ -94,11 +94,17 @@
     "systemd-coredump": [
         {"groupname": "systemd-coredump", "username": "systemd-coredump"}
     ],
+    "supabase-admin-agent": [
```
samrose, line 97: don't forget to also add to usernames list too 

https://github.com/supabase/postgres/blob/develop/ansible/files/permission_check.py#L200

samrose (approved):
other than adding name to list submitted on query, looks good

ansible/files/permission_check.py
Crispy1975: Ty! Added that in.

pcnc (approved):

PR #1684, opened 2025-06-30
feat: support multiple versions of the plpgsql_check extension

jfroche:
Build multiple versions of the plpgsql_check extension on different PostgreSQL versions.


hunleyd (dismissed):
rebase to fix conflicts?

samrose (approved):

PR #1709, opened 2025-07-17
feat: disable default database collector for postgres_exporter

darora:
The database collector emits a metric for the size of each database by
default. We emit a metric that represents the sum of all database,
which suffices for our purposes. This change removes the redundant
per-database metric.

### DIFF
--- a/ansible/files/postgres_exporter.service.j2
+++ b/ansible/files/postgres_exporter.service.j2
@@ -3,7 +3,7 @@ Description=Postgres Exporter
 
 [Service]
 Type=simple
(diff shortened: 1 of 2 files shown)


thebengeu (approved):

tgallacher (commented):
Will this impact v2 negatively? 

Assuming the "database collector" is not a v3 specific component?

darora:
I'm pretty sure this was inadvertently left out in #788 ("all the collectors are disabled with the additions to the CLI switches in the systemd unit file").
While it does represent a change, the metrics returned do not represent a stable API, and as such, imo it's fine to disable it.

However, out of an abundance of caution, I've updated the PR to only disable it in the QEMU artifact.

PR #1732, opened 2025-08-04
Update supautils.conf.j2 | removed users permission to update log_sta…

TheOtherBrian1:

Modifies supautils.

Users can change `log_statements`

## What is the new behavior?

Revokes users' permission to change the `log_statements` 

## Additional context

log_statements should be set to `ddl`. It was assumed that users would expand it to `ddl, write,...`, but not remove the default setting. A [Team Customer's instance](https://supabase.frontapp.com/open/msg_2sg2mkda?key=lzqj7UZM08CyudVUraFdOeJv7GfTmJWG) had a setting of `none`. As a result, uncovering why objects have been dropped has become nearly impossible. 

Ensuring DDL commands are logged is essential for both users and Support for uncovering hacks, rogue AI queries, and internal vandalism between project owners.


soedirgo:
This is a difficult move to justify because it's a breaking change and it prevents higher levels of logging (which may in turn result in more tickets). Looking into the ticket to get more details

steve-chavez:
The ideal solution would be to disable the `none` value instead. supautils has a feature that disable values for GUC placeholders ([ref1](https://github.com/supabase/supautils/blob/38094ff6b118fff05fdc3754cd40c17c5e2edee7/test/expected/restricted_placeholders.out#L8-L12) and [ref2](https://github.com/supabase/supautils/blob/master/src/supautils.c#L1241-L1259)), but this won't work for already defined GUCs IINM. We would need to see if that can be made to work somehow.

soedirgo:
> The ideal solution would be to disable the none value instead

It'd still be a breaking change.

Root cause has been addressed in the linked ticket, so closing this.

PR #1734, opened 2025-08-06
fix: hashicorp moved packer plugins

tomashley:
https://discuss.hashicorp.com/t/important-update-official-packer-plugin-distribution-moving-to-releases-hashicorp-com/75972

Plugins moved, affecting amazon plugin location


tomashley:
Unnecessary as this really relies on a newer version of packer on the action runners

PR #1735, opened 2025-08-06
feat: source specific version of packer and build with specific go

samrose:
use in all workflows except qemu-image-build.yml

Currenty a version of packer is baked into hosted builders.

This change will use a specific version of Go, and build a specific version of packer from source, cache in our binary cache, and use just that version of packer in all workflows and devenvs going forward.

### DIFF
--- a/.github/workflows/ami-release-nix-single.yml
+++ b/.github/workflows/ami-release-nix-single.yml
@@ -53,25 +53,22 @@ jobs:
           # Ensure there's a newline at the end of the file
           echo "" >> common-nix.vars.pkr.hcl
 
-      - name: Install Packer
-        run: nix profile install github:supabase/postgres/${{ steps.get_sha.outputs.sha }}#packer
-
       - name: Build AMI stage 1
         env:
           POSTGRES_MAJOR_VERSION: ${{ env.POSTGRES_MAJOR_VERSION }}
         run: |
-          packer init amazon-arm64-nix.pkr.hcl
           GIT_SHA=${{ steps.get_sha.outputs.sha }}
-          packer build -var "git-head-version=${GIT_SHA}" -var "packer-execution-id=${GITHUB_RUN_ID}"  -var-file="development-arm.vars.pkr.hcl" -var-file="common-nix.vars.pkr.hcl" -var "ansible_arguments=-e postgresql_major=${POSTGRES_MAJOR_VERSION}"  amazon-arm64-nix.pkr.hcl
+          sudo nix run github:supabase/postgres/${GIT_SHA}#packer -- init amazon-arm64-nix.pkr.hcl
+          sudo nix run github:supabase/postgres/${GIT_SHA}#packer -- build -var "git-head-version=${GIT_SHA}" -var "packer-execution-id=${GITHUB_RUN_ID}"  -var-file="development-arm.vars.pkr.hcl" -var-file="common-nix.vars.pkr.hcl" -var "ansible_arguments=-e postgresql_major=${POSTGRES_MAJOR_VERSION}"  amazon-arm64-nix.pkr.hcl
 
       - name: Build AMI stage 2
         env:
           POSTGRES_MAJOR_VERSION: ${{ env.POSTGRES_MAJOR_VERSION }}
         run: |
-          packer init stage2-nix-psql.pkr.hcl
           GIT_SHA=${{ steps.get_sha.outputs.sha }}
+          sudo nix run github:supabase/postgres/${GIT_SHA}#packer -- init stage2-nix-psql.pkr.hcl
           POSTGRES_MAJOR_VERSION=${{ env.POSTGRES_MAJOR_VERSION }}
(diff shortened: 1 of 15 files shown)


soedirgo (approved):

.github/workflows/ami-release-nix-single.yml
```diff
@@ -40,7 +40,13 @@ jobs:
         run: |
           echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
 
-      - uses: DeterminateSystems/nix-installer-action@main
+      - name: Install nix
+        uses: cachix/install-nix-action@v27
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.29.1/install
```
jfroche, line 46: We should be able to use 2.30.2.
2.30.1 has the fix we need for macos sandboxing (https://github.com/NixOS/nix/pull/13458).
We would benefit from the speed improvements from 2.30 (https://discourse.nixos.org/t/nix-2-30-0-released/66449#p-221238-performance-improvements-3).
Note that we could do another PR to test this.

.github/workflows/ami-release-nix-single.yml
```diff
@@ -40,7 +40,13 @@ jobs:
         run: |
           echo "sha=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
 
-      - uses: DeterminateSystems/nix-installer-action@main
+      - name: Install nix
+        uses: cachix/install-nix-action@v27
```
jfroche, line 44: We might want to update that action to `31.5.2`.

.github/workflows/ami-release-nix-single.yml
```diff
@@ -57,18 +63,18 @@ jobs:
         env:
           POSTGRES_MAJOR_VERSION: ${{ env.POSTGRES_MAJOR_VERSION }}
         run: |
-          packer init amazon-arm64-nix.pkr.hcl
           GIT_SHA=${{ steps.get_sha.outputs.sha }}
-          packer build -var "git-head-version=${GIT_SHA}" -var "packer-execution-id=${GITHUB_RUN_ID}"  -var-file="development-arm.vars.pkr.hcl" -var-file="common-nix.vars.pkr.hcl" -var "ansible_arguments=-e postgresql_major=${POSTGRES_MAJOR_VERSION}"  amazon-arm64-nix.pkr.hcl
+          nix run github:supabase/postgres/${GIT_SHA}#packer -- init amazon-arm64-nix.pkr.hcl
```
jfroche, line 67: We already have the checkout repo, we should be able to run:
```suggestion
          nix run .#packer -- init amazon-arm64-nix.pkr.hcl
```

.github/workflows/qemu-image-build.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           make init
           GIT_SHA=${{github.sha}}
           export PACKER_LOG=1
-          packer build -var "git_sha=${GIT_SHA}" -var-file="common-nix.vars.pkr.hcl" qemu-arm64-nix.pkr.hcl
+          sudo nix run github:supabase/postgres/${GIT_SHA}#packer -- build -var "git_sha=${GIT_SHA}" -var-file="common-nix.vars.pkr.hcl" qemu-arm64-nix.pkr.hcl
```
jfroche: Do we need sudo here ?

nix/packages/packer.nix
```diff
@@ -0,0 +1,50 @@
+{
+  pkgs,
+  inputs,
+  lib,
+  fetchFromGitHub,
+  installShellFiles,
+  ...
+}:
+let
+  go124 = inputs.nixpkgs-go124.legacyPackages.${pkgs.system}.go_1_24;
```
jfroche, line 10: note for #1714 - remove that when rebase

nix/packages/packer.nix
```diff
@@ -0,0 +1,50 @@
...
+  buildGoModule = pkgs.buildGoModule.override { go = go124; };
+in
+buildGoModule rec {
+  pname = "packer";
+  version = "1.14.1";
```
jfroche, line 15: We could upstream this as well

.github/workflows/ami-release-nix-single.yml
samrose: I agree. I'll do it as another PR. This one is to unblock other teams with the packer update issue.

.github/workflows/qemu-image-build.yml
samrose: We actually should not. On this one, I will revert to just using the packer that is baked in, and update it with the team working on qemu build when they are ready

.github/workflows/ami-release-nix-single.yml
samrose: I'll do this in the follow up to bump nix to 2.30.x

.github/workflows/ami-release-nix-single.yml
samrose: I agree, although I had been using this approach to help people in the company and community who are not very deeply knowledgeable with Nix understand more explicitly what is happening/intended with calls like this one.

nix/packages/packer.nix
samrose: Ok, will do

PR #1737, opened 2025-08-07
Multi version ext/pg cron

samrose:

pg_cron multi version following the new nix code pattern


jfroche (commented):
For other extensions which require other file than the nix file, we move the extension package into `<package name>/default.nix`. 
So `nix/ext/pg_cron.nix` would become `nix/ext/pg_cron/default.nix`.
Maybe it is the right moment to do so here too.

nix/ext/pg_cron.nix
```diff
@@ -5,31 +5,179 @@
...
+        description = "Run Cron jobs through PostgreSQL";
+        homepage = "https://github.com/citusdata/pg_cron";
+        changelog = "https://github.com/citusdata/pg_cron/raw/v${version}/CHANGELOG.md";
+        platforms = postgresql.meta.platforms;
+        license = licenses.postgresql;
+      };
+    };
+  packages = builtins.attrValues (lib.mapAttrs (name: value: build name value) supportedVersions);
+in
+stdenv.mkDerivation {
```
jfroche: In other extensions, we have used `buildEnv` to link the packages together.

nix/ext/pg_cron.nix
```diff
@@ -5,31 +5,179 @@
...
+      cat "$latest_control" >> $out/share/postgresql/extension/${pname}.control
+    fi
+
+    # Library symlink - only if it doesn't exist
+    if [ ! -f "$out/lib/${pname}${postgresql.dlSuffix}" ]; then
+      ln -sfnv ${pname}-$latest_version${postgresql.dlSuffix} $out/lib/${pname}${postgresql.dlSuffix}
+    fi
+
+    # Create version switcher script
+    cat > $out/bin/switch_pg_cron_version <<'EOF'
```
jfroche: We might want to do the same change we did for pg_net with the overlay and the writeShellApplication (here: https://github.com/supabase/postgres/pull/1744/files#diff-eacc41da93c540ef366640210fb5278d612bcd2f397e5cb247c5a02ebb1d55e2R62).

jfroche (commented):
I would also add the nixos test to verify the extension migration between versions as well as the postgresql upgrade like we did in pg_net.

samrose:
Passed all pause/restore, upgrade and functional tests.

Crispy1975 (approved):

PR #1739, opened 2025-08-08
Set-up `nix-github-actions`

yvan-sraka:


yvan-sraka:
Replaced by https://github.com/supabase/postgres/pull/1742

PR #1742, opened 2025-08-08
chores: set up `nix-github-actions`

yvan-sraka:


samrose (changes requested):
This all looks really good and useful

There seems to be some vestigal running of testinfra failing that shows up in the checks tab 
image


However those same tests seem to have passed when run as a child of Nix CI

image

What do you think about that issue?

yvan-sraka:
Supersede by https://github.com/supabase/postgres/pull/1745

PR #1743, opened 2025-08-08
 feat: use a reworked generic nixos test for `wrappers` extension

yvan-sraka:
This is a follow-up of https://github.com/supabase/postgres/pull/1637 and https://github.com/supabase/postgres/pull/1663

### DIFF
--- a/nix/github-actions.nix
+++ b/nix/github-actions.nix
@@ -1,6 +1,18 @@
 { inputs, ... }:
+let
+  githubPlatforms = {
+    "x86_64-linux" = "large-linux-x86";
+    "aarch64-linux" = "large-linux-arm";
+    "aarch64-darwin" = "macos-latest-xlarge";
+  };
+in
 {
   flake.githubActions = inputs.nix-github-actions.lib.mkGithubMatrix {
-    checks = inputs.nixpkgs.lib.getAttrs [ "x86_64-linux" ] inputs.self.checks;
+    checks = inputs.nixpkgs.lib.getAttrs [
+      "x86_64-linux"
+      "aarch64-linux"
+      "aarch64-darwin"
+    ] inputs.self.checks;
+    platforms = githubPlatforms;
   };
 }


hunleyd (approved):

jchancojr (approved):
lgtm

.github/workflows/nix-build.yml
```diff
@@ -104,23 +113,12 @@ jobs:
           sudo rm -rf /tmp/* 2>/dev/null || true
           echo "=== AFTER CLEANUP ==="
           df -h
-      - name: Build psql bundle
-        run: >
-          nix run "github:Mic92/nix-fast-build?rev=b1dae483ab7d4139a6297e02b6de9e5d30e43d48" 
-          -- --skip-cached --no-nom ${{ matrix.runner == 'macos-latest-xlarge' && '--max-jobs 1' || '' }}
-          --flake ".#checks.$(nix eval --raw --impure --expr 'builtins.currentSystem')"
+      - run: nix build -L '.#${{ matrix.attr }}'
```
samrose: @yvan-sraka sorry if I missed the discussion somewhere, in this approach do we move nix-fast-build to some other level, or do we just not need it any more

samrose (changes requested):

.github/workflows/nix-build.yml
yvan-sraka: We use `nix-fast-build` when we run `nix develop` _(as it's nice to quickly have a working developer environment)_. However, in CI, it's better, IMHO, to have jobs run sequentially, particularly when you want to reproduce an issue in CI _(if your build fails in multiple places, it's impractical to assert that you fixed one issue if they occur out of order)_. WDYT?

yvan-sraka: But it's indeed different if the runner is not ephemeral and has some persistent `/nix/store`. In that case, `nix-fast-build` has some value since it avoids some redownloading (it partially caches even failed builds). It might make sense to reintroduce it when we have the new runners :)

PR #1744, opened 2025-08-08
feat: multi version pg_net including 0.19.5

samrose:

Build multiple versions of the pg_net extension to on different PostgreSQL versions.
Add test for the extensions and their upgrade on PostgreSQL 15 and 17. bumps version to 0.19.5 for testing as well


samrose:
testing pause/restore and other e2e scenarios now

samrose:
Just waiting on a branch of infra to test pause/restore dynamics on

nix/ext/pg_net.nix
```diff
@@ -1,41 +1,202 @@
...
-    postgresql
-  ];
-
-  src = fetchFromGitHub {
-    owner = "supabase";
-    repo = pname;
-    rev = "refs/tags/v${version}";
-    hash = "sha256-c1pxhTyrE5j6dY+M5eKAboQNofIORS+Dccz+7HKEKQI=";
+let
+  switchPgNetVersion = writeShellApplication {
```
jfroche, line 62: I wonder if the script is trying to replace symlinks in the nix store ? Maybe we copy postgres lib/extension directory before ?

samrose: @jfroche good call. I talked with @jfroche today and he's going to propose a better way to do this for the short term for all packages with background worker.

jfroche: I have added a change for pg_net. We will be able to refactor for the next extension using a background worker.

nix/ext/pg_net.nix
```diff
@@ -1,41 +1,224 @@
...
+  pname = "pg_net";
+  build =
+    version: hash:
+    stdenv.mkDerivation rec {
+      inherit pname version;
+
+      buildInputs = [
+        curl
+        postgresql
+      ] ++ lib.optional (version == "0.6") libuv;
```
steve-chavez, line 123: Q: Is it necessary to support such and old version of the extension? The `libuv` version was not stable and I'm not sure if it was actually deployed on Supabase. Or if any project with such an old version survives.

samrose: @steve-chavez we'll take a look again at the projects, and if we don't need this we could drop it. Thanks!

nix/ext/pg_net.nix
```diff
@@ -1,41 +1,224 @@
...
+        else
+          cp sql/${pname}--${version}.sql $out/share/postgresql/extension/${pname}--${version}.sql
+        fi
+
+        # Install upgrade scripts
+        find . -name '${pname}--*--*.sql' -exec install -Dm644 {} $out/share/postgresql/extension/ \;
+
+        # Create versioned control file with modified module path
+        sed -e "/^default_version =/d" \
+            -e "s|^module_pathname = .*|module_pathname = '\$libdir/${pname}'|" \
```
za-arthur, line 167: This will not set `module_pathname` since `pg_net` itself doesn't set it:
https://github.com/supabase/pg_net/blob/master/pg_net.control.in

samrose: @za-arthur great catch and thank you! 

I am attempting to resolve it with https://github.com/supabase/postgres/pull/1744/commits/4996d81a8596ffbb903a8eceb2b1c8d0e153bc7c and will test and verify that that approach does everything we need. Feel free to continue to feedback on this of course.

samrose: I was wrong: https://github.com/supabase/postgres/commit/4996d81a8596ffbb903a8eceb2b1c8d0e153bc7c  won't make a difference because the db will have a value of `pg_net` for `probin` however, the script that creates aliases should solve this issue for any extension with this problem, and would need that coverage even if the extensions are fixed upstream (due to values in existing databases)

samrose: ...reverted with https://github.com/supabase/postgres/pull/1744/commits/381035f287a85287f94c63f517be230058ef30ac

Crispy1975 (approved):

nix/ext/pg_net.nix
```diff
@@ -1,41 +1,224 @@
...
+          rm sql/pg_net--0.2--0.3.sql
+          rm sql/pg_net--0.4--0.5.sql
+          rm sql/pg_net--0.5.1--0.6.sql
+        ''
+        + lib.optionalString (version == "0.7.1") ''
+          # handle collision with pg_net 0.10.0
+          rm sql/pg_net--0.5.1--0.6.sql
+        '';
+
+      env.NIX_CFLAGS_COMPILE = "-Wno-error";
```
steve-chavez: Shouldn't be needed starting from 0.19.1, check
https://github.com/NixOS/nixpkgs/pull/425390/files

valigula (approved):
LGTM

nix/ext/pg_net.nix
samrose: resolved in https://github.com/supabase/postgres/pull/1744/commits/146bd2787777e185417f488378b8c5371d22d4cd

PR #1748, opened 2025-08-11
feat: add older versions of the `wrappers` extension

yvan-sraka:
This PR add a few commits on top of https://github.com/supabase/postgres/pull/1743

## Summary by CodeRabbit

* **Chores**
  * Appended a suffix to select PostgreSQL version strings and expanded the wrappers registry with many new releases and metadata.
  * Broad dependency and packaging updates across multiple wrapper releases, regenerating lockfiles and refreshing supported toolchains and PG targets.
* **New Features**
  * Added a CLI helper to discover and maintain extension-version mappings automatically.
* **Documentation**
  * Added guidance for automating version updates with example usage and verification steps.


nix/ext/versions.json
```diff
@@ -504,84 +504,30 @@
...
+    "0.2.0": {
       "postgresql": [
         "15"
       ],
-      "hash": "sha256-ut3IQED6ANXgabiHoEUdfSrwkuuYYSpRoeWdtBvSe64=",
+      "hash": "sha256-F+S5uyubL3Tb3RTJ08Zf9gN8oLE/WkCWFA8RcKkDqes=",
       "pgrx": "0.11.3",
       "rust": "1.76.0"
     },
-    "0.4.3": {
```
yvan-sraka, line 616: (WIP) I just removed temporarily those versions while working on it, don't merge as is!

nix/ext/wrappers/default.nix
```diff
@@ -161,23 +161,92 @@ let
         };
       }
       // lib.optionalAttrs (version == "0.3.0") {
-        patches = [ ./0001-bump-pgrx-to-0.11.3.patch ];
```
jfroche, line 169: We do the same change for versions <= 0.3.0, maybe there is a way to have one condition here ?
I guess it will be the same for the older versions (0.1.7 -> 0.1.14) that are not here yet.

nix/ext/versions.json
jfroche: Don't forget to add them back

nix/ext/scripts/update_versions_json.py
jfroche, line 1: We might want to move it to `nix/ext/scripts` instead so that `nix/ext` only contains the extensions

nix/ext/scripts/update_versions_json.py
```diff
@@ -0,0 +1,90 @@
+#! /usr/bin/env nix-shell
+#! nix-shell -i python3 -p python3 git nix-prefetch-git python3Packages.packaging
+
```
jfroche, line 3: Add a few lines of documentation on what is does and maybe an example of how we can use it

nix/ext/wrappers/default.nix
```diff
@@ -194,7 +263,9 @@ let
   numberOfPreviouslyPackagedVersions = builtins.length previouslyPackagedVersions;
   allVersions = (builtins.fromJSON (builtins.readFile ../versions.json)).wrappers;
   supportedVersions = lib.filterAttrs (
-    _: value: builtins.elem (lib.versions.major postgresql.version) value.postgresql
+    _: value:
+    (!value ? ignore || value.ignore != true)
```
jfroche, line 181: Do we still need this ?

yvan-sraka: I believe it's still convenient to have it, when generating hashes in `nix/ext/version.json` with `nix/ext/update_versions_json.py`.

nix/ext/wrappers/default.nix
yvan-sraka: Actually I didn't patch `pgrx` for older versions, and I consider even in the end to not patch it for `0.2.0` or `0.3.0` and to rather add more versions to `nix/cargo-pgrx/versions.json`, WDYT?

nix/ext/scripts/update_versions_json.py
yvan-sraka: Yes indeed, should I put that in `nix/docs/adding-new-package.md` or `nix/docs/update-extension.md` or both?

nix/ext/wrappers/default.nix
jfroche: If the whole block is the same (including the cargo lock file and/or the patch) for all versions <= 0.3.0 we should only have 1 block

nix/ext/scripts/update_versions_json.py
jfroche: `adding-new-package.md` needs a complete rewrite, I would add it in `update-extension.md` first and do another PR/TODO for improving overall documentation (removing the old way) on how to create/update multi version extensions

samrose:
we'll re-open if we end up using this

PR #1749, opened 2025-08-11
feat: multiple versions for the `timescaledb-apache` extension

yvan-sraka:
Build multiple versions of the `timescaledb-apache` extension on PostgreSQL 15.


nix/ext/tests/default.nix
```diff
@@ -202,7 +186,7 @@ builtins.listToAttrs (
       "index_advisor"
       "pg_cron"
       "pg_net"
-      "timescaledb-apache"
+      "timescaledb"
```
jfroche: We can remove that from this list. We don't want to test timescaledb with the default nixos test

nix/tests/timescaledb.nix
```diff
@@ -0,0 +1,100 @@
...
+                  run_sql(f"""ALTER EXTENSION timescaledb UPDATE TO '{version}';""")
+                  installed_version = run_sql(r"""SELECT extversion FROM pg_extension WHERE extname = 'timescaledb';""")
+                  assert installed_version == version, f"Expected timescaledb version {version}, but found {installed_version}"
+
+          start_all()
+
+          server.wait_for_unit("multi-user.target")
+          server.wait_for_unit("postgresql.service")
+
+          check_upgrade_path()
```
jfroche: This should be indented at the same level as the def otherwise the test won't do anything.

nix/ext/timescaledb.nix
```diff
@@ -98,6 +100,16 @@ let
   packages = builtins.attrValues (
     lib.mapAttrs (name: value: build name value.hash (value.revision or name)) supportedVersions
   );
+  switch-timescaledb-loader = writeShellApplication {
+    name = "switch_timescaledb_loader";
+    runtimeInputs = [ coreutils ];
+    text = ''
+      EXT_LOADER_TO_USE="$EXT_WRAPPER_LIB/$EXT_NAME-loader-$VERSION${postgresql.dlSuffix}"
+      if [ -f "$EXT_LOADER_TO_USE" ]; then
```
samrose, line 108: unfortunately we cannot link to `EXT_LOADER_TO_USE="$EXT_WRAPPER_LIB/$EXT_NAME-loader-$VERSION${postgresql.dlSuffix}"` 

Because timescale will look for the loader at `timescaledb.so` when it starts so we have to link the versioned loader to timescaledb.so

samrose: trying to resolve with https://github.com/supabase/postgres/pull/1749/commits/c7604f1aa810aa8626103048f1ee1368a4ef3bae

samrose (approved):

PR #1753, opened 2025-08-12
feat: deploy envoy using system manager

jfroche:
Add the required tests to verify that the envoy service can be deployed using system manager.


brainrake:

PR #1754, opened 2025-08-12
tests: create a tmp image that has coredumps enabled for testing

samrose:
### these changes will NOT be merged into develop or promoted to production

The purpose of this draft pr is to create an image that has coredumps enabled per our internal playbook, for testing segfault issues. We will not merge this PR, or release the images created with it. It will be closed without merging when testing is completed.

### DIFF
--- a/ansible/tasks/setup-coredumps.yml
+++ b/ansible/tasks/setup-coredumps.yml
@@ -1,0 +1,219 @@
+---
+# Setup coredump collection for debugging PostgreSQL crashes
+# This configuration is temporary and should be cleaned up after debugging
+
+- name: Create coredump directory
+  become: yes
+  file:
+    path: /pg_coredump_debug
+    state: directory
+    mode: '0777'
+    owner: root
+    group: root
+  when: stage2_nix
+
+- name: Configure kernel core pattern
+  become: yes
+  ansible.posix.sysctl:
+    name: kernel.core_pattern
+    value: '/pg_coredump_debug/core.%e.%p.%t'
+    state: present
+    sysctl_file: /etc/sysctl.d/99-coredump.conf
+    reload: yes
+  when: stage2_nix
+
+- name: Enable core dumps with PID
+  become: yes
+  ansible.posix.sysctl:
+    name: kernel.core_uses_pid
+    value: '1'
+    state: present
+    sysctl_file: /etc/sysctl.d/99-coredump.conf
+    reload: yes
+  when: stage2_nix
+
+- name: Enable SUID dumpable
+  become: yes
+  ansible.posix.sysctl:
+    name: fs.suid_dumpable
+    value: '1'
+    state: present
+    sysctl_file: /etc/sysctl.d/99-coredump.conf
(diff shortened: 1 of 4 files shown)


ansible/tasks/setup-coredumps.yml
```diff
@@ -0,0 +1,230 @@
...
+      
+      ## Coredump Location:
+      Coredumps are saved to: /pg_coredump_debug/
+      
+      ## Debug Symbols:
+      Debug symbols and source files are already installed via nix during the build.
+      To find the paths:
+      ```
+      sudo -u postgres nix profile list | grep postgresql_.*_debug
+      sudo -u postgres nix profile list | grep postgresql_.*_src
```
za-arthur: I had to add `-I`, otherwise I get error `sudo: nix: command not found`:
```suggestion
      sudo -i -u postgres nix profile list | grep postgresql_.*_debug
      sudo -i -u postgres nix profile list | grep postgresql_.*_src
```

ansible/tasks/setup-coredumps.yml
```diff
@@ -0,0 +1,230 @@
...
+      ```
+      
+      ## Using GDB:
+      To analyze a coredump:
+      ```
+      sudo gdb postmaster -c /pg_coredump_debug/<core_file>
```
za-arthur, line 211: I had to pass full path instead of just `postmaster`:

```
sudo gdb /nix/store/56va3gd07cakgapybcw4iq1irhx9lryx-postgresql-and-plugins-17_11/bin/.postgres-wrapped -c /pg_coredump_debug/core..postgres-wrapp.2150.1755075766
```

I'm not sure how properly can I find the path. One can try:
```
ls -lh /nix/store/*/bin/.postgres-wrapped
-r-xr-xr-x 3 root root 11M Jan  1  1970 /nix/store/56va3gd07cakgapybcw4iq1irhx9lryx-postgresql-and-plugins-17_11/bin/.postgres-wrapped
```

samrose: I think you can also find it in /var/lib/postgresql/.nix-profile/bin so there actually is already a conventional location (I just used the same "home" dir for the postgres user that infra team created before I did all the nix packaging, to make it easier for people to orient themselves who were already familiar with /var/lib/postgresql homedir)

samrose:
We'll end up converting this from ansible to a nix package configuration that can be installed/uninstalled this was just a short term measure/test to see if this would work

ansible/tasks/setup-coredumps.yml
za-arthur: In my case I had to use the `.postgres-wrapped` file which isn't in `/var/lib/postgresql/.nix-profile/bin`. TBH not sure how it is related to the original `postgres` binary file.

brainrake:
Closing per the PR description ("will be closed without merging when testing is completed") — untouched since Aug 2025 and the coredump-enabled image approach was going to be redone as a nix package anyway.

PR #1758, opened 2025-08-13
feat: support multiple versions of the pgaudit extension

jfroche:
Build multiple versions of the pgaudit extension on different PostgreSQL versions.


hunleyd:
LGTM % the failing SSH connections in the CI checks

hunleyd (approved):

PR #1760, opened 2025-08-14
feat: support multiple versions of the hypopg extension 

jfroche:
Build multiple versions of the hypopg extension to on different PostgreSQL versions. Add test for the extensions and their upgrade on PostgreSQL 15 and 17.


hunleyd (approved):

jchancojr (approved):
lgtm

PR #1764, opened 2025-08-15
fix(ci): avoid testinfra failure due to loss of ssh connection

jfroche:
Tests are often failing due to the SSH access to the instance.

EC2 Instance Connect push the temporary SSH key which is then available
only for 60 seconds. Recently, errors often occur when the SSH key is
sent to the instance, resulting in a timeout.

We replace runtime SSH key injection via EC2 Instance Connect API with
cloud-init configuration to add the SSH public key during instance initialization.

Note that we are still using EC2 Instance Connect to create the SSH key
pair, but we are not using it to push the key to the instance.


hunleyd (approved):

jchancojr (approved):
lgtm

samrose (approved):
great improvement on this thank you!

PR #1765, opened 2025-08-15
feat: switch to include directives in pg_hba

staaldraad:
Using include directives makes changing the pg_hba.conf on the fly more flexible. Enabling / disabling ssl enforcement for example only requires creating or removing a file, leaving the pg_hba.conf untouched. Allowing for more repeatable and stable processes and no need for regex based replace or custom parsers.

This will also support the just-in-time access work by allowing jit to be dynamically enabled/disabled

~⚠️ do not merge yet: requires admin-api update, otherwise ssl enforcement enable/disable will stop functioning~

The required admin-api update is included as v0.88.0 added to ansible/vars.yml

## Summary by CodeRabbit

* **Improvements**
  * PostgreSQL authentication is now version-aware: v15 uses a dedicated HBA configuration, while other versions load modular HBA fragments for flexible access control.
  * New public and user-specific HBA templates add finer-grained connection rules and broaden replication access.
  * Startup and packaging flows updated so the correct HBA configurations are discovered and applied at runtime.

* **Chores**
  * PgBouncer and system service handling refined for more consistent deployment behavior.


staaldraad:
Requires pg16+

ansible/tasks/setup-postgres.yml
```diff
@@ -140,12 +140,35 @@
...
+    group: postgres
+  when: (debpkg_mode or nixpkg_mode) and (postgresql_major | int != 15)
+
+# Add pg_hba_users_public.conf
+- name: import pg_hba_users_public.conf
+  template:
+    src: files/postgresql_config/pg_hba_users_public.conf.j2
+    dest: /etc/postgresql/pg_hba_users_public.conf
+    group: postgres
+  when: (debpkg_mode or nixpkg_mode) and (postgresql_major | int != 15)
```
hunleyd:
```suggestion
- name: create pg_hba.conf with includes
  when: (debpkg_mode or nixpkg_mode) and (postgresql_major | int != 15)
  block:
  - name: import pg_hba.conf
    template:
      src: files/postgresql_config/pg_hba.conf.j2
      dest: /etc/postgresql/pg_hba.conf
      group: postgres

  # Add pg_hba_public.conf
  - name: import pg_hba_public.conf
    template:
      src: files/postgresql_config/pg_hba_public.conf.j2
      dest: /etc/postgresql/pg_hba_public.conf
      group: postgres

  # Add pg_hba_users_public.conf
  - name: import pg_hba_users_public.conf
    template:
      src: files/postgresql_config/pg_hba_users_public.conf.j2
      dest: /etc/postgresql/pg_hba_users_public.conf
      group: postgres
```

hunleyd: making this a block, you only evaluate the `when` once

hunleyd (changes requested):

hunleyd (dismissed):
lgtm, minus the conflict

samrose:
@staaldraad are we still working on this one?

staaldraad:
> @staaldraad are we still working on this one?

nope, should be ready. fixed up ansible version to rerun tests

hunleyd (approved):

brainrake:
@staaldraad Rebased onto current develop. Resolved conflicts:

- `setup-postgres.yml` — the configure pam block was restructured. Re-inserted the per-version pg_hba tasks into the current structure.
- `Dockerfile-multigres` no longer copies `pg_hba`, it now inherits from a new base image, `Dockerfile-supabase` ([#2160](https://github.com/supabase/postgres/pull/2160)), which didn't exist when this PR was opened but is now the base for the PG15/17 build matrix. It had one unconditional `pg_hba.conf.j2` copy, so I added the same PG16+ gating there.

ansible/files/postgresql_config/pg_hba.conf.j2
```diff
@@ -84,14 +14,25 @@
...
-host  all  all  ::0/0     scram-sha-256
+# if ssl is enforced, these files will exist and take precedence
+include_if_exists pg_hba_users_public_ssl.conf
+include_if_exists pg_hba_pam_public_ssl.conf
+include_if_exists pg_hba_public_ssl.conf
+
+# otherwise, non ssl enforced rules will apply
+include_if_exists pg_hba_users_public.conf
+include_if_exists pg_hba_pam_public.conf
+include_if_exists pg_hba_public.conf
```
mkindahl, line 35: Hmm... this hard-codes all configuration file names so if a downstream user (projects that use these Dockerfiles to create their own, for example Multigres), additional files cannot be added easily. There is probably also going to be problems if any files need to be added, renamed, or removed since file names are hard-coded.

A more flexible approach is to create a directory `pg_hba.conf.d` (following Debian convention for configuration directories) and just use an `include_dir` directive. In that case, files can be added and removed at will and will not generate any errors.

mkindahl (commented):
I would suggest a slightly different approach to make it easy for downstream projects.

ansible/files/postgresql_config/pg_hba.conf.j2
staaldraad: makes sense. At the time, I was concerned in ensuring the correct order, I think I did initially misunderstand the following from the docs:

> The first record with a matching connection type, client address, requested database, and user name is used to perform authentication. There is no “fall-through” or “backup”: if one record is chosen and the authentication fails, subsequent records are not considered. If no record matches, access is denied.

(originally left `pg_hba_public.conf` to always exist which was incorrect and allowed `sslmode=disable` to skip past the ssl enforced because there was a match for `host ....`)

I'll see about doing this with the `include_dir` - it needs some Supabase internals changes to understand a `pg_hba.conf.d` but not a big deal

PR #1766, opened 2025-08-15
feat: update minor versions of pg 17 and 15 

samrose:
due to critical updates in https://www.postgresql.org/about/news/postgresql-176-1610-1514-1419-1322-and-18-beta-3-released-3118/

### DIFF
--- a/nix/config.nix
+++ b/nix/config.nix
@@ -43,11 +43,11 @@ in
         postgres = {
           "15" = {
             version = "15.14";
-            hash = "";
+            hash = "sha256-Bt110wXNOHDuYrOTLmYcYkVD6vmuK6N83sCk+O3QUdI=";
           };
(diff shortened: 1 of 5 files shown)


hunleyd (approved):

steve-chavez (approved):

staaldraad (approved):

nix/config.nix
soedirgo, line 1: ~We'll need to bump the version for `17-orioledb` as well since it's based on 17.5~ all good to go

samrose: @soedirgo is correct, although we can follow up when a release for orioledb on this new version comes out. Thanks!

hunleyd (approved):

Crispy1975 (approved):

PR #1769, opened 2025-08-21
feat: deploy logrotate using system manager

yvan-sraka:


hunleyd (approved):

samrose (changes requested):
leaving change request just because I need to review these changes before they are merged

brainrake:

PR #1783, opened 2025-09-08
feat: tighten gotrue.service deps and startup behavior

cstockton:
Add stronger ordering and dependency constraints to reduce startup race conditions and noisy flapping:

- Wait for `cloud-init`, `supabase-admin-agent_salt`, `apparmor`, `systemd-sysctl`, and ufw` to complete before starting.
- Require `network-online.target` and `systemd-resolved` for stable DNS resolution; note Go's resolver can race with early boot DNS.
- Ensure `postgresql.service` is online before starting auth to avoid misleading error noise during slow boots.
- Lower `StartLimitIntervalSec` and `StartLimitBurst` to reduce repeated restarts in failure scenarios.
- Switch service type to `exec` instead of `simple`. This removes the tiny window in which systemd is supervising the wrapper process instead of the Go binary.

These changes aim to rule out capability changes, socket reuse races, and incomplete firewall/network config as causes of EADDRINUSE errors and unstable startup.

### DIFF
--- a/ansible/files/gotrue.service.j2
+++ b/ansible/files/gotrue.service.j2
@@ -1,53 +1,6 @@
 [Unit]
 Description=Gotrue
 
-# Avoid starting gotrue while cloud-init is running. It makes a lot of changes
-# and I would like to rule out side effects of it running concurrently along
-# side services.
-After=cloud-init.service
-Wants=cloud-init.target
-
-# I'm not certain waiting for this service will allow it to apply its changes
-# but I think it's prudent to do our best to let salt apply formula.
-After=supabase-admin-agent_salt.service
-
-# Given the fact that auth uses SO_REUSEADDR, I want to rule out capabilities
-# being modified between restarts early in boot. This plugs up the scenario that
-# EADDRINUSE errors originate from a previous gotrue process starting without
-# the SO_REUSEADDR flag (due to lacking capability at that point in boot proc)
-# so when the next gotrue starts it can't re-use a slow releasing socket.
-After=apparmor.service
-
-# We want sysctl's to be applied
-After=systemd-sysctl.service
-
-# UFW Is modified by cloud init, but started non-blocking, so configuration
-# could be in-flight while gotrue is starting. I want to ensure future rules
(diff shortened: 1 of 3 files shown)


ansible/files/gotrue.service.j2
```diff
@@ -24,4 +71,4 @@ EnvironmentFile=-/etc/gotrue.overrides.env
 Slice=services.slice
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=multi-user.target
```
samrose: small comment: needs newline at end of file

ansible/files/gotrue.service.j2
```diff
@@ -1,8 +1,55 @@
...
+After=network-online.target systemd-resolved.service
+
+# Auth server can't start unless postgres is online, lets remove a lot of auth
+# server noise during slow starts by requiring it.
+Wants=postgresql.service
+After=postgresql.service
+
+# Lower start limit ival and burst to prevent the noisy flapping
+StartLimitIntervalSec=10
+StartLimitBurst=5
```
samrose, line 45: @cstockton I am just curious how you go about testing these changes? Most of the time testing of these kinds of changes is happening in local infra. We're targeting having a way to test these in ci and on your machine by around end of sept 2025

samrose (approved):

ansible/files/gotrue.service.j2
cstockton: I tested these on my test project in production, I ran it through several resizes as well. Most of the variables I wanted to solve for come from our cloud init and boot sequence. This is one of those rare cases where testing on the machine really is best.

So I think a big QOL improvement for doing these sorts of tests would be lowering the feedback loop to test these things. It was very fast for me to simply ssh and vi a single unit file, but would have been more tedious to sync changes across multiple files each test resize. I'm excited to see what you come up with feel free to reach out if you wanted to chat about it.

cstockton: Fixed

hunleyd (approved):

hf (approved):

PR #1786, opened 2025-09-12
WIP feat: configure postgresql using system-manager for testing

jfroche:
We are building system manager modules that relies on postgresql. We need a module that enable testing these new modules.


ansible/tasks/setup-postgres.yml
```diff
@@ -1,338 +1,22 @@
-- name: Postgres - execute debpkg_mode tasks
```
samrose, line 1: Hmm I believe if you remove these lines, it will break our ability to deploy. Maybe I am missing something here?

samrose (changes requested):
looks like changes will impact ability to deploy? See comment

brainrake:

PR #1787, opened 2025-09-15
feat: deploy gotrue using system manager

yvan-sraka:
Depends on https://github.com/supabase/auth/pull/2388


cstockton:
Thanks for your efforts here, but we can't accept this pr as we don't intend on maintaining auth packages in this repository. For big undertakings like this my suggestion is to follow the contributing guidelines and open an issue to discuss things first.

PR #1790, opened 2025-09-15
fix: cover the old default value 1.6 (same as 1.6.4) with empty migration

samrose:
previously, the build scripts from pg_cron extension had output migrations patterned like 1.5-1.6 etc

recently we rewrote the packages to support multiple versions, and in the process did not have coverage for this form of extension version. 

This adds an empty migration from 1.6 - 1.6.4 to cover where people have the previous version signature.

### DIFF
--- a/nix/ext/tests/lib.py
+++ b/nix/ext/tests/lib.py
@@ -191,22 +191,41 @@ class PostgresExtensionTest(object):
         return reachable
 
     def check_all_upgrade_edges(self, pg_version: str, extension_dir: str):
-        """Exercise every reachable upgrade edge, canonical and legacy.
+        """Prove every real project state upgrades cleanly to latest.
 
-        For each ``<ext>--A--B.sql`` script whose source version ``A`` is
-        installable, create version A from scratch and run
-        ``ALTER EXTENSION ... UPDATE TO 'B'``, asserting the resulting version.
-        This covers symlinked alias scripts and empty version-alignment
-        migrations (e.g. the pg_cron 1.6 -> 1.6.4 no-op) that the canonical
+        Runs two passes over the upgrade-script graph discovered on disk
+        (including symlinked alias scripts and empty version-alignment
+        migrations such as the pg_cron 1.6 -> 1.6.4 no-op, which the canonical
         check_upgrade_path never reaches because they are absent from
-        versions.json.
+        versions.json):
 
-        Edges whose source version is unreachable (no base script and no
-        incoming edge, so no fresh cluster can sit at it) are skipped and
-        logged. These never lose script coverage: every such script is a
-        symlink to, or shares its body with, a reachable alias edge that does
-        run (pg_cron ``1.0--1.1`` aliases ``1.0.0--1.1.0``; ``1.4.0--1.4.1``
-        shares its body with the reachable ``1.4--1.4-1`` alias).
+        1. Per-edge body coverage. For each ``<ext>--A--B.sql`` whose source
+           ``A`` is reachable, create A from scratch and ``UPDATE TO 'B'``,
+           asserting the result. This guarantees every individual alias and
+           alignment script actually executes at least once; a shortest-path
+           upgrade could otherwise route around a broken one.
+
+        2. Full upgrade to latest from every reachable version. For every
+           version a real project could actually sit at -- canonical *or*
+           legacy-alias form -- create it from scratch and run a single
+           ``UPDATE TO '<latest>'``, asserting it lands on the latest
+           canonical version. This is the customer-faithful guarantee: any
+           existing project, whatever extversion string it reports, must reach
+           latest, with Postgres traversing alias edges as needed. If any
+           reachable version cannot reach latest, this pass fails loudly --
+           which is exactly the stranded-project bug we want to catch.
+
+        A version is "reachable" if a fresh cluster can come to rest at it: it
+        has a base install script or an incoming upgrade edge. Versions that
+        are neither (pg_cron ``1.0`` and ``1.4.0``: no base script, no incoming
+        edge) are not real project states -- no database, fresh or upgraded,
+        can ever sit at them -- so neither pass tests them. They lose no script
+        coverage either: each is a symlink to, or shares its body with, a
+        reachable alias edge that pass 1 runs (``1.0--1.1`` aliases
+        ``1.0.0--1.1.0``; ``1.4.0--1.4.1`` shares its body with the reachable
+        ``1.4--1.4-1`` alias). They exist only to keep the version graph
+        connected for legacy extversion strings, a structural property of the
+        package rather than an upgrade this test needs to exercise.
 
         The extension is restored to the latest canonical version on exit so
         the subtests that follow see the expected state.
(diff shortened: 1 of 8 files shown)


samrose:
working on testing this now

za-arthur (approved):

nix/ext/tests/lib.py
```diff
@@ -133,6 +133,115 @@ def check_upgrade_path(self, pg_version: str):
...
+            return
+
+        reachable = self._reachable_versions(bases, edges)
+        testable = [edge for edge in edges if edge[0] in reachable]
+        skipped = [edge for edge in edges if edge[0] not in reachable]
+        if skipped:
+            # Surface, never hide, what was not exercised: a source version
+            # with no install path. If a base script is ever added for one of
+            # these, this line is the cue to start covering it directly.
+            print(
```
za-arthur: I propose to fail there instead of just printing:
```suggestion
            raise AssertionError(
```

In case if there is a gap in upgrade path (for example from `pg_cron 1.6` to `pg_cron 1.6.4` the test will just print it out. It is better to fail to see this gap.

ul0gic:
@samrose — appreciate you jumping back on this one. Wanted to flag a complementary spot while you've got #1790 open, since it's adjacent to what you're doing here.

Tracing through the latest push (`0251ba0b`, *"fix: include aliases and testing to cover legacy format of ext versions"*, 2026-06-05): as I read it, #1790 adds the empty `pg_cron--1.6--1.6.4.sql` alignment migration, the symlink aliases for the non-canonical extversion strings (`1.0--1.1`, `1.4-1--1.5`, … `1.5--1.6.4`), and the `testAllUpgradeEdges = true` coverage in `nix/ext/tests/lib.py` — restoring a resolvable update path for the legacy `1.6` signature that the #1737 multi-version rework dropped (on top of the original 1.6.4 bump in #1208). So with this merged, the `1.6 → 1.6.4` path resolves at the source.

The adjacent spot that sits downstream of that — the pg_cron drop/recreate in `ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh` is gated on the extension being owned by `postgres`:

```bash
HAS_PG_CRON_OWNED_BY_POSTGRES=$(run_sql -A -t -c "select count(*) > 0 from pg_extension where extname = 'pg_cron' and extowner::regrole::text = 'postgres';")
if [ "$HAS_PG_CRON_OWNED_BY_POSTGRES" = "t" ]; then
    # drop extension pg_cron; create extension pg_cron schema pg_catalog; ...  (rebuilds at the bundled version)
fi
```

During an in-place upgrade, `postgres`-owned installs get a clean rebuild at the bundled version, while `supabase_admin`-owned installs skip the recreate and fall through to the generic `ALTER EXTENSION pg_cron UPDATE` in `run_generated_sql`. With your migration in place that generic path resolves, so this isn't load-bearing for the reported failure — but ungating the recreate to a plain existence check would make completion reconcile version **and** ownership uniformly, instead of depending on which path an install happens to take:

```bash
HAS_PG_CRON=$(run_sql -A -t -c "select count(*) > 0 from pg_extension where extname = 'pg_cron';")
if [ "$HAS_PG_CRON" = "t" ]; then
```

(The block already runs `update cron.job set username = 'postgres' where username = 'supabase_admin'`, so it's set up to normalize ownership either way.) No strong opinion on where it belongs — could fold into #1790 or I'm happy to open a separate small PR against `develop` so this stays focused on the migration. Either way, thanks for driving it.

samrose:
related customer issue https://github.com/supabase/supabase/issues/46669#issue-4594834777

samrose:
working on e2e tests do not merge yet

samrose:
## pg_cron upgrade-alias coverage: deployed fleet vs. what the test exercises

### Scope

This is strictly about the extension version-upgrade step not failing: every extversion string a live project can report must successfully run `ALTER EXTENSION pg_cron UPDATE` and land on the latest version. It does not assert anything about row-level data carried through the upgrade.

### The deployed fleet

A query of active projects by `pg_cron` extversion:

| extversion | projects | type |
|---|---|---|
| `1.6.4` | 795,780 | canonical latest |
| `1.6` | 88,844 | legacy short-form alias |
| `1.4-1` | 3,363 | legacy short-form alias |
| `1.4` | 352 | legacy short-form alias |
| `1.5` | 4 | legacy short-form alias |

Four of the five are legacy short-form strings that `versions.json` never lists (it only knows `1.3.1, 1.4.2, 1.5.2, 1.6.4`). The canonical `check_upgrade_path` test therefore never touched `1.6`, `1.4-1`, `1.4`, or `1.5`. These ~93k projects were untested before this PR.

### Why these legacy strings exist

`nix/ext/pg_cron/default.nix` renames upstream's scripts to Supabase's canonical version strings, then lays down symlink aliases so a cluster reporting an old short-form extversion still has a valid upgrade path. For example:

- `pg_cron--1.4--1.4-1.sql` -> symlink to `pg_cron--1.4.0--1.4.1.sql` (default.nix:75)
- `pg_cron--1.4-1--1.5.sql` -> symlink to `pg_cron--1.4.2--1.5.2.sql` (default.nix:76)

Postgres builds its version graph from filenames, so these aliases are real, traversable edges. They are what let a `1.4-1` project move forward.

### How the test covers every deployed version

`check_all_upgrade_edges` (lib.py:193) discovers the on-disk script graph, computes which versions a real cluster can actually sit at (`_reachable_versions`, lib.py:175), then runs two passes:

**Pass 1, per-script coverage** (lib.py:247-251): for every reachable edge `A--B`, install `A` from scratch and `UPDATE TO 'B'`. This guarantees each individual alias script body executes at least once, so a broken script cannot hide behind Postgres's shortest-path routing.

**Pass 2, full upgrade to latest** (lib.py:253-258): for every reachable version `V`, install `V` from scratch and run a single `UPDATE TO '1.6.4'`, asserting it lands on latest. This is the direct simulation of what a deployed project at `V` experiences.

Mapping the fleet to the test:

| extversion | reachable node | covered by |
|---|---|---|
| `1.6.4` | yes | already latest, no upgrade |
| `1.6` | yes | Pass 2: `1.6 -> 1.6.4` asserted |
| `1.4-1` | yes | Pass 1 (`1.4-1 -> 1.5`, `1.4-1 -> 1.5.2`) + Pass 2 (`1.4-1 -> 1.6.4`) |
| `1.4` | yes | Pass 1 (`1.4 -> 1.4-1`) + Pass 2 (`1.4 -> 1.6.4`) |
| `1.5` | yes | Pass 1 (`1.5 -> 1.6`, `1.5 -> 1.6.4`) + Pass 2 (`1.5 -> 1.6.4`) |

Every deployed version is an explicit, asserted test case.

### Why `1.4-1` (3,363 projects) is faithfully tested

`1.4-1` is a reachable node (incoming edge `1.4 -> 1.4-1`, default.nix:75), so the test can build it with `CREATE EXTENSION pg_cron VERSION '1.4-1'` and then run the exact `ALTER EXTENSION pg_cron UPDATE TO '1.6.4'` a real project would run, asserting success. The same install role is used as in production: the VM test connects as `supabase_admin`, which is the role the dashboard's pg-meta service uses to create and upgrade extensions.

### What is intentionally not tested, and why that is correct

Two edges are skipped because their source version is unreachable, meaning no cluster, fresh or upgraded, can ever sit at it: `1.0` (no base script, no incoming edge) and `1.4.0` (likewise). Neither appears in the fleet, which is consistent with them being unreachable. Their script bodies still get coverage because each is a symlink to, or shares its body with, a reachable alias that Pass 1 runs (`1.0--1.1` aliases `1.0.0--1.1.0`; `1.4.0--1.4.1` shares its body with the reachable `1.4--1.4-1`). They exist only to keep the version graph connected, which is a packaging property, not an upgrade any real project performs.

### Verification

The reachable set computed from the exact `default.nix` script graph is:

```
1.0.0  1.1.0  1.2.0  1.3.1  1.4.2  1.5.2  1.6.4   (canonical)
1.1    1.2    1.3    1.4    1.4-1  1.5    1.6      (legacy aliases)
```

All 13 non-latest reachable versions, including all four deployed legacy strings, have a verified forward path to `1.6.4`. None are stranded. CI confirms this: `pg_cron - Postgres 15 (x86_64-linux)` is green on the latest run, where the prior `AssertionError` would have failed identically.

brainrake:
Related: PSQL-1414

brainrake:
Closing in favor of #2300.

PR #1791, opened 2025-09-17
Revert "feat: tighten gotrue.service deps and startup behavior"

cstockton:
Reverts supabase/postgres#1783


cstockton:
No revert was needed here.

PR #1799, opened 2025-09-22
fix: move tmpdir for SAA to one that always exists

tomashley:
Sometimes at boot, saa is very quick to start and begins writing to the /tmp dir, that is then subseqently remounted losing any work


cynicaljoy (commented):
with a permanent disk do we need to worry about SAA temp files usage eating up the disk?

jfroche:
`After=systemd-tmpfiles-setup.service` would wait for /tmp to be mounted or
using `PrivateTmp=yes` might help for this as it adds a dependency on `systemd-tmpfiles-setup.service` and also hardens the service (https://github.com/systemd/systemd/blob/278953167d27731f46fcb56d77807d522d2ad9d2/src/core/unit.c#L1314C64-L1314C94) ?

tomashley:
> `After=systemd-tmpfiles-setup.service` would wait for /tmp to be mounted or using `PrivateTmp=yes` might help for this as it adds a dependency on `systemd-tmpfiles-setup.service` and also hardens the service (https://github.com/systemd/systemd/blob/278953167d27731f46fcb56d77807d522d2ad9d2/src/core/unit.c#L1314C64-L1314C94) ?

I've tried both of these options and neither works.
There is already `After tmp.mount` and this still causes failures when running. It seems that at some point tmp is remounted.

The issue is that saa is wrapping a salt-ssh command that is forked and doesn't respect/cannot operate within the isolation of the systemd tmpdir.

The environment variable here is purposefully being passed down to the salt-ssh process. I would prefer to keep this setting for tmpfiles passed through to salt

samrose (approved):

PR #1801, opened 2025-09-25
pg_stat_monitor with a logging backend

olirice:
Adds a low memory configuration for pg_stat_monitor + a logging backend to durably drive query insights.

this PR is to cut a staging image for experimentation. its in draft

Note that `supabase/infrastructure` specifically strip pg_stat_monitor from shared preloads due to a previous issue, so to test you need to manually ssh in to the staging instance, update postgresql.conf, and then restart postgres

### DIFF
--- a/ansible/files/postgresql_config/postgresql.conf.j2
+++ b/ansible/files/postgresql_config/postgresql.conf.j2
@@ -688,7 +688,7 @@ default_text_search_config = 'pg_catalog.english'
 #local_preload_libraries = ''
 #session_preload_libraries = ''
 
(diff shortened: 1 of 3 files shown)


samrose:
@olirice what's a good way to test this extension in your view? I can help implement if needed

olirice:
lets wait for feedback from @kemaldotearth before writing any tests. the implementation may have to change

at this point I think 
- no change in sql interface
- doesn't crash
is the best we can practically do for testing the logging part

kemaldotearth:
@olirice apologies for the slowness on this one, reports stuff took over and just saw the notification! 😓 

I actually had a question as to how I could test this locally. I managed to spin up a project on supabase.green and ssh into the shell to configure `pg_stat_monitor`, all seems to work great with the logs, however I am not entirely sure how I can consume those logs locally? 

I assume I'm missing a step or something obvious, do I need this repo locally?

olirice:
We decided not to do this and will instead wait to see what percona has planned for their next iteration

PR #1812, opened 2025-10-01
feat: run pg_regress during extension tests

jfroche:
While testing the upgrade of our PostgreSQL extensions, we also want to run the pg_regress tests that come with the extension. This ensures that the extension is functioning correctly after installation and upgrades.


samrose:
@jfroche is this tested with for instance `nix build .#checks.x86_64-linux.ext-vector -L` ?

jfroche:
> nix build .#checks.x86_64-linux.ext-vector -L

That's right

samrose (approved):
tested on x86_64-linux works great. Thank you!

PR #1816, opened 2025-10-02
Update pgsql-http to handle semver

encima:
fixes PSQL-740 as builds for 15 and 17 were failing on `develop`


samrose (approved):

nix/checks.nix
```diff
@@ -149,7 +149,7 @@
               ) filteredSqlTests;
               sortedTestList = builtins.sort (a: b: a < b) testList;
             in
-            pkgs.runCommand "postgres-${pgpkg.version}-check-harness"
+            pkgs.runCommand "postgres-${pgpkg.version}-ch"
```
samrose: checking to this change

samrose (changes requested):
one last item to check on this

nix/ext/versions.json
```diff
@@ -1,6 +1,6 @@
 {
   "http": {
-    "1.5": {
+    "1.5.0": {
       "postgresql": [
         "15"
       ],
```
samrose, line 6: We're going to update the semver for 1.6 in the versions too

samrose (changes requested):

nix/ext/versions.json
samrose: I am working on it now

samrose (approved):

PR #1820, opened 2025-10-06
refactor(postgresq): switch to 'include_dir' and then rename conf files to ensure ordering

hunleyd:
What kind of change does this PR introduce?

Moved all config files that are included in the `postgresql.conf` into the `/etc/postgresql-custom` dir then updated `postgresql.conf` to use `include_dir` instead of `include`. To ensure the same ordering of the various GUCs (to ensure proper overriding of values), rename files with digits prepended.
Finally, split auto_explain and pg_cron out into their own config files instead of them being tacked on the end of the main PG config.


tgallacher (approved):
looks sane

PR #1825, opened 2025-10-08
fix(pgmq): add missing helper function in migration script

jfroche:
pg_regress tests are failing on the pgmq extension because the helper function pgmq._extension_exists() is missing in the 1.5.0 to 1.5.1 migration script.
This change adds the missing function to the migration script.


samrose (approved):
Thanks for finding this issue!

PR #1828, opened 2025-10-08
fix(pgmq): replace drop_queue function if exists

jfroche:
We introduced a patch to override pgmq.drop_queue(TEXT) to conditionally drop objects only if they are part of the extension in `ansible/ansible/files/postgresql_extension_custom_scripts/pgmq/after-create.sql`. This script might have been installed on installations with existing 1.4.4 extensions.

When the user tries to upgrade pgmq from 1.4.4 to 1.5.1, the upgrade process will fail because the upgrade script doesn't expect `pgmq.drop_queue(TEXT)` to exist.

This change introduce the a patch to the pgmq extension to use `CREATE OR REPLACE FUNCTION` instead of `CREATE FUNCTION` for `pgmq.drop_queue(TEXT)` in the upgrade script from 1.4.5 to 1.5.0.

We also test that the install/upgrade of the extension works well with our after create script.

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.5.1.040-orioledb"
-  postgres17: "17.6.1.019"
-  postgres15: "15.14.1.019"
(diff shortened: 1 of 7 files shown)


samrose (approved):
passes all tests on relevant branches

PR #1830, opened 2025-10-09
feat(pg): use concurrent index builds

hunleyd:
## What does this PR do

Updates existing `create index` calls to `create index concurrently`.

Additionally adds `if not exists` if it was missing.

This is part of PSQL-752

### DIFF
--- a/migrations/db/init-scripts/00000000000001-auth-schema.sql
+++ b/migrations/db/init-scripts/00000000000001-auth-schema.sql
@@ -28,8 +28,8 @@ CREATE TABLE auth.users (
     updated_at timestamptz NULL,
     CONSTRAINT users_pkey PRIMARY KEY (id)
 );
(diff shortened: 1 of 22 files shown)


samrose (changes requested):
This one needs an update/careful check on pg_regress tests and output to assure it is what is intended. If you need help with that just let me know

soedirgo (changes requested):
Pls hold - these are legacy Auth/Storage migrations and they'll be thrown away as part of https://linear.app/supabase/project/path-to-pure-postgres-decouple-services-schemas-authrealtimestorage-3db51158b4ce/overview

These changes should go in [Auth](https://github.com/supabase/auth/tree/master/migrations)/[Storage](https://github.com/supabase/storage/tree/master/migrations/tenant) migrations instead

PR #1854, opened 2025-10-22
Multi version ext/pg partman

samrose:
refactor of pg_partman to include multiversion support, switching script, update of tests

PSQL-303


encima:
[pg_partman](https://github.com/pgpartman/pg_partman/releases/tag/v5.3.1) has a 5.3.1 release that could be worth moving to before merging (solves some uuidv7 things and better 17/18 compat)

encima:
@jfroche was/is there a reason to rename the background worker for the shared library? It is likely best to keep the path the same so it does not deviate from the docs

jfroche:
> @jfroche was/is there a reason to rename the background worker for the shared library? It is likely best to keep the path the same so it does not deviate from the docs

I didn't realize the project was using different names for the library and the extension. I have adapted the code for that in https://github.com/supabase/postgres/pull/1854/commits/d199683aa69c1f80eb99536763c531487258e624

encima:
Tested:
- [x] Studio shows extension
- [x] Upgrade from 15 to 17
- [x] Read Replicas

image

Crispy1975 (approved):

PR #1863, opened 2025-10-25
Enable Orioledb rewind GUCs

olirice:
Enables the GUCs that power OrioleDB fast rewind

Currently configured for 20 mins max rewind

Tested locally to work

This is to get an early prototype in the hands of interested customers. In the long term we'll get a dedicated API endpoint for it. In the short term it'll be usable by
- hitting `select orioledb_rewind_by_time(1200);` via `/v1/projects/{ref}/database/query`
- Restart the database with `/platform/projects/{ref}/restart-services`

For 2-7 second rewind times

### DIFF
--- a/Dockerfile-orioledb-17
+++ b/Dockerfile-orioledb-17
@@ -195,10 +195,11 @@ RUN echo "default_table_access_method = 'orioledb'" >> "/etc/postgresql/postgres
 
 # OrioleDB rewind configuration
 # Enables time-based rewind capability for up to 20 minutes (1200 seconds)
+# Buffer size: 1280 buffers * 8KB = 10MB for transaction retention
 RUN echo "orioledb.enable_rewind = true" >> "/etc/postgresql/postgresql.conf" && \
     echo "orioledb.rewind_max_time = 1200" >> "/etc/postgresql/postgresql.conf" && \
     echo "orioledb.rewind_max_transactions = 100000" >> "/etc/postgresql/postgresql.conf" && \
-    echo "orioledb.rewind_buffers = 13" >> "/etc/postgresql/postgresql.conf" 
+    echo "orioledb.rewind_buffers = 1280" >> "/etc/postgresql/postgresql.conf" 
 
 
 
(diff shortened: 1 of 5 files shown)


pashkinelfe:
Since the next Orioledb release we temporarily limited rewind to IS_DEV=1 builds only. This is for the prod users (hosted instances) couldn't shut down their clusters accidentally while trying this feature. So for trying this feature (for self-hosted users) I propose Orioledb should be built with IS_DEV=1. But not add IS_DEV=1 for the AMI for hosted users. @samrose

olirice:
@pashkinelfe we're moving to make this generally available on the platform so AI builders can start testing/integrating against the behavior prior to the betra

The image is in alpha with free-tier only users who have consented to the risks, so i think we're okay to enable it on the oriole ami

pashkinelfe (commented):
I see the previous discussion could lead to the bad decision of building "prod" orioledb instance for everyone with IS_DEV=1.

If we decide as @olirice  proposed that it's safe for hosted users being allowed to shut down cluster by calling rewind sql function, I'll remove rewind safeguard in Orioledb repo, so that rewind would be allowed on prod builds. (And we'll continue nix building for prod users without setting IS_DEV = 1.)

@akorotkov what do you think?

pashkinelfe:
If I misunderstood something and this PR won't be used for hosted OrioleDB AMI (except "special" AMI for AI builders), please LMK @olirice

olirice:
The intent is for it to be the official OrioleDB image on the platform. Right now I'm just cutting one for staging to make sure everything works. we can coordinate in slack for if/when its safe to make it the default config

pashkinelfe:
@olirice then let's do this PR without IS_DEV=1. It will work with the current release of Orioledb. And let's discuss how that should be in the next (upcoming) release later in slack.

olirice:
I moved it back into draft until we get through those changes. Currently just want to make sure i can get it passing E2E on staging

pashkinelfe (approved):

PR #1873, opened 2025-10-27
feat: support multiple versions of the pg_plan_filter extension

jfroche:
Build multiple versions of the pg_plan_filter extension on different PostgreSQL versions.

Add test for the extension and their upgrade on PostgreSQL 15 and 17.


samrose (approved):
tested on all relevant branches setting up for release

PR #1878, opened 2025-10-29
feat(ansible): add pgBackRest tasks and configurations

jchancojr:

Feature- pgBackRest backup solution

pgBackRest tasks and configs are absent

## What is the new behavior?

pgBackRest tasks and configs are missing are present

## Additional context
NA

## Summary by CodeRabbit

* **New Features**
  * Adds pgBackRest support with configurable global and repository settings (S3 storage, retention, encryption, compression, performance and logging). Some async/archive options remain as commented placeholders.

* **Chores / Install**
  * Adds automated install and setup: package install, user/group and sudoers provisioning, directory creation, config placement and a safe wrapper for invoking pgBackRest.
  * Updates packaged PostgreSQL release references.


ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,75 @@
...
+- name: Symlink pgBackRest binary
+  file:
+    path: /usr/bin/pgbackrest
+    src: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest
+    state: link
+
+- name: Sticky bit the pgBackRest binary
+  file:
+    path: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest
+    mode: '4755'
```
jfroche: Setting setuid won't work for multiple reasons:

- `/var/lib/pgbackrest/.nix-profile/bin/pgbackrest` is a symlink to a binary in the store, and the kernel ignores setuid bits on symlinks.
- We cannot modify the binary file in the Nix store as it is mounted with `ro`
- Binaries with setuid in the Nix store won't work as it is mounted with `nosuid`

I see two possible paths:

- Use sudo/doas
- Create a wrapper in /usr/local/bin (or /usr/bin) that execs the original pgbackrest in the store and set the setuid bit on that wrapper

I find the sudo approach safer and more maintainable than using a setuid wrapper.

samrose: Something like this could work

```
- name: Configure sudoers for pgBackRest
  lineinfile:
    path: /etc/sudoers.d/pgbackrest
    line: "postgres ALL=(pgbackrest) NOPASSWD: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest"
    create: yes
    mode: '0440'
    validate: 'visudo -cf %s'

- name: Create pgBackRest wrapper script
  copy:
    content: |
      #!/bin/bash
      exec sudo -u pgbackrest /var/lib/pgbackrest/.nix-profile/bin/pgbackrest "$@"
    dest: /usr/bin/pgbackrest
    mode: '0755'
    owner: root
    group: root
```
if not root/root could be some user/group such that pgbackrest cannot change permissions of it's own script

hunleyd: added

samrose: @staaldraad might want to take a peek at this just to make sure you're happy with this kind of approach too when you have a moment

staaldraad: there are a few arguments that look a little scary

* `--cmd`
* `--ssh-cmd`
* `--repo-host-cmd`

The `--config` variants may also allow access to the above arguments. 

The wrapper would feel safer, although less flexible, if we had fixed arguments per command in the wrapper, and then allow sudo on the wrapper.

Something like:

```bash
#!/bin/bash

if [[ "$1" == "archive-get" ]]; then
  exec /var/lib/pgbackrest/.nix-profile/bin/pgbackrest archive-get --stanza="$2"
else
  echo "Error: not allowed"
  exit 1
fi
```

```
- name: Configure sudoers for pgBackRest
  lineinfile:
    path: /etc/sudoers.d/pgbackrest
    line: "postgres ALL=(pgbackrest) NOPASSWD:  /usr/bin/pgbackrest"
    create: yes
    mode: '0440'
    validate: 'visudo -cf %s'
```

hunleyd: the backrest binary can only be called if you're already on the box, or by changing the `archive_command` in the postgresql.conf, which we disallow already. so it kinda feels like by the time someone could exploit those args, we'd already have bigger problems, no? @staaldraad

staaldraad: yes, in the ideal world. And that is what https://github.com/supabase/postgres/pull/1879 brings us closer to. Up until now, there was still avenue for `superuser` --> `COPY (select 1) FROM PROGRAM '.../pgbackrest'` (or first getting a more powerful shell and then having "local" access). 

Fully agree that we are protecting against "bigger problems", however it is good to think of layered defense. As we start narrowing how big a problem local access is (if the likeliest route is to get shell as the unprivileged `postgres` user). With things like https://github.com/supabase/postgres/pull/1846/files , we really squeeze down what is possible to do, even if someone manages to change `archive_command` or get shell via postgres. Definitely trying to think in chains of small bugs here, in isolation one bug or protection might not be much, but chained together small bugs can be powerful. And the same goes for small protections

jchancojr: Yep. This makes sense. Thanks very much will get this going

ansible/files/pgbackrest_config/pgbackrest.conf
```diff
@@ -0,0 +1,17 @@
...
+compress-type = zst
+delta = y
+expire-auto = n
+link-all = y
+log-level-console = info
+log-level-file = detail
+log-subprocess = y
+resume = n
+start-fast = y
+[supabase]
```
hunleyd, line 15:
```suggestion
start-fast = y

[supabase]
```

ansible/playbook.yml
```diff
@@ -40,6 +40,10 @@
       import_tasks: tasks/setup-wal-g.yml
       when: debpkg_mode or nixpkg_mode or stage2_nix
 
+    - name: Install pgBackRest
+      import_tasks: tasks/setup-pgbackrest.yml
+      when: debpkg_mode or nixpkg_mode or stage2_nix
```
hunleyd:
```suggestion
    - name: Install pgBackRest
      import_tasks: 
        file: tasks/setup-pgbackrest.yml
      when:
        - debpkg_mode or nixpkg_mode or stage2_nix
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
+- name: Create pgBackRest group
+  ansible.builtin.group:
+    name: pgbackrest
+    state: present
+  when: nixpkg_mode
```
hunleyd:
```suggestion
  when:
    - nixpkg_mode
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
...
+- name: Create pgBackRest user
+  ansible.builtin.user:
+    comment: pgBackRest user
+    group: pgbackrest
+    groups: pgbackrest, postgres
+    home: /var/lib/pgbackrest
+    name: pgbackrest
+    shell: /sbin/nologin
+    system: true
+  when: nixpkg_mode
```
hunleyd:
```suggestion
  when:
    - nixpkg_mode
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Allow pgBackRest to run commands
+  copy:
```
hunleyd:
```suggestion
  ansible.builtin.copy:
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
...
+    content: |
+      pgbackrest ALL=(ALL) NOPASSWD: /usr/bin/pgbackrest
+
+- name: Install pgBackRest
+  become: true
+  become_user: pgbackrest
+  ansible.builtin.shell: |
+    sudo -u pgbackrest bash -c ". /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix profile install github:supabase/postgres/{{ git_commit_sha }}#pg-backrest"
+  changed_when: true
+  when: stage2_nix
```
hunleyd:
```suggestion
  ansible.builtin.shell: |
    sudo -u pgbackrest bash -c ". /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix profile install github:supabase/postgres/{{ git_commit_sha }}#pg-backrest"
  become: true
  become_user: pgbackrest
  changed_when: true
  when:
    - stage2_nix
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
...
+    path: "{{ backrest_dir }}"
+    state: directory
+  loop:
+    - /etc/pgbackrest/conf.d
+    - /var/lib/pgbackrest
+    - /var/spool/pgbackrest
+    - /var/log/pgbackrest
+  loop_control:
+    loop_var: backrest_dir
+  when: nixpkg_mode
```
hunleyd:
```suggestion
  when:
    - nixpkg_mode
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,93 @@
...
+    src: "files/pgbackrest_config/{{ conf_item['name'] }}"
+  loop:
+    - {name: computed_globals.conf, path: conf.d}
+    - {name: pgbackrest.conf, path: ''}
+    - {name: repo1_async.conf, path: conf.d}
+    - {name: repo1_encrypted.conf, path: conf.d}
+    - {name: repo1.conf, path: conf.d}
+  loop_control:
+    loop_var: conf_item
+  when: stage2_nix
```
hunleyd:
```suggestion
  when:
    - stage2_nix
```

hunleyd (changes requested):

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,94 @@
...
+    create: yes
+    line: "{{ item }}"
+    mode: '0440'
+    path: '/etc/sudoers.d/pgbackrest'
+    validate: 'visudo -cf %s'
+  loop:
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
```
hunleyd, line 31:
```suggestion
    - 'postgres ALL=(pgbackrest) NOPASSWD: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest'
    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/pgbackrest'
    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'```

**Review by hunleyd — CHANGES_REQUESTED** [review]

File: ansible/tasks/setup-pgbackrest.yml [hunk 11]
```diff
@@ -0,0 +1,96 @@
...
+    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
+    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'```
```
hunleyd:
```suggestion
    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
```

hunleyd (changes requested):

ansible/tasks/setup-pgbackrest.yml
hunleyd: an attempt at addressing your concern is now in this PR @staaldraad . Please let us know what you think?

jchancojr:
Tested local infra with workflow ami:
✅ pause
✅ resume
✅ upgrades

hunleyd (approved):

ansible/files/pgbackrest_config/repo1.conf
```diff
@@ -0,0 +1,14 @@
+[supabase]
+repo1-block = y
+repo1-bundle = y
+# repo1-path = <foo>
+repo1-retention-diff = 1
+repo1-retention-full = 28
+repo1-retention-full-type = time
+repo1-retention-history = 0
+# repo1-s3-bucket= <foo>
+# repo1-s3-endpoint= <foo>
```
hunleyd:
```suggestion
# repo1-s3-bucket=  <foo>
# repo1-s3-endpoint = <foo>
```

hunleyd (changes requested):

hunleyd (approved):

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,96 @@
...
+    mode: '0440'
+    path: '/etc/sudoers.d/pgbackrest'
+    validate: 'visudo -cf %s'
+  loop:
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
+    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
+    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
```
hunleyd:
```suggestion
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,96 @@
...
+    force: true
+    path: /etc/pgbackrest/pgbackrest.conf
+    src: /etc/pgbackrest.conf
+    state: link
+
+- name: Move pgBackRest files to /etc/pgbackrest
+  ansible.legacy.copy:
+    group: postgres
+    dest: "/etc/pgbackrest/{{ conf_item['path'] }}/{{ conf_item['name'] }}"
+    mode: '0644'
```
hunleyd:
```suggestion
    mode: '0640'
```

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,96 @@
...
+  when:
+    - stage2_nix
+
+- name: Create pgBackRest wrapper script
+  ansible.builtin.copy:
+    content: |
+      #!/bin/bash
+      _raw_args="$@"
+      _sanitized_args=$(echo $_raw_args | sed -e 's/--cmd=[^ ]*//g; s/--repo-host-cmd=[^ ]*//g; s/--config=[^ ]*//g' )
+      exec sudo -u pgbackrest /var/lib/pgbackrest/.nix-profile/bin/pgbackrest "$_sanitized_args"
```
hunleyd:
```suggestion
      # Sanitize dangerous arguments
      sanitized_args=()
      while [[ $# -gt 0 ]]; do
        case "$1" in
          --cmd=*|--cmd)
            # Skip --cmd and its value
            [[ "$1" == "--cmd" ]] && shift
            shift || true
            ;;
          --ssh-cmd=*|--ssh-cmd)
            # Skip --ssh-cmd and its value
            [[ "$1" == "--ssh-cmd" ]] && shift
            shift || true
            ;;
          --repo-host-cmd=*|--repo-host-cmd)
            # Skip --repo-host-cmd and its value
            [[ "$1" == "--repo-host-cmd" ]] && shift
            shift || true
            ;;
          --config=*|--config)
            # Skip --config and its value
            [[ "$1" == "--config" ]] && shift
            shift || true
            ;;
          *)
            sanitized_args+=("$1")
            shift
            ;;
        esac
      done
      exec sudo -u pgbackrest /var/lib/pgbackrest/.nix-profile/bin/pgbackrest "${sanitized_args[@]}"
```

hunleyd (changes requested):

hunleyd (approved):

ansible/tasks/setup-pgbackrest.yml
```diff
@@ -0,0 +1,122 @@
...
+    create: yes
+    line: "{{ item }}"
+    mode: '0440'
+    path: '/etc/sudoers.d/pgbackrest'
+    validate: 'visudo -cf %s'
+  loop:
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /var/lib/pgbackrest/.nix-profile/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/pgbackrest'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
+    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
```
hunleyd, line 31:
```suggestion
    - 'postgres ALL=(pgbackrest) NOPASSWD: /usr/bin/nix'
    - 'pgbackrest ALL=(pgbackrest) NOPASSWD: /usr/bin/bash'
```

hunleyd (changes requested):

hunleyd (approved):

jchancojr:
test-postgres-engines-with-smoke.sh 

```
Test Summary

ENGINE          RESULT
------          ------
15              PASSED (375s)
17              PASSED (352s)
17-oriole       PASSED (367s)

[INFO] Postgres commit: add0a0f372300387e01d5652e45d077143861a83
```

hunleyd (approved):

PR #1888, opened 2025-11-03
tests: start fail2ban before healthcheck in testinfra

samrose:
fail2ban needed starting in test before healthcheck of the same service


samrose:
latest run showing https://github.com/supabase/postgres/actions/runs/19037413106/job/54366947732?pr=1888#step:16:135 fail2ban tested successfully. note this does not explicitly test configurations of fail2ban, but a different test that happens at end of AMI build does catch misconfiguration of fail2ban for postgres service.

soedirgo (approved):

PR #1891, opened 2025-11-04
feat(setup-postgres.yml): enable checksums via initdb for postgres

jchancojr:

Enables checksums for PostgreSQL

Checksums are disabled

## What is the new behavior?

Checksums are enabled for _new_ projects

## Additional context
Provides earlier detection by verifying data page integrity as pages are _read_ from disk
https://www.postgresql.org/docs/current/checksums.html


ansible/tasks/setup-postgres.yml
```diff
@@ -212,7 +212,7 @@
   become: true
   become_user: 'postgres'
   ansible.builtin.command:
-    cmd: /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data initdb -o "--allow-group-access" -o "--username=supabase_admin"
+    cmd: /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data initdb -o "--allow-group-access" -o "--username=supabase_admin" -o "--data-checksums"
```
hunleyd:
```suggestion
    cmd: /usr/lib/postgresql/bin/pg_ctl -D /var/lib/postgresql/data initdb -o "--allow-group-access" -o "--data-checksums" -o "--username=supabase_admin"
```

hunleyd (changes requested):

hunleyd (approved):

PR #1892, opened 2025-11-05
fix: grant execute on pg_reload_conf() to postgres

za-arthur:

Grants `postgres` EXECUTE privileges on the `pg_reload_conf()` function.

`postgres` cannot execute `pg_reload_conf()`.

## What is the new behavior?

`postgres` can execute `pg_reload_conf()`.

### DIFF
--- a/migrations/db/migrations/20251105172723_grant_pg_reload_conf_to_postgres.sql
+++ b/migrations/db/migrations/20251105172723_grant_pg_reload_conf_to_postgres.sql
@@ -1,0 +1,5 @@
+-- migrate:up
+grant execute on function pg_catalog.pg_reload_conf() to postgres;
+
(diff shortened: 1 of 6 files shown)


migrations/db/migrations/20251105172723_grant_pg_reload_conf_to_postgres.sql
```diff
@@ -0,0 +1,5 @@
+-- migrate:up
+grant execute on function pg_catalog.pg_reload_conf() to postgres;
```
soedirgo:
```suggestion
grant execute on function pg_catalog.pg_reload_conf() to postgres with grant option;
```

soedirgo (approved):

PR #1898, opened 2025-11-06
fix: needs to run on merge not pre merge

samrose:
We need the behavior and timing of this action retained to post merge similar to https://github.com/supabase/postgres/blob/develop/.github/workflows/ami-release-nix.yml


laurenceisla (approved):

jfroche (approved):
:+1:

darora (approved):

PR #1906, opened 2025-11-11
refactor: DRY extensions NixOS test

jfroche:
Extract common NixOS test node configuration and functions into lib.nix to eliminate duplication across extension tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stopped automatic preloading of plpgsql and plpgsql_check libraries by default.

* **Chores**
  * Centralized and simplified PostgreSQL extension test infrastructure into a shared test library, reducing duplication and standardizing test node setup across many extension tests.
  * Removed several now-unnecessary in-file test networking/port-forwarding configurations.

### DIFF
--- a/nix/ext/pg-safeupdate.nix
+++ b/nix/ext/pg-safeupdate.nix
@@ -32,6 +32,8 @@ let
         # Install versioned library
         install -Dm755 ${pname}${postgresql.dlSuffix} $out/lib/${pname}-${version}${postgresql.dlSuffix}
 
+        touch $out/share/postgresql/extension/${pname}--${version}.control
+        touch $out/share/postgresql/extension/${pname}--${version}.sql
         runHook postInstall
       '';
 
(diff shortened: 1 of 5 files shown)


hunleyd:
there are three instances of `initdb` herein, the first with all the options and the 2nd two with no options. and there's two copies of pg_upgrade. would it make sense to consolidate these into lib.nix as well so there's one definition and every use becomes the same @jfroche ?

jfroche:
> there are three instances of `initdb` herein, the first with all the options and the 2nd two with no options. and there's two copies of pg_upgrade. would it make sense to consolidate these into lib.nix as well so there's one definition and every use becomes the same @jfroche ?

`initdb` is called as part of the migration script. That migration script can be different from one extension to another, that's why you can see multiple `initdb` (and the same for `pg_upgrade`).

hunleyd:
> > there are three instances of `initdb` herein, the first with all the options and the 2nd two with no options. and there's two copies of pg_upgrade. would it make sense to consolidate these into lib.nix as well so there's one definition and every use becomes the same @jfroche ?
> 
> `initdb` is called as part of the migration script. That migration script can be different from one extension to another, that's why you can see multiple `initdb` (and the same for `pg_upgrade`).

thanks for explaining; makes sense! will review after conflicts are fixed.

hunleyd (approved):
looks sane to me

samrose (approved):

brainrake:
Superseded — the pg-safeupdate fix has since been superseded by later changes to nix/ext/pg-safeupdate.nix on develop, and the rest of the diff has drifted from the PR's stated purpose (DRY NixOS tests). Closing; the nix-eval-jobs memory/worker tweak may still be worth a fresh, small PR if still relevant.

PR #1917, opened 2025-11-13
fix: adaptive retry on aws

samrose:
Changes Made

  Added environment variables to both "Build AMI stage 1" and "Build AMI stage 2" steps in
  .github/workflows/testinfra-ami-build.yml:111,120:

  env:
    AWS_MAX_ATTEMPTS: 10
    AWS_RETRY_MODE: adaptive

  What This Does

  1. AWS_MAX_ATTEMPTS: 10 - The AWS SDK will retry failed API calls up to 10 times before giving up (default is 3)
  2. AWS_RETRY_MODE: adaptive - Uses intelligent retry logic that:
    - Backs off exponentially on throttling errors
    - Adapts retry timing based on AWS response patterns
    - Handles transient network issues (like your "unexpected EOF")
    - Works with rate limiting and temporary connectivity problems

  Expected Impact

  - Fixes the frequent "unexpected EOF" errors during snapshot tagging by automatically retrying when the AWS API connection drops
  - No workflow re-runs needed for transient failures - the build will self-heal
  - Minimal time overhead - Only adds delay when retries are actually needed
  - Works for both stages - Protects both the surrogate build and the stage 2 build

The frequent "unexpected EOF" was tied many times to a specific interaction with aws that doesn't require a full re-run of workflow

### DIFF
--- a/.github/workflows/testinfra-ami-build.yml
+++ b/.github/workflows/testinfra-ami-build.yml
@@ -109,12 +109,18 @@ jobs:
           echo "" >> common-nix.vars.pkr.hcl
 
       - name: Build AMI stage 1
+        env:
+          AWS_MAX_ATTEMPTS: 10
+          AWS_RETRY_MODE: adaptive
         run: |
           GIT_SHA=${{github.sha}}
           nix run github:supabase/postgres/${GIT_SHA}#packer -- init amazon-arm64-nix.pkr.hcl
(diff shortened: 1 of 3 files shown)


jchancojr (approved):
LGTM

hunleyd:
i cannot wait for this to be merged. the unexpected eof is _killing_ me

PR #1919, opened 2025-11-13
feat: draft to block copy ... program

samrose:
Removed execute permissions that enabled COPY...PROGRAM:
  - /usr/bin/cat rix
  - /usr/bin/dash rix
  - /usr/bin/mknod rix

  Added explicit deny rules:
  # Explicitly deny shell execution to block COPY ... PROGRAM
  deny /bin/sh x,
  deny /bin/bash x,
  deny /bin/dash x,
  deny /usr/bin/sh x,
  deny /usr/bin/bash x,
  deny /usr/bin/dash x,
  deny /usr/bin/cat x,
  deny /usr/bin/mknod x,

  Improved documentation for preserved operations (admin-mgr, wal-g)

  File 2: Permission Tests (ansible/files/permission_check.py)

  +93 lines

  Added check_copy_program_blocked() function that:
  - Tests COPY TO PROGRAM (write via cat)
  - Tests COPY FROM PROGRAM (read via echo)
  - Verifies both fail with permission errors
  - Confirms no files created
  - Integrated into main test suite

  Result

  OS-level blocking of COPY ... PROGRAM using AppArmor while preserving WAL-G and admin-mgr functionality, with automated testing in the AMI build process.

### DIFF
--- a/ansible/files/permission_check.py
+++ b/ansible/files/permission_check.py
@@ -244,6 +244,96 @@ def check_postgresql_mount():
     print("postgresql.service mounts /etc as read-only.")
 
 
+def check_copy_program_blocked():
+    """Check that COPY ... PROGRAM is blocked by AppArmor."""
+    import psycopg2
+    from psycopg2 import sql
+
+    errors = []
+
+    try:
+        # Connect to PostgreSQL as the postgres superuser
+        conn = psycopg2.connect(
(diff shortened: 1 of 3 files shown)


brainrake:
@samrose is this still planned? The AppArmor deny rules never landed on develop and the draft has been quiet since last November. Security-relevant, so I'd rather confirm before closing. Should we move it to a project/issue?

PR #1921, opened 2025-11-14
feat: support multiple versions of the supautils extension

jfroche:
## Summary by CodeRabbit

* **New Features**
  * Multi-version packaging for the supautils extension (per-version builds, latest-version symlink) with PostgreSQL 15 and 17 support; version metadata added and exposed.

* **Tests**
  * Integration test covering install, migration (15→17) and end-to-end verification.
  * SQL regression tests validating extension behavior on PostgreSQL 15 and 17.

* **Chores**
  * Exposed additional passthru metadata (versions, counts, installed extensions, withPackages helper, default settings).

### DIFF
--- a/nix/ext/supautils.nix
+++ b/nix/ext/supautils.nix
@@ -1,35 +1,96 @@
 {
+  pkgs,
   lib,
   stdenv,
   fetchFromGitHub,
   postgresql,
 }:
-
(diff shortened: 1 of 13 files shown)


samrose:
we only deploy latest, so this is unneeded after all

PR #1925, opened 2025-11-17
feat: reduction of closure size for plv8

samrose:
Reduced the closure size (total dependency footprint) of the plv8 PostgreSQL extension by optimizing its runtime dependencies.

  What was changed:

  1. nix/ext/plv8/default.nix - Replaced stdenv.cc.cc.lib with libcxx in all rpath settings
    - Added libcxx to the function inputs
    - Updated 4 locations where runtime library paths are set (both Darwin/macOS and Linux paths)
    - This reduces closure size because stdenv.cc.cc.lib pulls in the entire compiler toolchain, while libcxx only brings in the C++ standard library

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.5.1.073-orioledb"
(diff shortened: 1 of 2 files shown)


hunleyd (approved):

jchancojr (approved):
lgtm ty so much

PR #1745, opened 2025-08-11
ci: Custom GitHub Runners for Nix Builds

yvan-sraka:
This PR implements a migration from GitHub's standard runners to a hybrid infrastructure combining self-hosted and ephemeral Blacksmith runners for building Nix packages.
The implementation includes runner selection, dynamic build matrix generation, and optimized caching strategies to improve build performance and cost efficiency.

## Problem Statement

The previous CI implementation had several limitations:

1. **Monolithic build process**: A single job attempted to build all packages across all architectures
2. **Inefficient resource allocation**: All packages used the same runner type regardless of build complexity
3. **Limited parallelization**: Builds couldn't be efficiently distributed across different runner types
4. **Redundant builds**: No mechanism to skip packages already available in the binary cache
5. **Poor cost optimization**: Large, expensive builds ran on the same infrastructure as small, quick builds
6. **Poor job output clarity**: No separation of build results made it hard to identify issues

### High-Level Design

```
┌─────────────────┐
│   nix-eval      │  Evaluates flake, generates build matrix
│   (Blacksmith)  │  Identifies cached vs. uncached packages
└────────┬────────┘  Identifies large packages
         │
         ├──────────────┬──────────────┬
         │              │              │              
         v              v              v              
┌────────────────┐ ┌────────────────┐ ┌────────────────┐
│ aarch64-linux  │ │ aarch64-darwin │ │ x86_64-linux   │
│ Self-hosted/   │ │ Self-hosted    │ │ Blacksmith     │
│ Blacksmith     │ │ (macOS)        │ │ Ephemeral      │
└────────────────┘ └────────────────┘ └────────────────┘
```

### Architecture Components

1. **Nix Evaluation Phase** (`nix-eval.yml`):
   - Runs on powerful ephemeral runner (32vcpu)
   - Evaluates all flake outputs using `nix-eval-jobs`
   - Checks cache status for each package
   - Generates optimized build matrices per architecture

2. **Build Phases** (separate jobs per architecture):
   - **aarch64-linux**: Self-hosted or Blacksmith ARM runners
   - **aarch64-darwin**: Self-hosted macOS runners
   - **x86_64-linux**: Blacksmith ephemeral runners

3. **Runner Selection Logic**:
   - KVM-required packages → Self-hosted runners with KVM support
   - Large packages (Rust, PostGIS) → 32vcpu runners
   - Standard packages → 8vcpu runners
   - Darwin packages → Self-hosted macOS runners

[...]

PR #1927, opened 2025-11-18
feat(postgres): add PG 15.15

hunleyd:

add PG 15.15 to our AMI

no 15.15

## What is the new behavior?

has 15.15

## Additional context

https://www.postgresql.org/about/news/postgresql-181-177-1611-1515-1420-and-1323-released-3171/

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ This is the same PostgreSQL build that powers [Supabase](https://supabase.io), b
 
 
 ## Primary Features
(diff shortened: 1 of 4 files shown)


samrose (changes requested):
we'll test this first and then will approve thanks @hunleyd

yvan-sraka:
What's making the check here fails is building a lot of `pgrx`-based extensions versions in parallel, that make the runner crash. As demonstrated [here](https://github.com/supabase/postgres/pull/1932), without most of them the PR pass. This should be fixed by https://github.com/supabase/postgres/pull/1745

PR #1928, opened 2025-11-18
feat: iterate on ami, packer, ansible, ssh to investigate

samrose:
Tools for iterating on packer, ansible and ami creation without going through pull request (requires aws credentials, creates ec2 instances, etc)


hunleyd (approved):
Not strong in Go, but it makes sense and looks sane.

PR #1930, opened 2025-11-19
fix: disable temporary Nix checks on aarch64-linux

yvan-sraka:
The purpose is to unblock jobs like https://github.com/supabase/postgres/actions/runs/19473229950/job/55813935552?pr=1927, this change should be reverted by https://github.com/supabase/postgres/pull/1745.

### DIFF
--- a/nix/checks.nix
+++ b/nix/checks.nix
@@ -22,399 +22,402 @@
     in
     {
       checks =
-        let
-          # Create a testing harness for a PostgreSQL package. This is used for
-          # 'nix flake check', and works with any PostgreSQL package you hand it.
-          # deadnix: skip
-          makeCheckHarness =
-            pgpkg:
-            let
-              pg_prove = pkgs.perlPackages.TAPParserSourceHandlerpgTAP;
-              inherit (self'.packages) pg_regress;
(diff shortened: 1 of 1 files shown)


nix/checks.nix
```diff
@@ -22,399 +22,402 @@
...
-                  chmod 1777 "$KEY_DIR"
-
-                  if [[ ! -f "$KEY_FILE" ]]; then
-                    if ! (dd if=/dev/urandom bs=32 count=1 2>/dev/null | od -A n -t x1 | tr -d ' \n' > "$KEY_FILE"); then
-                      if ! (openssl rand -hex 32 > "$KEY_FILE"); then
-                        echo "00000000000000000000000000000000" > "$KEY_FILE"
-                        echo "Warning: Using fallback key" >&2
+        if system == "aarch64-linux" then
+          { }
+        else
```
yvan-sraka, line 27: Yeah, it's a bit of a hack ... the idea is indeed to revert it!

samrose: @yvan-sraka totally understood thanks for giving us a temporary way to do it!

samrose (approved):

PR #1936, opened 2025-11-20
fix: try to resize build-time volume and copy nix store when done + test size before copy

samrose:
trying to support this flow basically

```
mount /dev/xvdc /mnt/nix-temp

# Copy existing Nix installation to temp volume first
rsync -aHAXS /nix/ /mnt/nix-temp/

# Now bind mount over the original
mount --bind /mnt/nix-temp /nix

# Do all nix operations (they'll use the temp volume)
nix profile install ...

# Stop daemon before unmounting
systemctl stop nix-daemon || true

# Check size, unmount, and copy back
umount /nix
# /nix now shows the original pre-bind-mount content

# Clear it and copy from temp
rm -rf /nix/*
rsync -aHAXS /mnt/nix-temp/ /nix/

systemctl start nix-daemon || true
```

### DIFF
--- a/amazon-arm64-nix.pkr.hcl
+++ b/amazon-arm64-nix.pkr.hcl
@@ -149,7 +149,7 @@ source "amazon-ebssurrogate" "source" {
   launch_block_device_mappings {
     device_name           = "/dev/${var.build-vol}"
     delete_on_termination = true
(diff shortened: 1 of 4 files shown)


samrose:
closing in favor of #1938

PR #1950, opened 2025-11-26
Feat: add track_functions to supautils allowed calls

encima:

Feature

Hard to track and optimise functions when using Postgrest and RPC calls

## What is the new behavior?

Allow users to enable `track_functions` using supautils

## Additional context

Implements PSQL-831


nix/tests/sql/enable_tracking.sql
```diff
@@ -0,0 +1,2 @@
+alter role postgres set track_functions = 'pl';
+alter role postgres set track_io_timing = on;
```
soedirgo: Let's reset it in the same test so it doesn't affect other tests
```suggestion
alter role postgres set track_functions = 'pl';
alter role postgres set track_io_timing = on;

alter role postgres reset track_functions;
alter role postgres reset track_io_timing;
```

soedirgo (approved):

PR #1966, opened 2025-12-06
feat: nix docker images

samrose:


brainrake:
Good stuff, would like to revive at some point.

PR #1974, opened 2025-12-09
(WIP) feat: migrate pgrx extensions to crane build system

yvan-sraka:
This change adds [`crane`](https://crane.dev/) support to the existing `pgrx` extension builder, enabling better incremental build performance and caching for `pg_jsonschema` and other pgrx extensions. Crane separates dependency builds from main crate builds, allowing dependencies to be cached independently.

The unified `buildPgrxExtension.nix` now accepts an optional `craneLib` parameter. When provided, it uses crane's two-phase build (`buildDepsOnly` + `buildPackage`). Otherwise, it falls back to `rustPlatform.buildRustPackage`. Both paths share the same build and install logic, avoiding code duplication.

Crane requires `Cargo.lock` at the source root while `rustPlatform` accepts external `lockFile` paths. The builder handles this by ensuring external lock files are properly accessible during the build process. Crane's `cargoVendorDir` is only used for crane builds to avoid conflicts with `rustPlatform`'s cargo handling.

Extensions opt into crane builds by passing `useCrane = true` to `mkPgrxExtension`. All existing build parameters remain compatible with both backends.


samrose:
this one seems pretty important and would like to get to this one soon after nixpkgs update

yvan-sraka:
`pg_graphql` (with `useCrane = true`) hits a `tokio-postgres` compilation error:

```
error: builder for '/nix/store/80sbdnbf97j3wjiy38vwi0w9y8qj41l2-pg_graphql-deps-deps-1.2.0.drv' failed with exit code 101;
       last 25 log lines:
       >     Checking tracing-error v0.2.1
       >    Compiling semver v0.11.0
       >     Checking tokio-util v0.7.4
       >     Checking pgx-sql-entity-graph v0.7.1
       >     Checking futures-executor v0.3.26
       >     Checking tokio-postgres v0.7.8
       >     Checking futures v0.3.26
       >    Compiling rustc_version v0.3.3
       >    Compiling atomic-traits v0.3.0
       >     Checking clap v4.1.6
       > error[E0277]: `DataRowBody` doesn't implement `std::fmt::Debug`
       >    --> /nix/store/v7hfg1lha0s198amava5m9gfqqpj9vk1-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/tokio-postgres-0.7.8/src/row.rs:202:5
       >     |
       > 199 | #[derive(Debug)]
       >     |          ----- in this derive macro expansion
       > ...
       > 202 |     body: DataRowBody,
       >     |     ^^^^^^^^^^^^^^^^^ `DataRowBody` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
       >     |
       >     = help: the trait `std::fmt::Debug` is not implemented for `DataRowBody`
       >     = note: this error originates in the derive macro `Debug` (in Nightly builds, run with -Z macro-backtrace for more info)
       >
       > For more information about this error, try `rustc --explain E0277`.
       > error: could not compile `tokio-postgres` (lib) due to 1 previous error
       > warning: build failed, waiting for other jobs to finish...
       For full logs, run 'nix log /nix/store/80sbdnbf97j3wjiy38vwi0w9y8qj41l2-pg_graphql-deps-deps-1.2.0.drv'.
error: 1 dependencies of derivation '/nix/store/f8gq25rp5z9g14k5kbgsil4k4clhbdv5-pg_graphql-1.2.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/07fiwlxlzk4vpa5b57b6kzyj9nw696ma-pg_graphql.drv' failed to build
```

Basic crane integration works correctly (`pg_jsonschema` builds successfully) but crane uses the external `Cargo.lock` files directly while `rustPlatform` was applying patches that contained e.g. fixed dependency versions. Here, the external `Cargo.lock` files have `tokio-postgres 0.7.8` which has this debug trait issue.

To fix this, we need to either update the external `Cargo-X.X.X.lock` files with compatible dependency versions or apply the same patches that `rustPlatform` was using.

yvan-sraka:
After further consideration, I'm not convinced that integrating `crane` is a good idea _(regardless of the number of hours needed to actually successfully build all `pgrx` extensions with it, currently only `jsonschema` works)_.

`crane` (like e.g. [`naersk`](https://github.com/nix-community/naersk)) adds some overhead: **it creates 2 derivations rather than one** for each extension _(one for source code and one for cargo dependencies),_ which is a great feature when iterating on a rust codebase in a developer environment _(since sources change more often than dependencies)_ **but less useful here since** it's about building most of the time packages from scratch and **different versions of the same package almost never share the same `Cargo.toml`**, so there's no deduplication of work done here by the increased granularity of `crane` derivations that just add more nix overhead to building rust projects _(compared to current `rustPlatform` approach)_.

I tried to measure it, to back up my thoughts with some data, but without a lot of success so far _(benchmarking nix builds is notoriously a hard problem to get right)_.

samrose:
closed in favor #2001

PR #1980, opened 2025-12-15
chore: bump version and postgresql paths

staaldraad:
Adds an additional path to the systemd config


samrose:
@staaldraad do we still need this one?

staaldraad:
back from the past 😲 going to close, AA covers us, there is defence in depth here but we can plan for that as part of wider efforts.

PR #1996, opened 2026-01-06
Fix for Issue #41529: plpgsql_check Recursive Function Error

Adityakk9031:
Issue
plpgsql_check causes "cannot find parent statement on pldbgapi2 call stack" errors when running recursive PL/pgSQL functions. This happens because the extension's internal pldbgapi2 component (used for profiling/tracing) has a known limitation with tracking call stacks in recursive contexts. Since plpgsql_check was configured in shared_preload_libraries, it was automatically loaded for all sessions, affecting production workloads.

Ref: Issue [#41529](https://github.com/supabase/supabase/issues/41529)

Solution
Removed plpgsql_check from shared_preload_libraries in the Nix configuration.

The extension is no longer loaded automatically at startup.
It remains available for on-demand use via CREATE EXTENSION plpgsql_check; for development and linting purposes.
This prevents the recursion error in standard operations while preserving the tool's availability for developers.

## Summary by CodeRabbit

* **Changes**
  * The plpgsql_check extension is no longer automatically preloaded. It must now be manually loaded via CREATE EXTENSION when needed.

### DIFF
--- a/nix/ext/plpgsql-check.nix
+++ b/nix/ext/plpgsql-check.nix
@@ -131,13 +131,10 @@ buildEnv {
 
   passthru = {
     inherit versions numberOfVersions switch-ext-version;
-    hasBackgroundWorker = true;
-    defaultSettings = {
-      shared_preload_libraries = [
-        "plpgsql"
-        "plpgsql_check"
-      ];
-    };
+    # plpgsql_check is a development/linting tool that should be loaded on-demand
+    # via CREATE EXTENSION, not automatically preloaded. Automatic preloading causes
+    # "cannot find parent statement on pldbgapi2 call stack" errors in recursive
+    # PL/pgSQL functions (see issue #41529)
     version =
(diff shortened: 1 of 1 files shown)


samrose:
Hi thanks for your contribution!

In the case of this setting, if it were to be changed, it would need to be set/changed in the postgres configuration that is applied when the service deploys. The changes that were applied here are just for integration testing scope, and do not impact the deployed service.

 I have created [an issue in the upstream project with potential fixes](https://github.com/okbob/plpgsql_check/issues/197), and will work with that team to see if there is something that can be done. Supabase platform depends on plpgsql_check being in shared_preload_libraries since around 2021, for what it's worth. So we can't make a change on that in a simple way currently. But we can try to work with upstream project to see if there is a way to fix this reported issue while still retaining current supabase configuration. Going to close this PR in favor of that direction. Thanks again!

PR #2001, opened 2026-01-12
feat: enable sccache for cargo-pgrx builds in GitHub Actions

jfroche:
Speed up rust builds by enabling sccache caching for cargo-pgrx builds in GitHub Actions. We mount a persistent disk at /var/cache/sccache and configure sccache to use it.

### DIFF
--- a/.github/actions/nix-install-ephemeral/action.yml
+++ b/.github/actions/nix-install-ephemeral/action.yml
@@ -53,4 +53,7 @@ runs:
           trusted-public-keys = nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
           ${{ inputs.push-to-cache == 'true' && 'post-build-hook = /etc/nix/upload-to-cache.sh' || '' }}
           ${{ inputs.enable-sccache-sandbox-path == 'true' && 'extra-sandbox-paths = /nix/var/cache/sccache?' || '' }}
-          max-jobs = 4
+          max-jobs = 1
+          auto-allocate-uids = true
+          use-cgroups = true
+          experimental-features = nix-command flakes cgroups auto-allocate-uids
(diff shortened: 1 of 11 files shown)


samrose (changes requested):
I am worried about the following on this PR


Out of the version transitions where something changed:                                                                             
                                                                                                                                      
  - 0.4.2 → 0.4.3: Rust + pgrx changed — sccache does not help                                                                        
  - 0.4.4 → 0.4.5: Rust + pgrx changed — sccache does not help
  - 0.4.5 → 0.4.6: Rust only changed — sccache does not help                                                                          
  - 0.5.0 → 0.5.3: Rust + pgrx changed — sccache does not help
  - 0.5.4 → 0.5.5: pgrx only changed — sccache partially helps
  - 0.5.6 → 0.5.7: Rust + pgrx changed — sccache does not help

  5 out of 6 version transitions changed the Rust compiler, making sccache useless. Only one transition (0.5.4 → 0.5.5) was a
  pgrx-only bump where sccache would have partially helped.

  So sccache is solving a problem that almost never occurs in practice for this project. The max-jobs = 1 constraint, sandbox
  complexity, stickydisk management, and UID mapping gymnastics are a lot of machinery for something that would have helped in 1 out
  of 6 historical version bumps — and only partially at that.

Am I missing something? Thanks!

jfroche:
Our goal was to improve both builds and rebuilds. Rebuild of the extensions is the most common use case, and we have seen improvements in rebuilds while working on this PR (see cache hits in the CI logs: https://github.com/supabase/postgres/actions/runs/22390434451/job/64811231796).

As you mentioned, adding a new version of an extension might cause more cache misses if they changed rust/pgrx but we have seen other extensions updates without rust/pgrx changes that benefited from the cache.
Agree that we introduce some complexity but we still believe they are worth it for the improvements we get in rebuilds. We could limit further the downside of this `max-jobs = 1` constraint only on the build of the rust extensions.

samrose:
@jfroche I see so the area of struggle where sccache helps is mostly iterating on/rebuild when working on an extension package, which is a common occurance during PR, or local development.

PR #2005, opened 2026-01-15
tests: run postgis test suite on package itself

samrose:
Running the full test suite for postgis in our package


samrose:
closed in favor of #2004

PR #2006, opened 2026-01-19
feat: reduce attack surface due to packages

samrose:
Removing packages that are not needed, with priority on items that could be security attack vector


brainrake:
Seems worthwile, may need reviving!

PR #2009, opened 2026-01-20
feat: Small core version of supabase postgres

filipecabaco:

Builds a smaller version of supabase postgres with a limited subset of extensions. It also creates a snapshot of the system on boot so we can have faster boots when using this image.

Screenshot 2026-01-20 at 16 53 14


samrose:
Thanks @filipecabaco setting up a new branch and PR inspired by this work. Will close this one in favor of the new branch/PR

PR #2017, opened 2026-01-26
fix: enable stickydisk again with new nix-installer-action

jfroche:
Make sure the eval succeeds with both an empty /nix cache and a existing /nix cache.
Use the new nix installer.

## Summary by CodeRabbit

* **Chores**
  * Switched to a new ephemeral Nix installer, standardized installer inputs, added max-jobs, and moved post-build hook into installer config.
  * Replaced long install steps with the ephemeral installer in CI, added cache mount and pre-install socket/ownership prep.
  * Updated installer version across images and scripts.

* **Bug Fixes**
  * Fixed installer invocation/heredoc handling to ensure configuration is applied.

* **Documentation**
  * Updated getting-started guide and version references to the new Nix release.

### DIFF
--- a/.github/actions/nix-install-ephemeral/action.yml
+++ b/.github/actions/nix-install-ephemeral/action.yml
@@ -40,11 +40,10 @@ runs:
         sudo chmod +x /etc/nix/upload-to-cache.sh
       env:
         NIX_SIGN_SECRET_KEY: ${{ env.NIX_SIGN_SECRET_KEY }}
-    - name: Install nix
-      uses: cachix/install-nix-action@4e002c8ec80594ecd40e759629461e26c8abed15 # v31.9.0
+    - uses: NixOS/nix-installer-action@d6ef7ecd8f685af89869e5aca0580a33e3e3150c
       with:
-        install_url: https://releases.nixos.org/nix/nix-2.32.2/install
-        extra_nix_config: |
+        installer-version: 2.33.1
+        extra-conf: |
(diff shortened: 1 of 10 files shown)


.github/actions/nix-install-ephemeral/action.yml
```diff
@@ -40,11 +40,10 @@ runs:
         sudo chmod +x /etc/nix/upload-to-cache.sh
       env:
         NIX_SIGN_SECRET_KEY: ${{ env.NIX_SIGN_SECRET_KEY }}
-    - name: Install nix
-      uses: cachix/install-nix-action@4e002c8ec80594ecd40e759629461e26c8abed15 # v31.9.0
+    - uses: NixOS/nix-installer-action@d6ef7ecd8f685af89869e5aca0580a33e3e3150c
       with:
-        install_url: https://releases.nixos.org/nix/nix-2.32.2/install
-        extra_nix_config: |
+        installer-version: 2.33.1
```
samrose, line 45: will the newer version cause any conflict with using slightly older version elsewhere?

samrose (approved):

.github/actions/nix-install-ephemeral/action.yml
jfroche: No there are no breaking change in 2.33 but I could also upgrade the other references because 2.33 comes with nice performance improvements. It should also fix the problem we had with IPC leak on macos with postgres (see https://github.com/NixOS/nix/pull/14459), we might want to tell people to upgrade nix on their macos laptop to avoid this.

PR #2024, opened 2026-01-28
chore: write version file pam_jit_pg

staaldraad:
Write a version file for pam_jit_pg so that salt doesn't try overwrite when applying to an image that already has the library installed. Salt knows to check the version and only update if salt specifies a different version should be used.

## Summary by CodeRabbit

* **Chores**
  * Create a version file recording the deployed gatekeeper/pam_jit_pg version during PostgreSQL setup in two installation paths.
  * Ensure the version file is owned by root for consistency and traceability.


ansible/tasks/stage2-setup-postgres.yml
```diff
@@ -176,6 +176,13 @@
               shell: |
                 sudo ln -s /var/lib/postgresql/.nix-profile/lib/security/pam_jit_pg.so $(find /nix/store -type d -path "/nix/store/*-linux-pam-*/lib/security" -print -quit)/pam_jit_pg.so
 
+            - name: Write version file
+              ansible.builtin.copy:
+                dest: "/root/pam_jit_pg-version"
+                group: "root"
+                owner: "root"
+                content: "1.0.0"
```
soedirgo: Nit: can this be interpolated somehow from the version in gatekeeper.nix?

staaldraad: most probably and it should be, good callout. I was being lazy 😅

samrose (approved):

PR #2027, opened 2026-01-30
fix: load safeupdate but disable for all but Data API

### LINKED ISSUE #1308: Cannot use safeupdate for the database nor the postgres user (session_preload_libraries)
# Bug report

Basically, this worked in older projects. In newer projects, the permissions are stricter and one cannot enable this, safeguarding, `safeupdate` extension. But this is a crucial help for avoiding human failure.

```sql
-- 1. doesn't work
GRANT SET ON PARAMETER session_preload_libraries TO postgres;
ALTER ROLE postgres SET session_preload_libraries = 'safeupdate';

-- 2. also doesn't work
alter database postgres set session_preload_libraries = 'safeupdate';


-- test
CREATE TABLE IF NOT EXISTS todos (id SERIAL NOT NULL PRIMARY KEY, title TEXT);
INSERT INTO todos (title) VALUES ('test-todo-123');

DELETE FROM todos; -- should fail with safeupdate
```

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

It's all said.

## To Reproduce

Try either of the provided options in the intro.

## Expected behavior

`safeupdate` can be enabled in the database (I don't think we necessarily need `postgres` role setting but rather one config that enables this in the database or not).

## Screenshots

image

encima:

Bug fix

pg-safeupdate is only usable by the `authenticator` role

## What is the new behavior?

pg-safeupdate is enabled by default for the `authenticated` role - i.e. for the Data API
Can be enabled for the `postgres` user by running `SET safeupdate.enabled=1;`

## Additional context

Fixes https://github.com/supabase/postgres/issues/1308#issuecomment-2622879937
PSQL-910

## Summary by CodeRabbit

* **Chores**
  * Added a database migration to install and configure a PostgreSQL safety extension (installed and configured but disabled by default).

* **Tests**
  * Added test coverage to validate role setup, privilege elevation during maintenance, full-table cleanup, and schema removal behaviors.


encima:
Failing on 15 because the `authenticated` role does not exist...checking why

steve-chavez (approved):

migrations/db/migrations/20260130074514_load_disable_pg_safeupdate.sql
```diff
@@ -0,0 +1,8 @@
+-- migrate:up
+ALTER ROLE authenticated SET session_preload_libraries = 'safeupdate';
+load 'safeupdate';
```
steve-chavez: I think we should also have it for `anon`.

steve-chavez: Maybe even `service_role` (debatable)

encima: `anon` definitely makes sense, though `service_role` is more of a breaking change I guess? 

Can definitely add in `anon` to this PR if we want though?

steve-chavez: Yes, `anon` should be a no-brainer. Let's leave `service_role` for now.

steve-chavez:
@encima I think we should also announce this in https://github.com/orgs/supabase/discussions/categories/changelog, it's a "breaking change" in some cases I assume.

samrose:
@encima @steve-chavez if we do that we need to make it approx 3 week ahead of time change announcement per our breaking change process thanks for catching that @steve-chavez

migrations/db/migrations/20260130074514_load_disable_pg_safeupdate.sql
```diff
@@ -0,0 +1,9 @@
+-- migrate:up
+ALTER ROLE authenticated SET session_preload_libraries = 'safeupdate';
+ALTER ROLE anon SET session_preload_libraries = 'safeupdate';
```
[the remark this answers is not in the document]
encima: This is expected and the migration linked is from 2022; tests for roles show that safeupdate is only expected to be loaded for `anon` and `authenticator`

encima:
Closing in favour of #2105

PR #2028, opened 2026-01-30
feat: bump oriole to latest

samrose:
bumping to latest using our git commit sha capability, to get the latest features and fixes available now.

Adding a test

The test starts a Linux aarch64-linux VM with orioledb and the full Supabase initialization stack. Once that's ready, it switches to the OrioleDB 17 specialisation with rewind enabled (1280 main
  buffers, 1200s max time, 100K max transactions).
                                                                                                                                                                                             
  Subtest 1 verifies the switch worked. It confirms the orioledb extension is installed, queries pg_settings to print all rewind-related config, asserts that orioledb.enable_rewind is on,
  and prints the initial rewind queue and evicted lengths.                                                                                                                                   
                                                                                          
  Subtest 2 does a basic rewind matching the bash script. It creates a table with 100 rows, captures the xid, oxid, and md5 hash in one query, then inserts 40 more rows to dirty the state.
  It calls orioledb_rewind_to_transaction(xid, oxid) which crashes postgres. Systemd Restart=always brings it back. After restart it verifies the row count is back to 100 and the hash
  matches the pre-rewind snapshot.

  Subtest 3 does a rewind under buffer pressure. It creates a new table and inserts 5000 rows across 50 separate transactions, captures a checkpoint, then inserts 1000 more rows in 10
  batches. It prints the rewind queue and evicted lengths to show the buffer state, then rewinds to the checkpoint. After restart it verifies the row count matches the checkpoint count
  (5000).


nix/tests/expected/z_orioledb-17_rewind.out
```diff
@@ -0,0 +1,76 @@
...
+ orioledb_get_complete_xid          | integer     | 
+ orioledb_get_current_oxid          | bigint      | 
+ orioledb_get_rewind_evicted_length | bigint      | 
+ orioledb_get_rewind_queue_length   | bigint      | 
+ orioledb_rewind_by_time            | void        | rewind_time integer
+ orioledb_rewind_to_timestamp       | void        | rewind_timestamp timestamp with time zone
+ orioledb_rewind_to_transaction     | void        | xid integer, oxid bigint
+(8 rows)
+
+-- Test diagnostic functions (these are read-only and safe)
```
pgnickb: This isn't exactly accurate, `orioledb_get_current_oxid` allocates an `oxid` so it is similar to `pg_current_xact_id`.

nix/tests/sql/z_orioledb-17_rewind.sql
```diff
@@ -0,0 +1,34 @@
...
+    JOIN pg_catalog.pg_namespace n ON n.oid = p.pronamespace
+WHERE
+    p.proname LIKE 'orioledb_rewind%'
+    OR p.proname LIKE 'orioledb_get_complete%'
+    OR p.proname LIKE 'orioledb_get_current_oxid%'
+    OR p.proname LIKE 'orioledb_get_rewind%'
+ORDER BY
+    p.proname;
+
+-- Test diagnostic functions (these are read-only and safe)
```
pgnickb: Same concern regarding `orioledb_get_current_oxid()` as above.

samrose: I think we actually don't want a test here in regress at all for rewind. So will be changing this to TAP test soon

samrose:
next steps: 

- build image to test
- test rewind in local infra
- remove regress test for rewind

nix/tests/expected/z_orioledb-17_rewind.out
samrose: this test was removed, working on a different testing approach thank you

nix/tests/sql/z_orioledb-17_rewind.sql
samrose: this test was removed, working on a different testing approach thank you

samrose:
This PR has been tested in supadev for smoke tests, etc. it includes the oriole specific vm tests above as well, which will run any time a change is made to oriole inputs, and can be run locally

hunleyd (approved):

PR #2033, opened 2026-02-02
feat: add portable CLI binaries for multi-platform Supabase CLI distribution

LGUG2Z:

Adds a new, portable `psql_17_cli_portable` target.

All current builds are not portable (i.e. they have hard dependencies on the `/nix/store`), and they all include the full list of extensions.

## What is the new behavior?

The `psql_17_cli_portable` target can run without a `/nix/store` present, and only includes the `supautils` extension.

## Additional context

```
This commit implements portable, self-contained PostgreSQL binaries for the
Supabase CLI across macOS (ARM), Linux (x64), and Linux (ARM64), along with
automated CI/CD workflows for building and releasing these artifacts.

The Supabase CLI needs to ship PostgreSQL binaries that work on user machines
without requiring Nix or other system dependencies. This means extracting the
actual binaries from Nix's wrapper scripts, bundling all necessary shared
libraries, and patching them to use relative paths instead of hardcoded Nix
store paths.

A `variant` parameter was added to the postgres build pipeline to distinguish
between "full" (all extensions) and "cli" (minimal extensions for Supabase CLI).
The `cliExtensions` list contains 6 extensions required for running Supabase
migrations: supautils, pg_graphql, pgsodium, supabase_vault, pg_net, and pg_cron.
Built-in extensions (uuid-ossp, pgcrypto, pg_stat_statements) are included
automatically with PostgreSQL. `makeOurPostgresPkgs`/`makePostgresBin` were
modified to accept this parameter. A new `psql_17_cli` package is created using
`variant = "cli"`, while the full extension set is preserved for base packages
(`psql_15`, `psql_17`, `psql_orioledb-17`).

The portable CLI variant (`psql_17_cli_portable`) includes 6 extensions for
migration support while maintaining a significantly smaller size than the full
build. The implementation in `nix/packages/postgres-portable.nix` extracts
binaries from `psql_17_cli` using a `resolve_binary()` function that follows
wrapper layers to find the actual ELF/Mach-O binaries behind Nix's environment
setup scripts.

All Nix-provided libraries (ICU, readline, zlib, etc.) are bundled while
excluding system libraries (`libc`, `libpthread`, `libm`, `glibc`, `libdl`) that
must come from the host. This distinction is critical: Linux bundles must
exclude glibc due to kernel ABI dependencies, while macOS can include more libs
due to its different linking model. Dependency resolution runs multiple passes
to catch transitive deps (e.g., ICU → charset → etc.).

Platform-specific patching is applied: Linux binaries use the system interpreter
(`/lib64/ld-linux-*.so.2`) and `$ORIGIN`-based RPATHs, while macOS binaries use
`@rpath` with `@executable_path`. Wrapper scripts set `LD_LIBRARY_PATH` (Linux)
or `DYLD_LIBRARY_PATH` (macOS) to find bundled libraries. The bundle includes
PostgreSQL config templates (`postgresql.conf`, `pg_hba.conf`, `pg_ident.conf`)
tailored for CLI usage with minimal local dev settings, plus the complete
Supabase migration script (`migrate.sh`) with all init-scripts and migration
SQL files (55 files, 236KB).

A Docker-style initialization script (`supabase-postgres-init.sh`) provides
one-command database setup via environment variables (`POSTGRES_USER`,
`POSTGRES_PASSWORD`, `POSTGRES_DB`, `PGDATA`).

To achieve true portability, a `portable` parameter was added to the PostgreSQL
build in `nix/postgresql/generic.nix`. When `portable = true`, three critical
hardcoded paths are excluded from the build:

1. `--with-system-tzdata` is removed from configure flags, allowing PostgreSQL
to use bundled timezone data from the `share/` directory instead of a hardcoded
`/nix/store/.../tzdata` path;

2. the `locale-binary-path.patch` is excluded, so PostgreSQL calls `locale -a`
from system PATH rather than using an absolute path to glibc's locale command;

3. the `postFixup` initdb wrapper is disabled to avoid hardcoding glibc paths.
The `portable` parameter defaults to `false` in `nix/postgresql/default.nix` but
is overridden to `true` for the CLI variant in `nix/packages/postgres.nix` using
`.override { portable = true; }`.

This ensures standard PostgreSQL builds remain unchanged while the CLI variant
produces truly portable binaries without any `/nix/store` references.

A GitHub Actions workflow builds portable binaries across all three platforms
using a matrix strategy. Each build runs automated portability checks that
verify no `/nix/store` references remain, validate RPATH configuration, confirm
transitive dependencies are bundled, ensure system libraries are NOT bundled,
and check wrapper scripts contain proper library path setup. Post-build testing
validates binaries work without Nix (`postgres --version`, `psql --version`). On
tagged releases (`v*-cli`), the workflow creates GitHub releases with tarball
artifacts and checksums.

The test infrastructure needed significant changes to support variants with
different extension sets. An `isCliVariant` parameter was added to
`makeCheckHarness`, and the hardcoded `shared_preload_libraries` list in
`postgresql.conf.in` was replaced with a `@PRELOAD_LIBRARIES@` placeholder. A
`generatePreloadLibs` script now parses `receipt.json` at test time and
dynamically builds the preload list based on available extensions, removing the
previous timescaledb removal hack for OrioleDB.
```

## Summary by CodeRabbit

* **New Features**
  * Portable, multi-architecture PostgreSQL CLI bundles and a CLI runtime variant; lightweight local Postgres config/auth templates, CLI init SQL, and a pgsodium key-generation helper.
* **Tests**
  * CLI-specific init/test flow, dynamic preload library handling, and an extensive portability test harness validating binaries, libraries, tzdata, and extension/migration behavior.
* **Infrastructure**
  * CI workflow to build per-architecture bundles, run portability checks, package tarballs with checksums, and publish GitHub Releases.

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -1,0 +1,446 @@
+name: CLI Release
+
+on:
+  push:
+    tags:
+      - "v*-cli"
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version tag (e.g., v1.0.0-cli)"
+        required: true
+        type: string
+
+jobs:
+  build-native:
+    name: Build ${{ matrix.arch }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - system: aarch64-darwin
+            runner: macos-14
+            arch: darwin-arm64
+          - system: x86_64-linux
+            runner: ubuntu-latest
+            arch: linux-x64
+          - system: aarch64-linux
+            runner: ubuntu-24.04-arm
+            arch: linux-arm64
+    steps:
+      - name: Checkout repository
+        uses: supabase/postgres/.github/actions/shared-checkout@HEAD
+
+      - name: Install Nix
+        uses: ./.github/actions/nix-install-ephemeral
+
+      - name: Run portability check
+        run: |
+          nix build .#checks.${{ matrix.system }}.psql_17_cli_portable --system ${{ matrix.system }} -L --accept-flake-config
+
+      - name: Build portable CLI bundle
+        run: |
+          nix build .#psql_17_cli_portable --system ${{ matrix.system }} -L --accept-flake-config
+
+      - name: Determine version
+        id: version
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            VERSION="${{ github.event.inputs.version }}"
+          else
+            VERSION="${{ github.ref_name }}"
+          fi
+          # Sanitize version by replacing slashes with dashes
+          VERSION="${VERSION//\//-}"
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+
+      - name: Prepare build directory
+        run: |
+          VERSION="${{ steps.version.outputs.version }}"
+          ARCH="${{ matrix.arch }}"
+          BUILD_DIR="supabase-postgres-${VERSION}-${ARCH}"
+
+          mkdir -p "${BUILD_DIR}"
+          shopt -s dotglob
+          cp -rL result/* "${BUILD_DIR}/"
+          shopt -u dotglob
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: supabase-postgres-${{ matrix.arch }}
+          path: supabase-postgres-${{ steps.version.outputs.version }}-${{ matrix.arch }}
+          retention-days: 90
+          include-hidden-files: true
+
+  test-portability:
+    name: ${{ matrix.arch }} without Nix
+    needs: [build-native]
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: macos-14
+            arch: darwin-arm64
+          - runner: ubuntu-latest
+            arch: linux-x64
+          - runner: ubuntu-24.04-arm
+            arch: linux-arm64
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: supabase-postgres-${{ matrix.arch }}
+          path: .
+
+      - name: Test binaries work without Nix
+        run: |
+          cd bin
+          find . -maxdepth 1 -type f -exec chmod +x {} +
+          ls -la
+
+          echo "Testing postgres --version..."
+          ./postgres --version
+
+          echo "Testing pg_config --version..."
+          ./pg_config --version
+
+          echo "Testing psql --version..."
+          ./psql --version
+
+          echo "Testing initdb --version..."
+          ./initdb --version
+
+      - name: Test initialization script
+        run: |
+          set -e
+
+          echo "=========================================="
+          echo "Testing supabase-postgres-init.sh"
+          echo "=========================================="
+
+          # Make init script executable
+          chmod +x ./share/supabase-cli/bin/supabase-postgres-init.sh
+
+          # Test 1: Initialize database with init script
+          export PGDATA="$(pwd)/pgdata-init-test"
+          export POSTGRES_USER="test_user"
+          export POSTGRES_PASSWORD="test_password"
+          export POSTGRES_DB="test_db"
+
+          echo "Starting PostgreSQL with init script..."
+          ./share/supabase-cli/bin/supabase-postgres-init.sh -p 54321 > init.log 2>&1 &
+          INIT_PID=$!
+
+          # Wait for PostgreSQL to be ready
+          for i in {1..30}; do
+            if ./bin/pg_isready -p 54321 -h 127.0.0.1 -U test_user > /dev/null 2>&1; then
+              echo "PostgreSQL is ready"
+              break
+            fi
+            if [ $i -eq 30 ]; then
+              echo "PostgreSQL failed to start"
+              cat init.log
+              exit 1
+            fi
+            sleep 1
+          done
+
+          echo ""
+          echo "=========================================="
+          echo "Testing password authentication"
+          echo "=========================================="
+
+          # Test password authentication (should work with scram-sha-256)
+          PGPASSWORD="test_password" ./bin/psql -p 54321 -h 127.0.0.1 -U test_user -d test_db -c "SELECT version();" > /dev/null
+          if [ $? -eq 0 ]; then
+            echo "✓ Password authentication successful"
+          else
+            echo "✗ Password authentication failed"
+            exit 1
+          fi
+
+          # Test wrong password (should fail) - temporarily disable set -e
+          set +e
+          PGPASSWORD="wrong_password" ./bin/psql -p 54321 -h 127.0.0.1 -U test_user -d test_db -c "SELECT 1;" > /dev/null 2>&1
+          WRONG_PW_STATUS=$?
+          set -e
+          if [ $WRONG_PW_STATUS -ne 0 ]; then
+            echo "✓ Wrong password correctly rejected"
+          else
+            echo "✗ Wrong password was accepted (security issue!)"
+            exit 1
+          fi
+
+          echo ""
+          echo "=========================================="
+          echo "Testing database initialization"
+          echo "=========================================="
+
+          # Verify database exists
+          PGPASSWORD="test_password" ./bin/psql -p 54321 -h 127.0.0.1 -U test_user -d test_db -v ON_ERROR_STOP=1 -c "SELECT current_database();" | grep "test_db"
+          if [ $? -eq 0 ]; then
+            echo "✓ Custom database created"
+          else
+            echo "✗ Custom database not found"
+            exit 1
+          fi
+
+          # Verify configuration files were set up
+          if [ -f "$PGDATA/postgresql.conf" ] && [ -f "$PGDATA/pg_hba.conf" ]; then
+            echo "✓ Configuration files created"
+          else
+            echo "✗ Configuration files missing"
+            exit 1
+          fi
+
+          # Verify pgsodium/vault getkey scripts are configured
+          if grep -q "pgsodium.getkey_script" "$PGDATA/postgresql.conf" && grep -q "vault.getkey_script" "$PGDATA/postgresql.conf"; then
+            echo "✓ Getkey scripts configured"
+          else
+            echo "✗ Getkey scripts not configured"
+            exit 1
+          fi
+
+          echo ""
+          echo "=========================================="
+          echo "Testing idempotency"
+          echo "=========================================="
+
+          # Stop PostgreSQL
+          ./bin/pg_ctl -D "$PGDATA" stop -m fast
+          wait $INIT_PID 2>/dev/null || true
+
+          # Run init script again (should skip initialization)
+          ./share/supabase-cli/bin/supabase-postgres-init.sh -p 54321 > init2.log 2>&1 &
+          INIT_PID2=$!
+
+          # Wait for PostgreSQL to be ready
+          for i in {1..30}; do
+            if ./bin/pg_isready -p 54321 -h 127.0.0.1 -U test_user > /dev/null 2>&1; then
+              echo "PostgreSQL is ready after re-init"
+              break
+            fi
+            if [ $i -eq 30 ]; then
+              echo "PostgreSQL failed to start on second run"
+              cat init2.log
+              exit 1
+            fi
+            sleep 1
+          done
+
+          # Verify we can still connect with same password
+          PGPASSWORD="test_password" ./bin/psql -p 54321 -h 127.0.0.1 -U test_user -d test_db -c "SELECT 1;" > /dev/null
+          if [ $? -eq 0 ]; then
+            echo "✓ Idempotency verified - database not re-initialized"
+          else
+            echo "✗ Failed to connect after re-running init script"
+            exit 1
+          fi
+
+          # Check log to verify it skipped initialization
+          if grep -q "already initialized" init2.log; then
+            echo "✓ Init script correctly detected existing database"
+          else
+            echo "⚠ Warning: Init script may have re-initialized (check init2.log)"
+          fi
+
+          # Stop PostgreSQL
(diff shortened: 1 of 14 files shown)


samrose (approved):

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,154 @@
...
+          ./postgres --version
+
+          echo "Testing pg_config --version..."
+          ./pg_config --version
+
+          echo "Testing psql --version..."
+          ./psql --version
+
+          echo "Testing initdb --version..."
+          ./initdb --version
```
sweatybridge, line 115: Could you add a test for running our standard supabase migrations on the build artifacts? https://github.com/supabase/postgres/blob/develop/migrations/db/migrate.sh

I had problems running it so just want to be sure. Please also feel free to adjust the migrate.sh file if you need to. We have tests on both docker and ami to make sure it won't break existing builds.

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,243 @@
...
+
+          # Check migration results (allow pgbouncer migration to fail as it's not in CLI variant)
+          if [ $MIGRATION_STATUS -ne 0 ]; then
+            if grep -q "pgbouncer" migration.log; then
+              echo "Migration failed on expected pgbouncer error (not in CLI variant) - continuing"
+            else
+              echo "Migrations failed with unexpected errors"
+              exit 1
+            fi
+          fi
```
sweatybridge, line 324: I believe the migrate.sh [stops](https://github.com/supabase/postgres/actions/runs/21657509181/job/62435425756#step:4:388) when encoutering pgbouncer error. So the database is partially migrated if ignore here.

The minimal fix is to apply a dummy pgbouncer schema https://github.com/supabase/postgres/blob/develop/ansible/files/pgbouncer_config/pgbouncer_auth_schema.sql

But open to other suggestions too.

sweatybridge (approved):

jgoux:
Here are the issues I had integrating the artefact into the CLI:

# PostgreSQL Bundle Issues

This document describes the issues encountered with the current PostgreSQL bundle (`supabase-postgres-<version>-<os>-<arch>.zip`) and the workarounds implemented in the CLI.

**Goal:** The bundle should be extractable and runnable without any CLI-side workarounds.

---

## 1. Code Signing on macOS (Critical)

**Status:** Blocking - requires CLI workaround

**Issue:** After extracting the zip, all binaries crash immediately with:
```
SIGKILL (Code Signature Invalid)
termination: {"flags":0,"code":2,"namespace":"CODESIGNING","indicator":"Invalid Page"}
```

**Root Cause:** The binaries have invalid or missing code signatures after extraction. macOS's security system (Gatekeeper) kills unsigned/invalid binaries.

**Current Workaround:** Re-sign all executables and `.dylib` files with ad-hoc signature after extraction:
```go
// binary.go:434-463 - codesignPostgresDir()
cmd := exec.Command("codesign", "-f", "-s", "-", path)
```

**Ideal Fix:** Sign binaries during packaging, or use proper Mach-O flags that don't require signing.

---

## 2. Timezone Data - Hardcoded Nix Paths (Critical)

**Status:** Blocking PostgREST - no effective workaround

**Issue:** The postgres binary was built with:
```
--with-system-tzdata=/nix/store/fy3qa8s8kzb7a6abzmyidzp1c8axz3s3-tzdata-2025b/share/zoneinfo
```

This causes `pg_timezone_names` to fail:
```
ERROR: could not open directory "/nix/store/fy3qa8s8kzb7a6abzmyidzp1c8axz3s3-tzdata-2025b/share/zoneinfo": No such file or directory
```

**Impact:**
- PostgREST cannot load schema cache (calls `pg_timezone_names` during introspection)
- Any query involving timezone enumeration fails

**Attempted Workarounds:**
1. Setting `TZDIR=/var/db/timezone/zoneinfo` environment variable - doesn't work for `pg_timezone_names`
2. Creating symlink at `/nix/store/...` - requires root privileges

**Ideal Fix:** Either:
- Build postgres WITHOUT `--with-system-tzdata` (use bundled tzdata)
- Include tzdata in the bundle at a relative path
- Build with relocatable paths

---

## 3. safeupdate Extension Missing

**Status:** Fixed with CLI workaround

**Issue:** The bundled migrations configure the `authenticator` role with:
```sql
session_preload_libraries=safeupdate
```

But the `safeupdate` extension is not included in the bundle, causing connection failures:
```
FATAL: could not access file "safeupdate": No such file or directory
```

**Current Workaround:** Reset the setting after migrations run:
```sql
-- process-compose.yaml.tmpl:56-57
ALTER ROLE authenticator RESET session_preload_libraries;
```

**Ideal Fix:** Either:
- Include `safeupdate.so`/`safeupdate.dylib` in the bundle
- Remove the `session_preload_libraries=safeupdate` from migrations for sandbox mode

---

## 4. pgsodium/vault getkey Script Paths

**Status:** Fixed with CLI workaround

**Issue:** The `postgresql.conf.template` has commented-out paths for encryption key scripts:
```
#pgsodium.getkey_script = '/path/to/pgsodium_getkey.sh'
#vault.getkey_script = '/path/to/pgsodium_getkey.sh'
```

Without these, postgres fails to start with pgsodium/vault enabled.

**Current Workaround:** Append explicit paths after copying template:
```go
// sandbox.go:129-130
conf += fmt.Sprintf("\npgsodium.getkey_script = '%s'\n", pgsodiumScript)
conf += fmt.Sprintf("vault.getkey_script = '%s'\n", pgsodiumScript)
```

**Ideal Fix:** Use relative paths in the template, or document the expected configuration.

---

## 5. pg_hba.conf Authentication

**Status:** Fixed with CLI workaround

**Issue:** The bundled `pg_hba.conf.template` uses password authentication:
```
host    all    all    127.0.0.1/32    scram-sha-256
```

But `initdb` creates the superuser without a password, so connections fail:
```
FATAL: password authentication failed for user "supabase_admin"
DETAIL: User "supabase_admin" has no password assigned.
```

**Current Workaround:** Override with trust auth for localhost:
```go
// sandbox.go:136-152
host    all    all    127.0.0.1/32    trust
```

**Ideal Fix:** Either:
- Use `trust` in the template for local development
- Document that passwords must be set after initdb
- Provide a separate template for sandbox mode

---

## 6. Locale Path - Hardcoded Nix Path (Warning)

**Status:** Warning only - not blocking

**Issue:** The binary tries to use a Nix-specific locale binary:
```
sh: /nix/store/zkzf0q482kvjs7bb6467sgvbzwj27ggj-adv_cmds-235/bin/locale: No such file or directory
WARNING: no usable system locales were found
```

**Impact:** Warning messages during initdb, but doesn't break functionality.

**Ideal Fix:** Build without hardcoded Nix paths for locale detection.

---

## Summary: Ideal Bundle Requirements

For the bundle to work without workarounds:

| Issue | Fix Required |
|-------|--------------|
| Code signing | Sign binaries during packaging |
| Timezone data | Include tzdata OR build without `--with-system-tzdata` |
| safeupdate | Include extension OR remove from migrations |
| pgsodium paths | Use relative/templated paths in config |
| pg_hba.conf | Use `trust` for local dev OR document setup |
| Locale paths | Build without Nix-hardcoded paths |

jgoux:
I asked Claude to compare what we're doing in the Docker image vs what's included in the artifact.

## How the Docker Image Works (Reference)

The Supabase postgres Docker image (`supabase/postgres`) handles initialization differently and doesn't have these issues. Understanding this helps define what the bundle should include.

### Docker Initialization Flow

1. **Container starts with `POSTGRES_PASSWORD` env var**
   ```go
   // internal/db/start/start.go:64
   "POSTGRES_PASSWORD=" + utils.Config.Db.Password,
   ```

2. **`docker-entrypoint.sh` runs** (from official postgres image):
   - Checks if data directory exists
   - Runs `initdb` if needed
   - Creates `postgres` superuser WITH the password from `POSTGRES_PASSWORD`
   - Runs any init scripts in `/docker-entrypoint-initdb.d/`
   - Starts postgres

3. **Roles are created with hardcoded password 'postgres'** in globals.sql:
   ```sql
   -- internal/utils/templates/globals.sql:19,35
   ALTER ROLE authenticator WITH ... PASSWORD 'postgres';
   ALTER ROLE supabase_admin WITH SUPERUSER ... PASSWORD 'postgres';
   ```

4. **schema.sql updates passwords** to the actual configured password:
   ```sql
   -- internal/db/start/templates/schema.sql:8-14
   \set pgpass `echo "$PGPASSWORD"`
   ALTER USER postgres WITH PASSWORD :'pgpass';
   ALTER USER authenticator WITH PASSWORD :'pgpass';
   -- etc.
   ```

### Why Docker Works and Bundle Doesn't

| Aspect | Docker Image | Native Bundle |
|--------|--------------|---------------|
| Init script | `docker-entrypoint.sh` sets password during init | No init script - `initdb` creates user without password |
| pg_hba.conf | scram-sha-256 works because passwords are set | scram-sha-256 fails - no passwords yet |
| Timezone data | Bundled or proper system paths | Hardcoded Nix paths that don't exist |
| Code signing | Linux - not needed | macOS - invalid signatures after extraction |
| Extensions | All included (safeupdate, etc.) | safeupdate missing |

---

## Recommended Bundle Changes

To make the bundle work like the Docker image without CLI workarounds:

### 1. Add an Initialization Script

Include a `supabase-postgres-init` script similar to `docker-entrypoint.sh`:

```bash
# share/supabase-cli/bin/supabase-postgres-init.sh
#!/bin/bash
set -e

PGDATA="${PGDATA:-./data}"
POSTGRES_USER="${POSTGRES_USER:-supabase_admin}"
POSTGRES_PASSWORD="${POSTGRES_PASSWORD:-postgres}"

# Get the directory where this script lives
SCRIPT_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
PGBIN="$SCRIPT_DIR/bin"

# Initialize if needed
if [ ! -s "$PGDATA/PG_VERSION" ]; then
    "$PGBIN/initdb" -D "$PGDATA" -U "$POSTGRES_USER" --encoding=UTF8 --locale=C

    # Set password for the superuser immediately (using single-user mode)
    "$PGBIN/postgres" --single -D "$PGDATA" -j <<-EOSQL
        ALTER USER "$POSTGRES_USER" WITH PASSWORD '$POSTGRES_PASSWORD';
EOSQL

    # Copy config templates
    cp "$SCRIPT_DIR/share/supabase-cli/config/postgresql.conf.template" "$PGDATA/postgresql.conf"
    cp "$SCRIPT_DIR/share/supabase-cli/config/pg_hba.conf.template" "$PGDATA/pg_hba.conf"
fi

# Start postgres (or exec to replace shell)
exec "$PGBIN/postgres" -D "$PGDATA" "$@"
```

### 2. Bundle Timezone Data

**Option A** (Recommended): Build postgres WITHOUT `--with-system-tzdata`
- This uses bundled tzdata in `share/postgresql/timezone/`
- No hardcoded paths

**Option B**: Include tzdata in bundle
- Add `share/zoneinfo/` directory
- Build with that relative path

### 3. Include safeupdate Extension

Add to bundle:
```
lib/safeupdate.so              # Linux
lib/safeupdate.dylib           # macOS
share/extension/safeupdate.control
share/extension/safeupdate--*.sql
```

### 4. Fix Config Templates

**pg_hba.conf.template** - Can use scram-sha-256 if init script sets password:
```
# pg_hba.conf - passwords are set by init script
local   all    all                     scram-sha-256
host    all    all    127.0.0.1/32     scram-sha-256
host    all    all    ::1/128          scram-sha-256
```

**postgresql.conf.template** - Use relative/discoverable paths:
```ini
# Use path relative to PGDATA or make it configurable
# pgsodium.getkey_script should be set by init script based on install location
```

### 5. Sign macOS Binaries During Build

Add to build process:
```bash
# Sign all binaries and dylibs before creating zip
find . -type f \( -perm +111 -o -name "*.dylib" \) -exec codesign -s "Developer ID" {} \;
```

Or use ad-hoc signing:
```bash
find . -type f \( -perm +111 -o -name "*.dylib" \) -exec codesign -f -s - {} \;
```

### 6. Build Without Nix-Specific Paths

Remove from configure:
- `--with-system-tzdata=/nix/store/...`
- Any locale paths pointing to `/nix/store/`

---

## Ideal Bundle Structure

```
supabase-postgres-<version>-<os>-<arch>.zip
├── bin/
│   ├── postgres
│   ├── initdb
│   ├── psql
│   ├── pg_isready
│   └── ...
├── lib/
│   ├── *.so / *.dylib           # All shared libraries
│   ├── safeupdate.so            # MISSING: include this
│   └── ...
├── share/
│   ├── postgresql/
│   │   ├── timezone/            # NEEDED: bundled tzdata
│   │   └── extension/
│   │       └── safeupdate*      # MISSING: include this
│   └── supabase-cli/
│       ├── bin/
│       │   └── supabase-postgres-init.sh   # NEW: init script
│       ├── config/
│       │   ├── postgresql.conf.template
│       │   ├── pg_hba.conf.template
│       │   └── pgsodium_getkey.sh
│       └── migrations/
│           └── migrate.sh
```

---

## Summary of Build Changes Required

| Issue | Build Change |
|-------|--------------|
| Timezone data | Remove `--with-system-tzdata=/nix/store/...` |
| Locale paths | Remove Nix-specific locale paths |
| Code signing (macOS) | Add `codesign` step to build |
| safeupdate | Add to extension build list |
| Config paths | Use relative paths or make configurable |
| Init script | Add `supabase-postgres-init.sh` wrapper |

With these changes, the CLI would only need to:
1. Extract the zip
2. Run `supabase-postgres-init.sh` with appropriate env vars
3. Run migrations

No workarounds needed.

LGUG2Z:
@jgoux 

I've updated the PR to

* Comprehensively strip out hardcoded paths at compile time and added the integration test we discussed on Slack
* Added the safeupdate ext
* Use `trust` in the hba conf for now until we decide how to move ahead with the init

I have not made any changes to the template - I think it's normal to update a config template at runtime

Adhoc codesigning isn't working and is hard to debug in CI, I think it's ok to continue with client-side codesigning for macOS for now since that is what homebrew kegs do when installing

Let's discuss the init with others on Slack and then we can update this PR with whatever is decided

LGUG2Z:
@coderabbit pause

PR #2034, opened 2026-02-03
slim builds for postgres versions just latest extension

samrose:
a build of postgres with just the latest extension version and postgres itself

## Summary by CodeRabbit

* **New Features**
  * Slim PostgreSQL package variants for smaller container images.
  * Optional "latest-only" build mode for many PostgreSQL extensions to produce single-version builds.

* **Chores**
  * Updated PostgreSQL version identifiers to include slim build metadata.


samrose:
We won't need this one after all closing

PR #2037, opened 2026-02-05
fix: nixos tests

samrose:
Trying to fix nixos-tests issues


samrose:
was not the issue closing

PR #2038, opened 2026-02-06
refactor nixos tests to use our config, build, migrations, init etc 

samrose:
Started with refactor of pg_repack to use this new module. Will then use this as basis to refactor/add the rest.

Will then update this PR description with the outcome


encima:
Could we also update the documents for the running of the tests and how they currently work with this PR?

nix/ext/tests/lib.nix
```diff
@@ -0,0 +1,586 @@
...
+
+  defaultPort = 5432;
+
+  # Simple getkey script that returns a static hex key for testing
+  getkeyScript = pkgs.pkgsLinux.writeShellScript "pgsodium-getkey" ''
+    echo "0000000000000000000000000000000000000000000000000000000000000000"
+  '';
+
+  # Source paths for ansible config files and migrations
+  ansibleConfigDir = builtins.path {
```
encima, line 19: This is very cool (unifying nix with ansible more closely) but it does mean that changes from ansible have to be incorporated here and vice-versa; is there a way to unify it ? Maybe pulling in from the Ansible files?

nix/ext/tests/lib.nix
```diff
@@ -0,0 +1,586 @@
...
+      } > $out/pg_hba.conf.tmp
+      mv $out/pg_hba.conf.tmp $out/pg_hba.conf
+
+      # Add ident mappings for root -> supabase_admin and postgres -> supabase_admin
+      echo "supabase_map root supabase_admin" >> $out/pg_ident.conf
+      echo "supabase_map postgres supabase_admin" >> $out/pg_ident.conf
+
+      # Version-specific adjustments (mirroring run-server.sh.in:250-295)
+      ${
+        if majorVersion == "17" || majorVersion == "orioledb-17" then
```
encima, line 103: Also like this handling of the config based on major version in a single place! Could be useful to do this outside of tests as well maybe?

nix/ext/tests/vault.nix
```diff
@@ -2,200 +2,166 @@
...
       start_all()
 
-      server.wait_for_unit("multi-user.target")
-      server.wait_for_unit("postgresql.service")
+      # Wait for full Supabase initialization (postgres + init-scripts + migrations)
+      server.wait_for_unit("supabase-db-init.service")
 
-      test = PostgresExtensionTest(server, extension_name, versions, sql_test_directory, support_upgrade)
+      with subtest("Verify PostgreSQL 15 is our custom build"):
+        pg_version = server.succeed(
```
encima, line 58: you can also check in the data dir under `PG_VERSION` which saves a connection

nix/ext/tests/vault.nix
```diff
@@ -2,200 +2,166 @@
...
+
+      with subtest("Verify PostgreSQL 17 is our custom build"):
+        pg_version = server.succeed(
+          "psql -U supabase_admin -d postgres -t -A -c \"SELECT version();\""
+        ).strip()
+        assert "${testLib.expectedVersions."17"}" in pg_version, (
+          f"Expected version ${testLib.expectedVersions."17"}, got: {pg_version}"
+        )
+
+        postgres_pid = server.succeed(
```
encima, line 120: Most extensions have the same boilerplate for "check version" and "check is running", it seems like these would be better as "pre tests" and the extensions themselves would just specify the PG versions they should run on (as they currently do)

nix/ext/tests/lib.nix
samrose: yes, because we're using ansible config files as source of truth if someone reorganized where they live path-wise, would need to update this too. However they were made uniform fairly recently, and so we should not have a need to move them around. The alternative would be to turn them into a nix module, which we can't get into doing in this PR, as it would greatly increase the scope of it.

nix/ext/tests/lib.nix
samrose: We definitely do this, although I may be missing specifics that you are thinking about.

encima (approved):

encima (approved):

LGUG2Z (approved):

PR #2039, opened 2026-02-07
fix: add limit to nix copy to avoid hitting s3 upload limits

samrose:
in PR ci runs like https://github.com/supabase/postgres/actions/runs/21781372775/job/62845918939?pr=2028 we hit s3 upload limits since no max-jobs was set, nix copy has no limit.

We have multiple CI jobs potentially uploading concurrently to the same /nar/ prefix, a conservative per-job limit would be:

  - 5 concurrent uploads (nix copy --max-jobs 5)
  - Or roughly 50 requests/second per job

This should help with the fact that we have potential to launch 5-10 CI runs at once.

### DIFF
--- a/.github/actions/nix-install-ephemeral/action.yml
+++ b/.github/actions/nix-install-ephemeral/action.yml
@@ -35,7 +35,7 @@ runs:
         set -f
 
         export IFS=' '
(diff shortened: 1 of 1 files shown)


valigula (approved):
LGTM

soedirgo (approved):

PR #2054, opened 2026-02-17
feat: images tests and release on pg 17 and orioledb for multigres

samrose:
New files (2):                                                                                                                               
  - Dockerfile-multigres-17 — near-identical copy of Dockerfile-17 with 4 changes (no docker-entrypoint.sh, no WAL-G scripts, tail -f /dev/null
   entrypoint, no pgBackRest)                                                                                                                  
  - Dockerfile-multigres-orioledb-17 — same treatment for the OrioleDB variant 

  Modified files (4):
  - .github/workflows/docker-image-test.yml — added multigres Dockerfiles to matrix, skip standard tests for them, added smoke test with
  --entrypoint "" override
  - .github/workflows/dockerhub-release-matrix.yml — auto-discovers multigres Dockerfiles, derives -multigres tags from base variant versions
  - .github/workflows/manual-docker-release.yml — same auto-discovery + tag derivation, plus preserves -multigres suffix when
  inputs.postgresVersion is set
  - nix/packages/docker-image-inputs.nix — tracks the two new Dockerfiles in the fileset

Goal, create 2 images that multigres can use to integrate with supabase postgres, and orioledb postgres


Dockerfile-multigres-17
```diff
@@ -0,0 +1,189 @@
...
+
+# Setup configs
+COPY --chown=postgres:postgres ansible/files/postgresql_config/postgresql.conf.j2 /etc/postgresql/postgresql.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_hba.conf.j2 /etc/postgresql/pg_hba.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_ident.conf.j2 /etc/postgresql/pg_ident.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/conf.d /etc/postgresql-custom/conf.d
+COPY --chown=postgres:postgres ansible/files/postgresql_config/postgresql-stdout-log.conf /etc/postgresql/logging.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/supautils.conf.j2 /etc/postgresql-custom/supautils.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_extension_custom_scripts /etc/postgresql-custom/extension-custom-scripts
+COPY --chown=postgres:postgres ansible/files/pgsodium_getkey_urandom.sh.j2 /usr/lib/postgresql/bin/pgsodium_getkey.sh
```
deepthi: Can we remove all references to pgsodium? Docs say that we will be deprecating it, and no new usage.

samrose: deprecation of pgsodium in vault still WIP. After discussion with @deepthi and @soedirgo we decided to retain this for now. Will follow up with PR and effort to finish deprecation ASAP

Dockerfile-multigres
```diff
@@ -0,0 +1,290 @@
...
+    ln -s /usr/lib/postgresql/bin/pgsodium_getkey.sh /usr/share/postgresql/extension/pgsodium_getkey && \
+    chmod +x /usr/lib/postgresql/bin/pgsodium_getkey.sh
+
+COPY migrations/db /docker-entrypoint-initdb.d/
+COPY ansible/files/pgbouncer_config/pgbouncer_auth_schema.sql /docker-entrypoint-initdb.d/init-scripts/00-schema.sql
+COPY ansible/files/stat_extension.sql /docker-entrypoint-initdb.d/migrations/00-extension.sql
+
+ENV PGDATA=/var/lib/postgresql/data
+ENV POSTGRES_HOST=/var/run/postgresql
+ENV POSTGRES_USER=supabase_admin
```
mkindahl, line 164: This will work with multigres after https://github.com/multigres/multigres/pull/683 is merged.

Dockerfile-multigres
```diff
@@ -0,0 +1,290 @@
...
+ENV POSTGRES_DB=postgres
+ENV POSTGRES_INITDB_ARGS="--allow-group-access --locale-provider=icu --encoding=UTF-8 --icu-locale=en_US.UTF-8"
+ENV LANG=en_US.UTF-8
+ENV LANGUAGE=en_US:en
+ENV LC_ALL=en_US.UTF-8
```
mkindahl, line 169: Since these are set in the containers (as they should be), we should probably refrain from setting them in production code and merge https://github.com/multigres/multigres/pull/691

.github/workflows/docker-image-test.yml
```diff
@@ -52,18 +52,20 @@ jobs:
...
     timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:
-        dockerfile:
-          - Dockerfile-15
-          - Dockerfile-17
-          - Dockerfile-orioledb-17
+        include:
+          - { dockerfile: Dockerfile-15,          target: "",                  name: 15                      }
```
mkindahl, line 64: Not important, but if you want, you can use tabular YAML.

```suggestion
          - dockerfile: Dockerfile-15
            name: 15
```

docker/pgctld/orioledb-postgresql.conf.tmpl
```diff
@@ -0,0 +1,314 @@
...
+
+# - Other Platforms and Clients -
+
+
+#------------------------------------------------------------------------------
+# ERROR HANDLING
+#------------------------------------------------------------------------------
+
+#exit_on_error = off			# terminate session on any error?
+#restart_after_crash = on		# reinitialize after backend crash?
```
mkindahl: I am not sure we want this for normal k8s deployments. When deployed in k8s the pod is normally restarted by k8s: the server inside the pod does not automatically restart.

It depends on what best practices that Supabase normally advocates.

docker/pgctld/orioledb-postgresql.conf.tmpl
```diff
@@ -0,0 +1,314 @@
...
+
+# default configuration for text search
+default_text_search_config = 'pg_catalog.english'
+
+# These settings are initialized by initdb, but they can be changed.
+lc_messages = 'en_US.UTF-8'			# locale for system error message
+					# strings
+lc_monetary = 'en_US.UTF-8'			# locale for monetary formatting
+lc_numeric = 'en_US.UTF-8'			# locale for number formatting
+lc_time = 'en_US.UTF-8'				# locale for time formatting
```
mkindahl: Since the locale is already set in the container, should we also set the local in the configuration file? Will that not just lead to confusion if you, for example, change the setting in the container and this is overridden by the setting in the configuration file?

docker/pgctld/orioledb-postgresql.conf.tmpl
```diff
@@ -0,0 +1,314 @@
...
+
+# default configuration for text search
+default_text_search_config = 'pg_catalog.english'
+
+# - Shared Library Preloading -
+
+## TODO: @rafael - We are starting from a Vanilla postgres. We still
+## ened to do work here to make sure we start postgres with libraries that
+## will be used by supabase infra.
+# shared_preload_libraries = 'pg_stat_statements, pgaudit, plpgsql, plpgsql_check, pg_cron, pg_net, pgsodium, timescaledb, auto_explain, pg_tle, plan_filter, supabase_vault'	# (change requires restart)
```
mkindahl: Shouldn't this be the same as for normal Supabase configurations?

nix/packages/pgctld.nix
```diff
@@ -0,0 +1,31 @@
...
+buildGoModule {
+  pname = "pgctld";
+  version = "0.1.0";
+  src = multigres-src;
+  subPackages = [ "go/cmd/pgctld" ];
+  env.CGO_ENABLED = "0";
+  ldflags = [
+    "-s"
+    "-w"
+  ];
```
mkindahl: I think we might want to have debug symbols in the build to be able to get stack traces and debug problems we run into. Is the size of the binary big enough to warrant stripping symbols and DWARF?

Dockerfile-multigres
```diff
@@ -0,0 +1,290 @@
...
+    adduser -S -s /bin/bash -G wal-g wal-g
+
+RUN cat <<EOF > /tmp/extra-nix.conf
+extra-experimental-features = nix-command flakes
+extra-substituters = https://nix-postgres-artifacts.s3.amazonaws.com
+extra-trusted-public-keys = nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI=
+EOF
+RUN curl -L https://releases.nixos.org/nix/nix-2.33.2/install | sh -s -- --daemon --no-channel-add --yes --nix-extra-conf-file /tmp/extra-nix.conf
+
+ENV PATH="${PATH}:/nix/var/nix/profiles/default/bin"
```
mkindahl: The start of each of these stages are identical; you might want to create a separate (builder) base stage for these settings to ease maintenance and build faster.

cuongdo:
https://github.com/supabase/postgres/actions/runs/22699135346/job/65812194561?pr=2054#step:6:1411

shows that pgbackrest 2.57.0 is being pulled in. Ideally, we'd use 2.58.0, but 2.57.0 is fine. The pgbackrest part specifically looks good to me. I'm not sure how to review the rest.

Dockerfile-multigres
```diff
@@ -0,0 +1,290 @@
...
+    chmod +x /usr/lib/postgresql/bin/pgsodium_getkey.sh
+
+COPY migrations/db /docker-entrypoint-initdb.d/
+COPY ansible/files/pgbouncer_config/pgbouncer_auth_schema.sql /docker-entrypoint-initdb.d/init-scripts/00-schema.sql
+COPY ansible/files/stat_extension.sql /docker-entrypoint-initdb.d/migrations/00-extension.sql
+
+ENV PGDATA=/var/lib/postgresql/data
+ENV POSTGRES_HOST=/var/run/postgresql
+ENV POSTGRES_USER=supabase_admin
+ENV POSTGRES_DB=postgres
```
mkindahl, line 165: We do currently not read this environment variable to figure out the database to connect to. We should probably add that to multigres.

mkindahl (commented):
Thanks for working on this. I have some short questions below for your consideration.

cuongdo:
I took another look at this, and this looks like a good starting point for `pgctld` and `pgbackrest`. We will have to test and iterate to get things completely right, because Multigres generates Postgres config files based on its own template. We will need to modify the Multigres code to use the Postgres config here as a starting point. This PR looks good from my point of view.

Dockerfile-multigres
samrose: likely needs to be resolved in multigres

Dockerfile-multigres
samrose: consolidated as much as possible

nix/packages/pgctld.nix
samrose: removed per request

docker/pgctld/orioledb-postgresql.conf.tmpl
samrose: got it straightened out thanks for pointing that out

docker/pgctld/orioledb-postgresql.conf.tmpl
samrose: fixed!

docker/pgctld/orioledb-postgresql.conf.tmpl
samrose: followed your request and turned it off by default

Dockerfile-multigres
samrose: resolved

Dockerfile-multigres
samrose: pulled in latest changes

soedirgo (approved):

nix/tests/prime-multigres.sql
```diff
@@ -0,0 +1,88 @@
+-- prime-multigres.sql
+-- Like prime.sql but omits extensions unavailable in multigres images:
+--   pgsodium  - requires pgsodium_getkey.sh (not present in multigres images)
```
Crispy1975, line 3: I think `pgsodium_getkey.sh` is present: https://github.com/supabase/postgres/pull/2054/changes#diff-81487dc9b62aa6169873ec3dbfd88d9307e00289d5ea4c9fcbad2280565d5646R136t

Dockerfile-multigres
mkindahl: Available in https://github.com/multigres/multigres/pull/706

mkindahl (approved):
LGTM

nix/tests/prime-multigres.sql
```diff
@@ -0,0 +1,88 @@
...
+create extension if not exists moddatetime;
+create extension if not exists pageinspect;
+create extension if not exists pg_buffercache;
+
+/*
+TODO: Does not enable locally mode
+requires a change to postgresql.conf to set
+cron.database_name = 'testing'
+*/
+-- create extension if not exists pg_cron;
```
Crispy1975, line 41: Intentionally disabled?

flake.nix
```diff
@@ -13,6 +13,8 @@
     flake-utils.url = "github:numtide/flake-utils";
     git-hooks.inputs.nixpkgs.follows = "nixpkgs";
     git-hooks.url = "github:cachix/git-hooks.nix";
+    multigres.url = "github:multigres/multigres";
+    multigres.flake = false;
```
Crispy1975, line 17: Would running `nix flake update` silently upgrade to newer versions? The lock file is pinned to a rev, but here we are not? Should we track a specific release?

nix/tests/prime-multigres.sql
samrose: yes this comment needs to be removed agree

flake.nix
samrose: yes, but that would cause a problem for any nix flake


There is a flake.lock file that will track the version, instead of pinning it in the flake. Our convention here is to run `nix flake update multigres` and it will only target that one. It will advance multigres to the latest version on the main branch, which is what we want now.

PR #1978, opened 2025-12-12
chore: add monthly flake.lock inputs update workflow

yvan-sraka:
Adds GitHub Action to automatically update `flake.lock` every Month. Uses  [Mic92/update-flake-inputs](https://github.com/Mic92/update-flake-inputs) to create PRs with updated dependencies, preventing large, painful updates like the recent 2-year gap... https://github.com/supabase/postgres/pull/1714

Targets develop branch with automated and dependencies labels. Can also be triggered manually via `workflow_dispatch`.

The workflow is configured with `GITHUB_TOKEN` which works but **won't trigger CI workflows** on the created pull requests (GitHub prevents this to avoid infinite loops).

For CI workflows to run on the created PRs, you should set up a GitHub App:

1. **Easy setup**: Use the [web interface](https://mic92.github.io/update-flake-inputs/) to create a GitHub App with correct permissions
2. **Configure secrets**: Save the App ID as `APP_ID` and private key as `APP_PRIVATE_KEY` in repository secrets
3. **Update workflow**: Replace the `github-token` step to use the GitHub App token

See the [full documentation](https://github.com/Mic92/update-flake-inputs#using-with-github-app-token-recommended) for detailed instructions.

The current basic setup will work fine for creating PRs, the GitHub App is only needed if you want CI to automatically run on those PRs.

## Summary by CodeRabbit

* **Chores**
  * Added an automated workflow to update the project's lockfile and keep dependencies current.
  * Runs monthly and can be triggered manually; will open automated pull requests with appropriate labels to review and merge updates.

PR #2055, opened 2026-02-17
chore: `$ nix flake update`

yvan-sraka:
Since https://github.com/supabase/postgres/pull/1978 haven't been merged yet, I ran some `nix flake update` by hand!


samrose:
we'll address the update soon, but don't want to use the app in #1978 and can create our own way to automate this

PR #2058, opened 2026-02-19
fix: reserve additional port for adminapi

delgado3d:


ansible/tasks/setup-system.yml
```diff
@@ -167,10 +167,11 @@
         state: 'present'
 
     # postgres_exporter runs on port 9187 and postgresT occasionlly chooses it as random srcport
-    - name: Set net.ipv4.ip_local_reserved_ports=9187
+    # adminapi for 8085
+    - name: Set net.ipv4.ip_local_reserved_ports
       ansible.builtin.sysctl:
         name: 'net.ipv4.ip_local_reserved_ports'
-        value: 9187
+        value: '9187,8085'
```
darora, line 174: should we add the ports for gotrue, adminapi and other services we run as well?

darora (approved):

ansible/tasks/setup-system.yml
delgado3d: would like to get this fix out first

PR #2068, opened 2026-02-26
fix: bump to release auth latest

samrose:
just needed to increment one more to get release out

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.046-orioledb"
-  postgres17: "17.6.1.089"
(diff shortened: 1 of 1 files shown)


valigula (approved):
LGTM

ansible/vars.yml
Crispy1975, line 28: Not bumped?

samrose: good catch, althoough it was actually bumped here https://github.com/supabase/postgres/pull/2066/changes#diff-372784febf11c83d9ecf943bec37e9d020784ad32dc405f033cf7552d2923506L28

But then someone has a PR that stomped on the version that Chris tried to increment the AMI release to. So, the version bump of auth merged to develop branch, but the release of AMI failed. So this pr fixes that.

Crispy1975 (approved):

PR #2069, opened 2026-02-26
feat(tuned): ensure CPUs are giving it their all

hunleyd:
This commit enhances the custom `tuned` profile for PostgreSQL by forcing the CPU into a maximum performance
state. These changes are critical for maintaining consistent low-latency database performance.

Key Changes and Benefits:

- CPU Governor & Energy Bias: Switched to 'performance' mode. This prevents the kernel from scaling down CPU frequencies during periods of lower activity, eliminating the "ramp-up" latency when a burst of queries arrives.
- Min Performance Percentage: Fixed at 100% to ensure the processor stays at its maximum base frequency, reducing jitter in query execution times.

(ripgrep of all Supa repos for the 3 tunables herein returned nothing, so no chance of clobbering the setting)


hunleyd:
```
Test Summary

ENGINE          RESULT
------          ------
15              PASSED (435s)
17              PASSED (352s)
17-oriole       PASSED (716s)

[INFO] Postgres commit: e568cea207339c1243b920512f4dfb738ae4778a
[INFO] Cache location: /Users/jimchancojr/.cache/supadev/postgres/e568cea207339c1243b920512f4dfb738ae4778a
[SUCCESS] All tests passed!
```

samrose (changes requested):
Critical Issues (fix before merge)

  1. Typo: min_pef_pct → min_perf_pct (setup-tuned.yml:79)

  The option name is misspelled — missing the r in perf. The tuned daemon will silently write the malformed key and ignore it, meaning the CPU minimum P-state floor will
  never be enforced despite the intent. This is the most impactful bug.

  # Wrong:
  - { section: 'cpu', option: 'min_pef_pct',  value: '100' }
  # Correct:
  - { section: 'cpu', option: 'min_perf_pct', value: '100' }

  2. Symlink creation will fail on existing hosts (setup-tuned.yml:23-31)

  If a host was provisioned before this branch (where /etc/tuned/postgresql was a directory), the state: 'link' task will fail with [Errno 17] File exists. Ansible won't
  remove an existing directory to create a symlink in its place. This is lower risk in an AMI build workflow (fresh builds), but needs a force: true or pre-removal task for
  safe re-application.

  ---
  Important Issues

  3. ARM/Graviton instances silently ignore energy_perf_bias and P-state settings: these are Intel-only controls. Consider a when: platform == 'amd64' guard or
  documentation.

  4. tuned-adm profile postgresql always reports changed: add changed_when: false since activating an already-active profile is a no-op.

  5. Unconditional state: 'restarted' on tuned service:: safe in AMI builds, but risky if the playbook is ever re-applied to a live production instance (causes a brief
  performance regression window).

hunleyd:
> 5. Unconditional state: 'restarted' on tuned service:: safe in AMI builds, but risky if the playbook is ever re-applied to a live production instance (causes a brief
>    performance regression window).

i think this is acceptable as that's how we currently run (w/o optimizations) and we definitely want any new updates to the profile to be applied, which we wouldn't get with `started`.

samrose (approved):

PR #2070, opened 2026-02-27
feat: use blacksmith for nixos tests on aarch64-linux

jfroche:


samrose:
made it work, moved this work to https://github.com/supabase/postgres/pull/2085 to complete it

PR #2072, opened 2026-02-27
Install fail2ban using system manager

picnoir:
Introducing a system manager fail2ban module. This module introduce
two extra filters and fails on top of the builtin ones adapted from
the current ansible deployment for postgresql and pgbouncer.

NOTE: this PR targets https://github.com/supabase/postgres/pull/2063


brainrake:

PR #2074, opened 2026-03-01
fix: capture UUID before unmounting in unmount

valigula:
`findmnt` queries /proc/mounts, so the UUID lookup must happen before `umount` is called.  And clean the fstab entry if exists.


samrose:
@valigula do we still need to get this tested and merged?

brainrake:
@valigula is this still causing a problem? where and when?

PR #2077, opened 2026-03-02
perf(tuned): activate and configure zswap for postgresql tuned profile

hunleyd:
This change enables zswap with the zsmalloc allocator and zstd compression within the postgresql tuned profile to improve system responsiveness and database performance under memory pressure.

### Technical Overview: What is zswap?
zswap is a Linux kernel feature that provides a compressed write-back cache for swapped pages. Instead of immediately moving pages from RAM to the physical swap device (HDD/SSD) when memory is full, zswap intercepts them, compresses them, and stores them in a dynamically allocated RAM pool. This effectively increases the "perceived" amount of RAM available to the system.

### Why zsmalloc over zbud?
We have selected 'zsmalloc' as the zpool allocator instead of 'zbud' for several reasons:
- Density: zsmalloc is a slab-based allocator that can pack many compressed objects into a single physical page, whereas zbud is hard-coded to a maximum of 2 objects per page (2:1 ratio).
- Efficiency: On modern kernels (Linux 6.3+), zsmalloc supports proper eviction to disk, removing the primary historical advantage of zbud. It provides significantly better memory utilization and higher compression density.

### Why zstd over lz4?
While lz4 offers faster compression/decompression speeds, we have opted for 'zstd' for the following reasons:
- Compression Ratio: zstd provides a substantially higher compression ratio than lz4. In database workloads where memory density is critical, fitting more pages into the zswap pool is more beneficial than the marginal speed gains of lz4.
- Balance: zstd offers a superior balance between CPU overhead and compression efficiency, especially for the 4KB-8KB pages typically handled by the kernel and PostgreSQL.

### Benefits for PostgreSQL Performance
Implementing zswap is particularly beneficial for PostgreSQL workloads:
- Mitigation of Swap Thrashing: PostgreSQL performance degrades exponentially when active buffers are swapped to disk. zswap ensures that these pages stay in RAM (compressed), allowing "major page faults" to be resolved at RAM speeds rather than waiting for disk I/O.
- I/O Latency Stability: By reducing the frequency of physical writes to the swap partition, zswap prevents I/O contention between the kernel's swap subsystem and PostgreSQL's own WAL and data writes.
- Consistent Query Latency: Queries that access less-frequently used data that has been compressed into zswap will return much faster than if they had to be fetched from physical swap, leading to more predictable p99 latencies.

### DIFF
--- a/ansible/tasks/setup-tuned.yml
+++ b/ansible/tasks/setup-tuned.yml
@@ -106,8 +106,7 @@
         params: 'zpool=zsmalloc compressor=zstd enabled=1'
         persistent: 'present'
         state: 'present'
-      when:
-        - ansible_facts['distribution'] == 'Ubuntu'  # AWS runners don't have this module
+      failed_when: false  # AWS runners don't have this module
 
     - name: Activate the tuned service
       ansible.builtin.systemd_service:
--- a/ansible/tasks/setup-tuned.yml
+++ b/ansible/tasks/setup-tuned.yml
@@ -124,20 +124,13 @@
             state: 'present'
 
         - name: tuned - Configure and enable zswap
-          ansible.builtin.lineinfile:
-            group: 'root'
-            line: "{{ zswap_item['value'] }}"
-            mode: '0644'
-            owner: 'root'
-            path: "/sys/module/zswap/parameters/{{ zswap_item['param'] }}"
-            search_string: "{{ zswap_item['orig_value'] }}"
-            state: 'present'
-          become: true
+          ansible.builtin.shell:
+            cmd: "echo {{ zswap_item['value'] }}  > /sys/module/zswap/parameters/{{ zswap_item['param'] }}"
           loop:
-            - { param: 'compressor',       value: 'zstd',     orig_value: 'lzo' }
-            - { param: 'max_pool_percent', value: '10',       orig_value: 20 }
-            - { param: 'zpool',            value: 'zsmalloc', orig_value: 'zbud' }
-            - { param: 'enabled',          value: 'Y',        orig_value: 'N' }
+            - { param: 'compressor',       value: 'zstd' }
+            - { param: 'max_pool_percent', value: '10' }
+            - { param: 'zpool',            value: 'zsmalloc' }
+            - { param: 'enabled',          value: 'Y' }
           loop_control:
             loop_var: 'zswap_item'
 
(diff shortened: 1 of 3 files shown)


ansible/tasks/setup-tuned.yml
```diff
@@ -99,6 +99,18 @@
...
+        create: true
+        group: 'root'
+        mode: '0644'
+        no_extra_spaces: true
+        option: 'cmdline_zswap'
+        path: '/etc/tuned/profiles/postgresql/tuned.conf'
+        section: 'bootloader'
+        state: 'present'
+        value: 'zswap.enabled=1 zswap.zpool=zsmalloc zswap.compressor=zstd'
     - name: Activate the tuned service
```
hunleyd:
```suggestion
        value: 'zswap.enabled=1 zswap.zpool=zsmalloc zswap.compressor=zstd'

- name: Activate the tuned service
```

hunleyd:
@samrose i do _not_ like having to add that when, but the runners don't have the module. open to ideas

hunleyd:
```
Test Summary

ENGINE          RESULT
------          ------
15              PASSED (397s)
17              PASSED (3839s)
17-oriole       PASSED (914s)

[INFO] Postgres commit: 420197266c6095c95f198651b50dc944a182a6bc
[INFO] Cache location: /Users/jimchancojr/.cache/supadev/postgres/420197266c6095c95f198651b50dc944a182a6bc
[SUCCESS] All tests passed!
```

tgallacher (approved):
s/lgtm

samrose (changes requested):
need to discuss and review before merge

hitmands (approved):

ansible/vars.yml
```diff
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.050-orioledb"
-  postgres17: "17.6.1.093"
-  postgres15: "15.14.1.093"
+  postgresorioledb-17: "17.6.0.051-orioledb"
+  postgres17: "17.6.1.094"
+  postgres15: "15.14.1.094"
```
samrose: you will need to compare what the latest actual release is and incment these before merge, or your release will fail

samrose (approved):
Just need to address accurate increment of version. Or it will have to be done over again when you merge, and the release will fail

ansible/vars.yml
hunleyd: tweaked

samrose:
All tests passing https://github.com/supabase/supadev/actions/runs/22747411633

PR #2079, opened 2026-03-03
chore: use apt's built-in (undocumented) lock checker

tomashley:
* set to 600s/10mins
* https://bugs.launchpad.net/cloud-init/+bug/1693361/comments/23
* undocumented feature added in ubuntu 20.04
* checks every second for lock to be released
* default is 120sec on `apt` commands, but nothing on `apt-get` so we add it

** Pkill any apt-get procs over 1 hour old


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -375,13 +375,16 @@ EXTRA_NIX_CONF
...
+    # One hour is old enough to be bad
+    pkill -f apt-get --older 3600 2>/dev/null || echo "No hanging apt-get processes found"
     DEBIAN_FRONTEND=noninteractive dpkg --configure -a --force-confold || true # handle errors generated by dpkg
 
     # Needed for PostGIS, since it's compiled with Protobuf-C support now
     echo "3. Installing libprotobuf-c1 and libicu66 if missing"
     if [[ ! "$(apt list --installed libprotobuf-c1 | grep "installed")" ]]; then
-        apt-get update -y
-        apt --fix-broken install -y libprotobuf-c1 libicu66 || true
+        apt-get -o DPkg::Lock::Timeout=600 update -y # wait up to 10 minutes for any dpkg locks to clear before updating package lists
```
hunleyd, line 386: any reason we don't use `apt update`? wouldnt that get us the built-in 2m timeout

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -400,8 +403,8 @@ EXTRA_NIX_CONF
         UBUNTU_VERSION=$(lsb_release -rs)
         # install amazon disk utilities if not present on 24.04
         if [ "${UBUNTU_VERSION}" = "24.04" ] && ! /usr/bin/dpkg-query -W amazon-ec2-utils >/dev/null 2>&1; then
-            apt-get update
-            apt-get install -y amazon-ec2-utils || true
+            apt-get -o DPkg::Lock::Timeout=600 update
+            apt-get -o DPkg::Lock::Timeout=600 install -y amazon-ec2-utils || true
```
hunleyd, line 407: can't these also be `apt`?

hunleyd (changes requested):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
tomashley: old school people know `apt-get` before `apt` was a thing alone

I'm adding a 10 minute timeout here, so we'd need to set the option regardless of which tool we use.
Traditionally everything is `apt-get` so I'd lean to keep it, unless we align all commands

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
tomashley: same comment as above

hunleyd (approved):

PR #2081, opened 2026-03-05
chore: bump minor version gatekeeper

staaldraad:
Minor version bump to grab latest version with a small bug fix.

### DIFF
--- a/nix/packages/gatekeeper.nix
+++ b/nix/packages/gatekeeper.nix
@@ -6,11 +6,11 @@ let
 
   upstream-gatekeeper = buildGoModule124 {
     pname = "jit-db-gatekeeper";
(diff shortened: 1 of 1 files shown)


samrose (approved):
@staaldraad approving to unblock, I take it this is not a live feature in the image yet so doesn't need testing for functionality is that correct?

staaldraad:
> @staaldraad approving to unblock, I take it this is not a live feature in the image yet so doesn't need testing for functionality is that correct?

correct, I have also double checked by installing it with Salt in staging. This didn't change any functionality but addressed an edge case. 
It not being live is why I also opted to skip an image version bump and to leave this to join the next image version naturally

PR #2090, opened 2026-03-18
feat: reduce AMI data volume size

jfreeland:
When we build an instance we explicitly specify the size of the data volume.  The data volume that's built as part of this AMI should be 1GB.


ansible/vars.yml
```diff
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.055-orioledb"
-  postgres17: "17.6.1.098"
-  postgres15: "15.14.1.098"
+  postgresorioledb-17: "17.6.0.055-orioledb-testdatavol"
```
jfreeland: these need to be reset before this is merged

samrose (approved):

hitmands (approved):

samrose (approved):

PR #2094, opened 2026-03-26
chore: upgrade SAA version

tomashley:
align with version declared in salt


soedirgo:
It's still 1.6.0 in v3staging & v3prod fwiw

soedirgo (approved):

PR #2105, opened 2026-04-03
Fix: allow safe update for postgres user

encima:

Fix and feature

Safeupdate enabled for authenticator, not loadable/usable by other roles

## What is the new behavior?

Load safeupdate as a plugin so it can be loaded by any role (not just superuser)
`load $libdir/plugins/safeupdate`

Loaded by default for anon, authenticator, authenticated and postgres.
Enabled by default for anon, authenticator and authenticated

## Additional context

Supersedes #2027 
`session_preload_libraries` is reset for authenticator role and safeupdate is only built into the `plugins` dir, not into the extension dir additionally


samrose (approved):

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
```diff
@@ -0,0 +1,15 @@
+-- migrate:up
+ALTER ROLE anon SET local_preload_libraries = '$libdir/plugins/safeupdate';
+ALTER ROLE authenticator SET local_preload_libraries = '$libdir/plugins/safeupdate';
+ALTER ROLE authenticated SET local_preload_libraries = '$libdir/plugins/safeupdate';
+ALTER ROLE authenticator RESET session_preload_libraries;
```
steve-chavez: Would this nullify `session_preload_libraries`? We would need that to at least include `supautils` considering https://github.com/supabase/postgres/pull/2113/

steve-chavez:
> Load safeupdate as a plugin so it can be loaded by any role (not just superuser)
load $libdir/plugins/safeupdate

Just to summarize, `session_preload_libraries` can only be set by superuser so if some users want to disable `safeupdate` they wouldn't be able to.. while if we set `safeupdate` on `local_preload_libraries` they would be able to disable right? Is that the main motivation?

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
```diff
@@ -0,0 +1,15 @@
+-- migrate:up
+ALTER ROLE anon SET local_preload_libraries = '$libdir/plugins/safeupdate';
+ALTER ROLE authenticator SET local_preload_libraries = '$libdir/plugins/safeupdate';
+ALTER ROLE authenticated SET local_preload_libraries = '$libdir/plugins/safeupdate';
```
steve-chavez: It doesn't make sense to add `local_preload_libraries` to `anon` or `authenticated` since they never start a connection, they're only impersonated roles (only active through a `SET LOCAL <role>`). As [pg docs say](https://postgresqlco.nf/doc/en/param/local_preload_libraries/) this setting only works at "connection start".

```suggestion
ALTER ROLE authenticator SET local_preload_libraries = '$libdir/plugins/safeupdate';
```

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
```diff
@@ -0,0 +1,15 @@
...
+ALTER ROLE authenticator RESET session_preload_libraries;
+ALTER ROLE postgres SET local_preload_libraries = '$libdir/plugins/safeupdate';
+
+ALTER ROLE anon SET safeupdate.enabled = 1;
+ALTER ROLE authenticator SET safeupdate.enabled = 1;
+ALTER ROLE authenticated SET safeupdate.enabled = 1;
+ALTER ROLE postgres SET safeupdate.enabled = 0;
```
steve-chavez: Ditto
```suggestion
ALTER ROLE authenticator SET safeupdate.enabled = 1;
ALTER ROLE postgres SET safeupdate.enabled = 0;
```

steve-chavez (changes requested):
From a quick manual test it seems safeupdate has been working well for Data API/PostgREST, `anon` and `authenticated` have it enabled. This is thanks to `authenticator` having it in `session_preload_libraries`.

I added some feedback mentioning it's unnecessary and in fact a no-op adding `local_preload_libraries` to `anon`, `authenticated`.

Now, I should ask if the motivation of this PR is to have the `postgres` role able to use the `safeupdate` so other pg clients can benefit? **Edit**: Re-reading the original issue https://github.com/supabase/postgres/issues/1308, indeed looks this was the main motivation. 

If so, we should focus on modifying the `postgres` role and keep the `authenticator` settings as is.

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
encima: We can also close this PR and roll the changes into the one linked?

encima:
> From a quick manual test it seems safeupdate has been working well for Data API/PostgREST, `anon` and `authenticated` have it enabled. This is thanks to `authenticator` having it in `session_preload_libraries`.
> 
> I added some feedback mentioning it's unnecessary and in fact a no-op adding `local_preload_libraries` to `anon`, `authenticated`.
> 
> Now, I should ask if the motivation of this PR is to have the `postgres` role able to use the `safeupdate` so other pg clients can benefit? **Edit**: Re-reading the original issue #1308, indeed looks this was the main motivation.
> 
> If so, we should focus on modifying the `postgres` role and keep the `authenticator` settings as is.

@steve-chavez Can you provide your test case please?

Running this on `15.14.1.084` does not show it as working for the `authenticated` role: 
image

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
encima: good point, thanks, will modify these to `session_preload_libraries`

steve-chavez:
@encima No, it won't work like that. `{session_preload/local_preload}_libraries` only kicks in at login time, note that all the Data API roles have `NOLOGIN`:

https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/migrations/db/init-scripts/00000000000000-initial-schema.sql#L29-L31

The only role that has `LOGIN` for Data API is `authenticator` (this is implied when doing `CREATE USER` instead of `CREATE ROLE`)

https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/migrations/db/init-scripts/00000000000000-initial-schema.sql#L33

So in SQL if you LOGIN as `authenticator` and then do the `SET ROLE anon` then you'd get the `safeupdate` feature. This is what PostgREST/Data API does.

---

I only tested this through PostgREST (I don't think the `authenticator` role password can be modified to do it in plain SQL):

```bash
$ curl -X PATCH 'https://<url>.supabase.co/rest/v1/items?select=id' -H "apikey: <redacted>" -H "Content-Type: application/json"
 --data @- <<JSON
{"name": "asdf"}
JSON

{"code":"21000","details":null,"hint":null,"message":"UPDATE requires a WHERE clause"}$ 

$ curl -X DELETE 'https://<url>.supabase.co/rest/v1/items?select=id' -H "apikey: <redacted>"  -H "Content-Type: application/json
" --data @- <<JSON
{"name": "asdf"}
JSON

{"code":"21000","details":null,"hint":null,"message":"DELETE requires a WHERE clause"}$ 
```

encima:
> @encima No, it won't work like that. `{session_preload/local_preload}_libraries` only kicks in at login time, note that all the Data API roles have `NOLOGIN`:
> 
> 
> 
> https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/migrations/db/init-scripts/00000000000000-initial-schema.sql#L29-L31
> 
> 
> 
> The only role that has `LOGIN` for Data API is `authenticator` (this is implied when doing `CREATE USER` instead of `CREATE ROLE`)
> 
> 
> 
> https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/migrations/db/init-scripts/00000000000000-initial-schema.sql#L33
> 
> 
> 
> So in SQL if you LOGIN as `authenticator` and then do the `SET ROLE anon` then you'd get the `safeupdate` feature. This is what PostgREST/Data API does.
> 
> 
> 
> ---
> 
> 
> 
> I only tested this through PostgREST (I don't think the `authenticator` role password can be modified to do it in plain SQL):
> 
> 
> 
> ```bash
> 
> $ curl -X PATCH 'https://<url>.supabase.co/rest/v1/items?select=id' -H "apikey: <redacted>" -H "Content-Type: application/json"
> 
>  --data @- <<JSON
> 
> {"name": "asdf"}
> 
> JSON
> 
> 
> 
> {"code":"21000","details":null,"hint":null,"message":"UPDATE requires a WHERE clause"}$ 
> 
> 
> 
> $ curl -X DELETE 'https://<url>.supabase.co/rest/v1/items?select=id' -H "apikey: <redacted>"  -H "Content-Type: application/json
> 
> " --data @- <<JSON
> 
> {"name": "asdf"}
> 
> JSON
> 
> 
> 
> {"code":"21000","details":null,"hint":null,"message":"DELETE requires a WHERE clause"}$ 
> 
> ```

Got it, thanks. OK, so the aim of the PR is only to fix postgres not using safeupdate. If that's the case, we don't need to do any breaking changes and can just merge after I modify.

steve-chavez:
Correct, it should be a pure addition without breaking changes.

steve-chavez:
Maybe a new PR called `Fix: allow safeupdate for postgres role`, to avoid confusion.

encima:
> Maybe a new PR called `Fix: allow safeupdate for postgres role`, to avoid confusion.

Renamed this PR to match that. 🙏

migrations/db/migrations/20260403172611_safeupdate-data-api-enable.sql
```diff
@@ -0,0 +1,12 @@
+-- migrate:up
+ALTER ROLE authenticator SET session_preload_libraries = 'supautils, $libdir/plugins/safeupdate';
+ALTER ROLE postgres SET local_preload_libraries = '$libdir/plugins/safeupdate';
+
+ALTER ROLE anon SET safeupdate.enabled = 1;
+ALTER ROLE authenticator SET safeupdate.enabled = 1;
+ALTER ROLE authenticated SET safeupdate.enabled = 1;
```
steve-chavez: Let's not modify anything about Data API roles to ensure no breakage
```suggestion
ALTER ROLE postgres SET local_preload_libraries = '$libdir/plugins/safeupdate';
```

encima: The library will need to be changed for `authenticator` as the output path is different but can remove the rest, yes

nix/ext/pg-safeupdate.nix
```diff
@@ -29,9 +29,9 @@ let
         runHook preInstall
 
         mkdir -p $out/share/postgresql/extension
-
+        mkdir -p $out/lib/plugins
         # Install versioned library
-        install -Dm755 ${pname}${postgresql.dlSuffix} $out/lib/${pname}-${version}${postgresql.dlSuffix}
+        install -Dm755 ${pname}${postgresql.dlSuffix} $out/lib/plugins/${pname}-${version}${postgresql.dlSuffix}
```
steve-chavez, line 34: Why not both paths? That way we don't do any changes on `authenticator`. cc @samrose

nix/ext/pg-safeupdate.nix
```diff
@@ -87,7 +87,7 @@ pkgs.buildEnv {
     numberOfVersions = numberOfVersionsBuilt;
     inherit pname latestOnly;
     defaultSettings = {
-      shared_preload_libraries = [ "safeupdate" ];
+      local_preload_libraries = [ "safeupdate" ];
```
steve-chavez, line 92: Same here, we offer both paths so `safeupdate` can be added on `shared/session_preload_libraries` or `local_preload_libraries`.

nix/ext/pg-safeupdate.nix
encima: Yeah we can do but then we are polluting the output with duplicate binaries. It feels like less maintenance burden to have one library give one output but that's just my view, happy to go with either

steve-chavez: Maybe it's possible to symlink it instead of copying it?

steve-chavez, line 35: Any reason why this cannot stay on the previous path?

Symlinks are looking good, and this path might have the same end result but to reduce risk of breakage we should avoid any modification and only do additions.

encima: Loading is restricted to that dir by PG (see https://postgresqlco.nf/doc/en/param/local_preload_libraries/)

steve-chavez: Great, thanks for the explanation, it'd be good to have this as comment:

```suggestion
        # we use the plugins path because loading libraries with `local_preload_libraries` is restricted to this path only, see https://postgresqlco.nf/doc/en/param/local_preload_libraries/
        install -Dm755 ${pname}${postgresql.dlSuffix} $out/lib/plugins/${pname}-${version}${postgresql.dlSuffix}
```

steve-chavez (approved):

samrose:
@encima what should we do next with this PR?

encima:
> @encima what should we do next with this PR?

@samrose merge ideally please. no platform or other changes needed as this is purely a DB op. Unless you need anything else from me?

PR #2099, opened 2026-03-30
feat: additional config for pgbackrest

Crispy1975:
## What kind of change does this PR introduce?
Additional config for `pgBackRest` backups and restores. Details further in: https://linear.app/supabase/issue/INDATA-403/admin-agent-add-new-command-to-control-pgbackrest

PR #2114, opened 2026-04-15
fix: reload needs signal

staaldraad:
`systemctl reload postgresql.service`  requires signal in the apparmor profile (`/bin/kill` is not sufficient). 
adds `/usr/bin/kill` for redundancy.


samrose (approved):

staaldraad:
addressed in https://github.com/supabase/postgres/pull/2099

PR #2119, opened 2026-04-21
feat: turn pg_graphql off by default for new projects

samrose:
## Summary

Ships pg_graphql disabled by default on newly-provisioned Postgres instances. Existing projects are unaffected because the new migrations only run on first boot (sentinel-guarded).

## Changes

**New migrations**
- `20260421000000_pg_graphql-off-by-default.sql`: drops `pg_graphql` as the final migration step so new DBs start with the extension off. Down migration recreates it if available.
- `20260421000001_rescope_pg_graphql_access_trigger.sql`: fixes a pre-existing latent bug in the `grant_pg_graphql_access` event trigger. Previously it fired on `CREATE FUNCTION` and ran body logic for any function named `resolve`, causing failures like `ERROR: must be owner of function graphql_public.graphql` for unrelated user code. The trigger now fires on `CREATE EXTENSION` and early-exits unless the extension being created is `pg_graphql`.

**Wrapper attachment fix (inside the rescoped trigger)**
Because `ddl_command_end` fires after extension creation completes, the rebuilt `graphql_public.graphql(...)` wrapper is not in `creating_extension` context and is not an extension member. Without an explicit attach, `DROP EXTENSION` would not cascade to the wrapper and the `set_graphql_placeholder` trigger would not reinstall the "not enabled" stub. Added `ALTER EXTENSION pg_graphql ADD FUNCTION graphql_public.graphql(...)` inside the trigger body to make drop cascades work correctly.

**Schema dumps regenerated**
- `migrations/schema-{15,17,orioledb-17}.sql`: removed `CREATE EXTENSION pg_graphql`, added the standalone placeholder `graphql_public.graphql()`, updated `grant_pg_graphql_access` body, flipped event trigger WHEN TAG from `CREATE FUNCTION` to `CREATE EXTENSION`.

**Test expectations updated**
- `nix/tests/expected/*evtrigs.out`: row reordering only. The event trigger is recreated by the new migration, giving it a new OID and shifting row order in the output. No semantic changes.

**Version bumps**
- `ansible/vars.yml`: postgres releases tagged with `-gql-off` suffix.

## Behavioral verification

Tested locally via psql:
1. Fresh DB boot without pg_graphql, wrapper is the "not enabled" placeholder
2. `CREATE EXTENSION pg_graphql` installs the real wrapper (plpgsql, prolang 13)
3. `DROP EXTENSION pg_graphql` reinstates the placeholder (sql, prolang 14)
4. Re-enable path works
5. Landmine test: `CREATE FUNCTION myapp.resolve(...)` as non-superuser no longer triggers `grant_pg_graphql_access` and does not error

## Scope of impact

- New projects: pg_graphql is off by default; customers who want it run `CREATE EXTENSION pg_graphql`
- Existing projects: unaffected. The infrastructure `run_migrations()` call in `/infrastructure/init-scripts/project/00-init.sh` is sentinel-guarded, so these migrations run only at first boot

## Test plan

- [x] Docker image tests pass (`nix run .#docker-image-test`)
- [x] pg_regress suite passes for all three flavors (pg15, pg17, orioledb-17)
- [x] Smoke test a newly-provisioned project: verify `pg_graphql` is absent until explicitly enabled
- [x] Verify `CREATE EXTENSION pg_graphql` still installs cleanly on a fresh instance
- [x] Verify `DROP EXTENSION pg_graphql` reinstates the placeholder function
- [x] Verify creating a user function named `resolve` no longer triggers the graphql wrapper replacement path

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -330,7 +330,7 @@ jobs:
           ./bin/psql -p 54322 -h 127.0.0.1 -U supabase_admin -d postgres -v ON_ERROR_STOP=1 -c "\dx" | tee extensions.log
 
           # Check for required extensions
-          for ext in pg_graphql pgcrypto uuid-ossp supabase_vault; do
+          for ext in pgcrypto uuid-ossp supabase_vault; do
             if ! grep -q "$ext" extensions.log; then
               echo "Required extension $ext not found"
               exit 1
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.067-orioledb"
-  postgres17: "17.6.1.110"
-  postgres15: "15.14.1.110"
+  postgresorioledb-17: "17.6.0.065-orioledb-gql-off"
+  postgres17: "17.6.1.108-gql-off"
+  postgres15: "15.14.1.108-gql-off"
 
 # Non Postgres Extensions
 pgbouncer_release: 1.25.1
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.065-orioledb-gql-off"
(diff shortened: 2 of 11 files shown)


pashkinelfe (approved):

jeanlucaslima:
But why?

PR #2122, opened 2026-04-22
fix: do not fail if data disk is not mounted

pcnc:
INFRA-2115

 Two functional changes                                                                                                                                                                                                                      
                                                                                                                                                                                                                                              
  1. Make /data mount non-fatal at boot  in the generated /etc/fstab, the /data entry gains nofail,x-systemd.device-timeout=5s. If the data EBS volume is missing or slow to appear, systemd won't drop the instance into emergency mode; it 
  waits 5 s, then proceeds.  
  2. Pin debootstrap to the regional EC2 Ubuntu mirror  surrogate-bootstrap-nix.sh now detects REGION before the bootstrap and passes http://${REGION}.cloud.ports.ubuntu.com/ubuntu-ports to debootstrap, instead of letting it default to the
   global ports.ubuntu.com. Avoids AMI-build timeouts caused by the global mirror stalling.

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.066-disk-no-fail-sam-2"
-  postgres17: "17.6.1.109-disk-no-fail-sam-2"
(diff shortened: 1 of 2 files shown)


samrose (approved):
@pcnc approved to unblock, trusting you'll thoroughly test before release

PR #2128, opened 2026-04-29
feat: x86 support GENCOMP-44

darora:
Adds support for building x86 AMIs. 
Also updates the AMI build process to be faster and a little more resilient to package download timeouts etc.

### DIFF
--- a/amazon-arm64-nix.pkr.hcl
+++ b/amazon-arm64-nix.pkr.hcl
@@ -177,11 +177,11 @@ source "amazon-ebssurrogate" "source" {
   ssh_timeout = "5m"
 
   ami_root_device {
-    source_device_name = "/dev/xvdf"
-    device_name = "/dev/xvda"
+    source_device_name    = "/dev/xvdf"
+    device_name           = "/dev/xvda"
     delete_on_termination = true
-    volume_size = 10
-    volume_type = "gp3"
+    volume_size           = 10
+    volume_type           = "gp3"
   }
 
   associate_public_ip_address = true
--- a/amazon-arm64-nix.pkr.hcl
+++ b/amazon-arm64-nix.pkr.hcl
@@ -122,28 +122,32 @@ source "amazon-ebssurrogate" "source" {
 
   ena_support = true
   launch_block_device_mappings {
-    device_name = "/dev/xvdf"
+    device_name           = "/dev/xvdf"
     delete_on_termination = true
-    volume_size = 10
-    volume_type = "gp3"
+    volume_size           = 20
+    volume_type           = "gp3"
+    iops                  = 10000
+    throughput            = 1000
    }
 
   # NOTE: /dev/xvdh is mounted as /data (PostgreSQL data/WAL). The 1 GiB size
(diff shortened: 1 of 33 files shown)


nix/packages/build-ami.nix
```diff
@@ -19,6 +19,7 @@ let
         (root + "/ansible")
         (root + "/migrations")
         (root + "/scripts")
+        (root + "/amazon-amd64-nix.pkr.hcl")
```
darora, line 22: this is possibly insufficient to fix the "single release" workflow. But not spending more time in nix weeds for now.

ansible/tasks/internal/admin-mgr.yml
```diff
@@ -13,6 +13,10 @@
     url: "https://supabase-public-artifacts-bucket.s3.amazonaws.com/admin-mgr/v{{ adminmgr_release }}/admin-mgr_{{ adminmgr_release }}_linux_{{ arch }}.tar.gz"
     dest: "/tmp/admin-mgr.tar.gz"
     timeout: 90
+  register: download_result
```
darora, line 16: Adding these to avoid failing the entire build when a single network request flakes

amazon-arm64-nix.pkr.hcl
```diff
@@ -122,28 +122,32 @@ source "amazon-ebssurrogate" "source" {
...
   ena_support = true
   launch_block_device_mappings {
-    device_name = "/dev/xvdf"
+    device_name           = "/dev/xvdf"
     delete_on_termination = true
-    volume_size = 10
-    volume_type = "gp3"
+    volume_size           = 20
+    volume_type           = "gp3"
+    iops                  = 10000
```
darora, line 129: stage1 builds after this are in the range of 14-17mins, were generally 24-45 prior. Let's assume 24, since the variance is likely being improved by the wgetrc change

ansible/tasks/internal/setup-ansible-pull.yml
```diff
@@ -2,7 +2,8 @@
   shell:
     cmd: |
       apt install -y software-properties-common
-      add-apt-repository --yes --update ppa:ansible/ansible
+      # TODO (darora): temporarily disabling while Launchpad is under ddos attack and very frequently timing out
```
samrose, line 5: These kinds of things have been randomy happening for months across these ubuntu package repo dependencies. Ansible can be installed from nixpkgs and for things like this we may switch to doing so. We can then also mirror in our own binary cache.

samrose (approved):

sweatybridge (approved):

ansible/tasks/internal/setup-ansible-pull.yml
darora: I've seen nix installations failing on timeouts from github (possible other sources) as well? Feels like we need our own cache to really resolve it, whether that's for ubuntu packages or for nix things

samrose: We do have our own cache for nix, and an issue linear to cache upstream items from our flake lock. Ubuntu seems harder but there may be solutions

PR #2130, opened 2026-04-30
fix: lower default max_slot_wal_keep_size from 4096MB to 512MB

Crispy1975:
## Summary

- `max_slot_wal_keep_size` was set to `4096MB`, exceeding the total disk size of free tier projects (2GB) — a lagging or disconnected replication slot could fill the disk entirely before Postgres invalidates it
- Lowered to `512MB`, which fits safely within free tier disk overhead
- Free tier projects cannot attach physical replicas, but logical replication slots are still possible (e.g. Supabase Realtime, logical replication, CDC tools) and this setting applies to all slot types
- WAL archiving to S3 via `archive_command` means local WAL retention is not required for recovery, so a lower cap has no impact on PITR

## Test plan

- [x] Verify Postgres starts correctly with the new default on an AMI build
- [x] Confirm Supabase Realtime slot is not affected under normal operation
- [x] Confirm a lagging slot is invalidated at ~512MB rather than ~4GB
- [x] Consider whether this should be tier-specific in a follow-up

### DIFF
--- a/ansible/files/postgresql_config/postgresql.conf.j2
+++ b/ansible/files/postgresql_config/postgresql.conf.j2
@@ -299,7 +299,7 @@ max_wal_senders = 10		# max number of walsender processes
 max_replication_slots = 5	# max number of replication slots
 				# (change requires restart)
 #wal_keep_size = 0		# in megabytes; 0 disables
-max_slot_wal_keep_size = 4096   # in megabytes; -1 disables
+max_slot_wal_keep_size = 512    # in megabytes; -1 disables
 #wal_sender_timeout = 60s	# in milliseconds; 0 disables
 #track_commit_timestamp = off	# collect timestamp of transaction commit
 				# (change requires restart)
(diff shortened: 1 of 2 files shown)


hunleyd (commented):
LGTM

samrose (approved):

hitmands (commented):
this change isn't scoped to the free tier only, and as per my comment here <https://github.com/supabase/supabase-admin-api/pull/316#pullrequestreview-4210781796>, I think this affects how  the replication streaming works.

hitmands (approved):

Crispy1975:
> this change isn't scoped to the free tier only, and as per my comment here [supabase/supabase-admin-api#316 (review)](https://github.com/supabase/supabase-admin-api/pull/316#pullrequestreview-4210781796), I think this affects how the replication streaming works.

The AMI is a base default, but `admin-api` makes the per compute changes.

samrose (approved):

PR #2145, opened 2026-05-07
fix: bump admin mgr to fix a bug

Crispy1975:
Bump `admin-mgr` to fix a type issue.


tgallacher (approved):
can't remember OTOH, but the referenced AMIs have already been released?

PR #2146, opened 2026-05-07
pg_graphql: Avoid mass-rebuild on new latest version

mmlb:

- Avoid mass rebuild of pg_graphql when a new latest is added


mmlb:
Was working on adding pg_graphql v1.6.1 and noticed that when I did so the old/previously built extensions were also being built. That didn't really make sense to me so I investigated. First attempt was going to keep the `if $latestVersion` but in nix instead of bash so only the previousLatest and newLatest would be built but claude suggested do this instead which is much nicer indeed.

samrose (approved):
Looks like this is actually structurally the correct place to put this. 

Moves the "which version is the default" decision out of the per-version derivation (where it shouldn't have been  old versions don't know or care which is the latest) and into the outer
   buildEnv that aggregates them, so per-version hashes become a pure function of that version's own inputs.


thank you nice work.

PR #2150, opened 2026-05-11
fix: prevent non-superuser roles from dropping supabase_privileged_role

utkarash2991:
Add supabase_privileged_role to supautils.reserved_roles so non-superusers with CREATEROLE cannot drop it on PG 15, where PG's native ADMIN OPTION check (introduced in PG 16) does not apply.

PG 16+ requires ADMIN OPTION to drop a non-superuser role: https://github.com/postgres/postgres/blame/REL_16_STABLE/src/backend/commands/user.c#L1175

This check is absent in PG 15:
https://github.com/postgres/postgres/blob/REL_15_STABLE/src/backend/commands/user.c#L986

Fixes: https://linear.app/supabase/issue/PSQL-1205


nix/tests/sql/supautils_reserved_roles.sql
```diff
@@ -0,0 +1,16 @@
+-- verify non-superuser with CREATEROLE cannot drop supabase_privileged_role
+BEGIN;
+
+CREATE ROLE test_reserved_attacker CREATEROLE;
+GRANT test_reserved_attacker TO postgres;
+SET ROLE test_reserved_attacker;
```
za-arthur: I suggest to test `postgres` role which is non-superuser on tests.
```suggestion
-- Switch to the postgres role (non-superuser) to test supautils behavior
SET ROLE postgres;
```

nix/tests/sql/supautils_reserved_roles.sql
```diff
@@ -0,0 +1,16 @@
...
+
+SAVEPOINT before_drop;
```
za-arthur, line 8: Is SAVEPOINT here necessary?

utkarash2991: The SAVEPOINT lets us rollback to before the error, clearing the aborted state, so the transaction can continue and we can confirm the role was not dropped.

nix/tests/sql/supautils_reserved_roles.sql
utkarash2991: sure . Taking care in next revision.

soedirgo (approved):

za-arthur (approved):

PR #2151, opened 2026-05-13
fix: upload upgrade scripts for x86 instances

darora:
x86 instances expect their artifacts to be at a slightly different
location. While most of their are already uploaded, the upgrade
scripts are currently not.

Towards GENCOMP-69


soedirgo (approved):

darora:
On second look, perhaps the cleaner way to address this is to make edits at https://github.com/supabase/postgres/blob/develop/.github/workflows/ami-release-nix.yml#L290 where we're currently only triggering the workflow for the arm architecture

soedirgo:
ISTM we need to update both files, plus `publish-nix-pgupgrade-bin-flake-version.yml`

PR #2156, opened 2026-05-15
ci: add OCI smoke gate workflow

dliappis:

CI — adds a new GH Actions workflow. Not wired as a required check; just runs and reports.

There's no fast pre-EC2 validation for PRs touching the AMI build. `testinfra-ami-build.yml` is the only path, and it does the full Packer + EC2 round-trip (~20-40 min) before any service-level behavior is exercised.

## What is the new behavior?

A new workflow `.github/workflows/oci-smoke-gate.yml` that runs on PRs touching `ansible/`, `nix/`, `migrations/`, `flake.{nix,lock}`, `Dockerfile-*`, or itself (plus `workflow_dispatch` and `merge_group`).

It builds the AMI as an OCI image via `supabase/supabox`'s `support/ami/Dockerfile`, brings the supabox platform stack up, and runs `dctest test/supadev-smoke.yaml`. Diagnostics + container logs (on failure) are uploaded as a 14-day artifact.

This PR **does not** make any other workflow depend on it. It runs alongside `testinfra-ami-build.yml` and reports its own status. Promoting it to a required check is a follow-up once we see how reliable it is across real PRs.

## Additional context

- Local trial: validated on macOS Docker against supabox `a0fe25c` on 2026-05-15 — 59/59 supadev-smoke tests pass, ~5.5 min of dctest after init. CI wall-clock expected ~15-25 min cold cache.
- `SUPABOX_REF` is SHA-pinned, not tracking `main`, so sibling-team changes can't silently break this workflow. Bump deliberately.
- `pause-restore.yaml` coverage is intentionally deferred — upstream supabox needs a YAML parse fix first, and the spec is slower.
- No PG-version matrix yet (pg17 only). Add later.

Relates [RELENG-31](https://linear.app/supabase/issue/RELENG-31/wire-oci-smoke-gate-into-supabasepostgres-ci-before-test-ami-nixpy).

### DIFF
--- a/.github/workflows/oci-smoke-gate.yml
+++ b/.github/workflows/oci-smoke-gate.yml
@@ -1,0 +1,130 @@
+name: OCI smoke gate
+
+# Builds the AMI as an OCI image via supabox's support/ami/Dockerfile,
+# brings up the supabox platform stack, and runs supadev-smoke as a
+# fast pre-flight check before the more expensive testinfra/EC2 path.
+#
+# Intent: catch service-level, migration, and contract regressions in
+# ~5-10 min on a GH runner, so that PRs touching ansible/nix/migrations
+# fail fast and never burn EC2 minutes on broken builds.
+#
+# pause-restore.yaml coverage will be added in a follow-up once the
+# upstream test/pause-restore.yaml YAML parse fix lands.
+#
+# Tracked under RELENG-31.
+
+on:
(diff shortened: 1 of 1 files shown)


dliappis:
Closing this as we'll likely be taking a different approach, to trigger the needed workflow rather than attempt to clone the tools here.

PR #2160, opened 2026-05-20
feat: add Dockerfile-supabase and rework Dockerfile-multigres as layered images

mkindahl:
### Summary

Introduces `Dockerfile-supabase`, a single parameterised Dockerfile that replaces the near-identical `Dockerfile-15` and `Dockerfile-17`. Rewrites `Dockerfile-multigres` from a self-contained Nix-based build into a thin image layered on top of `Dockerfile-supabase`, reducing duplication and making the multigres image easier to keep in sync with the base.

`Dockerfile-15`, `Dockerfile-17`, and `Dockerfile-orioledb-17` are **unchanged** and remain in the repo for the legacy build path.

---

#### `Dockerfile-supabase` (new)

Parameterised replacement for `Dockerfile-15` and `Dockerfile-17`. The three-stage structure (`nix-builder`, `gosu-builder`, `production`) mirrors the existing files. Version-specific differences (stripping `timescaledb`/`plv8`, commenting out `db_user_namespace`) are gated on `[ "$PG_VERSION" -ge 17 ]` at build time.

```bash
docker build -f Dockerfile-supabase -t supabase-postgres:17 .
docker build -f Dockerfile-supabase --build-arg PG_VERSION=15 -t supabase-postgres:15 .
```

#### `Dockerfile-multigres` (rewritten)

Reduced from ~320 lines to ~125. Three stages:

- **`pgbackrest-builder`** — Alpine + Nix daemon, builds pgBackRest from the flake and saves the store path
- **`pgctld-builder`** — compiles `pgctld` from Go source, pinned to the commit that introduced `--pg-initdb-sql-dirs` (MUL-484)
- **`production`** — `FROM ${SUPABASE_IMAGE}`, merges only the pgBackRest nix store closure (not the profile), symlinks the binary, copies the pgctld wrapper, config template, init scripts, and removes inherited wal-g files

The base image is overridable via `--build-arg SUPABASE_IMAGE`. Dockerfile syntax bumped from 1.2 to 1.4 (required for heredoc support in `RUN`).

```bash
docker build -f Dockerfile-supabase -t supabase-postgres:17 .
docker build -f Dockerfile-multigres -t multigres:17 .

# Override base
docker build -f Dockerfile-multigres \
    --build-arg SUPABASE_IMAGE=registry.example.com/supabase-postgres:17 \
    -t multigres:17 .
```

The `variant-17` / `variant-orioledb-17` final stages are replaced by a single `production` stage. The `multigres-orioledb-17` target is removed; a multigres image layered on orioledb is not built in this PR.

#### `docker/pgctld/pgctld-layered.sh` (new wrapper script)

Installed at `/usr/local/bin/pgctld` (raw binary at `/usr/local/bin/pgctld-bin`). Injects `--postgres-config-template` and `--pg-initdb-sql-dirs` on every `pgctld` call so Kubernetes manifests work without extra flags.

#### `docker/pgctld/postgresql.conf.tmpl`

Removed duplicate bare `include` for `supautils.conf` (kept `include_if_exists`).

#### `ansible/vars.yml`

Single source of truth for which images are built and how they are tagged. Two new sections:

- `release_matrix_base` — base Dockerfiles to build (`Dockerfile-supabase` at pg15 and pg17)
- `release_matrix_layered` — layered Dockerfiles (`Dockerfile-multigres`, `Dockerfile-orioledb-17`)

Each entry carries a `release_key` (maps to a full version in `postgres_release`) and an optional `tag_suffix`. Tags are derived entirely at workflow runtime.

#### Release workflows (`dockerhub-release-matrix.yml`, `manual-docker-release.yml`)

Build flow split into two sequential phases:

- `build_base_images` — builds and pushes supabase base images first
- `build_layered_images` — depends on `build_base_images`, passes `SUPABASE_IMAGE=supabase/postgres:<base_tag>_<arch>` as a build arg

`manual-docker-release.yml` gains two new inputs:
- `imageLabel` — extra label appended to the image tag for test/staging builds (e.g. a short SHA), so test images do not overwrite the production tag
- `dockerfiles` — selects `legacy` (Dockerfile-15/17/orioledb-17) or `layered` (Dockerfile-supabase + Dockerfile-multigres) build path

Fixed nushell `combine_results` script to read `matrix.tag` instead of the non-existent `matrix.version` field.

#### `docker-image-test.yml`

- Matrix entries for `Dockerfile-15` / `Dockerfile-17` replaced by `Dockerfile-supabase` entries with `pg_version: "15"/"17"`
- `multigres-17` entry gains `base_dockerfile: Dockerfile-supabase`, triggering an inline base build before the layered build (each CI runner has an isolated Docker daemon)
- `--target production` hardcoded — all Dockerfiles use a `production` final stage

#### `nix/packages/docker-image-test.nix`, `nix/packages/image-size-analyzer.nix`

- Both gain a `--pg-version` flag to pass `--build-arg PG_VERSION` when building `Dockerfile-supabase`
- `docker-image-test.nix`: pgctld init and start checks now treat `already initialized` / `already running` as errors — these states are invalid for a freshly started container and indicate an environment problem
- `image-size-analyzer.nix`: fixed unbound-variable bash error by using `${2:-}` under `set -u`

---

### Testing

CI runs `docker-image-test` against:

- `Dockerfile-supabase` at pg15 and pg17
- `Dockerfile-multigres` (multigres-17), with `Dockerfile-supabase` built inline as the base

`Dockerfile-orioledb-17` tests are unchanged.

Fixes MUL-483

### DIFF
--- a/.github/workflows/manual-docker-release.yml
+++ b/.github/workflows/manual-docker-release.yml
@@ -1,4 +1,4 @@
-name: Manual Docker Artifacts Release 
+name: Manual Docker Artifacts Release
 
 on:
   workflow_dispatch:
@@ -6,7 +6,19 @@ on:
       postgresVersion:
         description: 'Optional. Postgres version to publish against, i.e. 15.1.1.78'
         required: false
- 
+      imageLabel:
+        description: 'Optional. Extra label appended to the image tag, e.g. a short SHA or test suffix (default: no suffix)'
+        required: false
+        default: ''
+      dockerfiles:
+        description: 'Which Dockerfiles to build. "legacy" uses Dockerfile-15/17/orioledb-17; "layered" uses Dockerfile-supabase + Dockerfile-multigres.'
+        required: false
+        default: 'legacy'
+        type: choice
+        options:
+          - legacy
+          - layered
+
 permissions:
   id-token: write
   contents: read
@@ -16,52 +28,46 @@ jobs:
     runs-on: blacksmith-8vcpu-ubuntu-2404
     outputs:
       matrix_config: ${{ steps.set-matrix.outputs.matrix_config }}
+      base_matrix: ${{ steps.set-matrix.outputs.base_matrix }}
+      layered_matrix: ${{ steps.set-matrix.outputs.layered_matrix }}
+      use_layered: ${{ steps.set-matrix.outputs.use_layered }}
     steps:
       - uses: ./.github/actions/nix-install-ephemeral
       - name: Checkout Repo
         uses: supabase/postgres/.github/actions/shared-checkout@HEAD
       - name: Generate build matrix
         id: set-matrix
+        env:
+          DOCKERFILES_MODE: ${{ inputs.dockerfiles || 'legacy' }}
         run: |
-          nix run nixpkgs#nushell -- -c 'let versions = (open ansible/vars.yml | get postgres_major)
-          let base_matrix = ($versions | each { |ver|
-            let version = ($ver | str trim)
-            let dockerfile = $"Dockerfile-($version)"
-            if ($dockerfile | path exists) {
-              {
-                version: $version,
-                dockerfile: $dockerfile,
-                target: "production"
-              }
-            } else {
-              null
-            }
-          } | compact)
+          nix run nixpkgs#nushell -- -c '
+          let releases = (open ansible/vars.yml | get postgres_release)
+          let use_layered = ($env.DOCKERFILES_MODE == "layered")
 
-          # Discover multigres variants by checking for matching targets in Dockerfile-multigres
-          let multigres_matrix = ($versions | each { |ver|
-            let version = ($ver | str trim)
-            let mg_version = $"multigres-($version)"
-            let mg_dockerfile = "Dockerfile-multigres"
-            let mg_target = $"variant-($version)"
-            if ($mg_dockerfile | path exists) and (open --raw $mg_dockerfile | str contains $"AS ($mg_target)") {
-              {
-                version: $mg_version,
-                dockerfile: $mg_dockerfile,
-                target: $mg_target
-              }
-            } else {
-              null
-            }
-          } | compact)
-
-          let matrix = ($base_matrix | append $multigres_matrix)
-
-          let matrix_config = {
-            include: $matrix
+          let base = if $use_layered {
+            (open ansible/vars.yml | get release_matrix_base
+              | each { |e| $e | insert tag ($releases | get $e.release_key) })
+          } else {
+            (open ansible/vars.yml | get release_matrix_legacy
+              | each { |e| $e | insert tag ($releases | get $e.release_key) })
           }
 
-          $"matrix_config=($matrix_config | to json -r)" | save --append $env.GITHUB_OUTPUT'
+          let layered = if $use_layered {
+            (open ansible/vars.yml | get release_matrix_layered
+              | each { |e|
+                  let ver = ($releases | get $e.release_key)
+                  $e | insert tag $"($ver)($e.tag_suffix)" | insert base_tag $ver
+                })
+          } else {
+            []
+          }
+
+          let combined = ($base | append $layered)
+          $"use_layered=($use_layered)" | save --append $env.GITHUB_OUTPUT
+          $"base_matrix=({include: $base} | to json -r)" | save --append $env.GITHUB_OUTPUT
+          $"layered_matrix=({include: $layered} | to json -r)" | save --append $env.GITHUB_OUTPUT
+          $"matrix_config=({include: $combined} | to json -r)" | save --append $env.GITHUB_OUTPUT
+          '
   build:
     needs: prepare
     strategy:
@@ -73,24 +79,21 @@ jobs:
       - name: Checkout Repo
         uses: supabase/postgres/.github/actions/shared-checkout@HEAD
       - uses: ./.github/actions/nix-install-ephemeral
-      - name: Set PostgreSQL version environment variable
-        run: echo "POSTGRES_MAJOR_VERSION=${{ matrix.version }}" >> $GITHUB_ENV
-      
       - id: args
         run: |
           nix run nixpkgs#nushell -- -c '
-            open ansible/vars.yml 
-            | items { |key value| {name: $key, item: $value} } 
-            | where { |it| ($it.item | describe) == "string" } 
-            | each { |it| $"($it.name)=($it.item)" } 
-            | str join "\n" 
+            open ansible/vars.yml
+            | items { |key value| {name: $key, item: $value} }
+            | where { |it| ($it.item | describe) == "string" }
+            | each { |it| $"($it.name)=($it.item)" }
+            | str join "\n"
             | save --append $env.GITHUB_OUTPUT
-          '    
-  build_release_image:
+          '
+  build_base_images:
     needs: [prepare, build]
     strategy:
       matrix:
-        postgres: ${{ fromJson(needs.prepare.outputs.matrix_config).include }}
+        postgres: ${{ fromJson(needs.prepare.outputs.base_matrix).include }}
         arch: [amd64, arm64]
     runs-on: ${{ matrix.arch == 'amd64' && 'blacksmith-8vcpu-ubuntu-2404' || 'large-linux-arm' }}
     timeout-minutes: 180
@@ -106,62 +109,59 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-      - name: Get image tag
-        id: image
-        run: |
-          if [[ "${{ matrix.arch }}" == "arm64" ]]; then
-            pg_version=$(nix run nixpkgs#nushell -- -c '
-              let version = "${{ matrix.postgres.version }}"
-              let is_multigres = ($version | str starts-with "multigres-")
-              let base_version = if $is_multigres { $version | str replace "multigres-" "" } else { $version }
-              let release_key = if ($base_version | str contains "orioledb") {
-                $"postgresorioledb-17"
-              } else {
-                $"postgres($base_version)"
-              }
-              let base_tag = (open ansible/vars.yml | get postgres_release | get $release_key | str trim)
-              let final_version = if "${{ inputs.postgresVersion }}" != "" {
-                "${{ inputs.postgresVersion }}"
-              } else {
-                $base_tag
-              }
-              if $is_multigres { $"($final_version)-multigres" | str trim } else { $final_version | str trim }
-            ')
-            echo "pg_version=supabase/postgres:$pg_version" >> $GITHUB_OUTPUT
-          else
-            pg_version=$(nix run nixpkgs#nushell -- -c '
-              let version = "${{ matrix.postgres.version }}"
-              let is_multigres = ($version | str starts-with "multigres-")
-              let base_version = if $is_multigres { $version | str replace "multigres-" "" } else { $version }
-              let release_key = if ($base_version | str contains "orioledb") {
-                $"postgresorioledb-17"
-              } else {
-                $"postgres($base_version)"
-              }
-              let base_tag = (open ansible/vars.yml | get postgres_release | get $release_key | str trim)
-              let final_version = if "${{ inputs.postgresVersion }}" != "" {
-                "${{ inputs.postgresVersion }}"
-              } else {
-                $base_tag
-              }
-              if $is_multigres { $"($final_version)-multigres" | str trim } else { $final_version | str trim }
-            ')
-            echo "pg_version=supabase/postgres:$pg_version" >> $GITHUB_OUTPUT
-          fi
       - id: build
         uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5.4.0
         with:
           push: true
           build-args: |
             ${{ needs.build.outputs.build_args }}
+            PG_VERSION=${{ matrix.postgres.pg_version }}
           target: ${{ matrix.postgres.target }}
-          tags: ${{ steps.image.outputs.pg_version }}_${{ matrix.arch }}
+          tags: supabase/postgres:${{ inputs.postgresVersion || matrix.postgres.tag }}${{ inputs.imageLabel != '' && format('-{0}', inputs.imageLabel) || '' }}_${{ matrix.arch }}
           platforms: linux/${{ matrix.arch }}
           cache-from: type=gha,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
           cache-to: type=gha,mode=max,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
           file: ${{ matrix.postgres.dockerfile }}
+
+  build_layered_images:
+    if: needs.prepare.outputs.use_layered == 'true'
+    needs: [prepare, build, build_base_images]
+    strategy:
+      matrix:
+        postgres: ${{ fromJson(needs.prepare.outputs.layered_matrix).include }}
+        arch: [amd64, arm64]
+    runs-on: ${{ matrix.arch == 'amd64' && 'blacksmith-8vcpu-ubuntu-2404' || 'large-linux-arm' }}
+    timeout-minutes: 180
+    steps:
+      - name: Checkout Repo
+        uses: supabase/postgres/.github/actions/shared-checkout@HEAD
+      - uses: ./.github/actions/nix-install-ephemeral
+      - run: docker context create builders
+      - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+        with:
+          endpoint: builders
+      - uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - id: build
+        uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5.4.0
+        with:
+          push: true
+          build-args: |
+            ${{ needs.build.outputs.build_args }}
+            PG_VERSION=${{ matrix.postgres.pg_version }}
+            SUPABASE_IMAGE=supabase/postgres:${{ inputs.postgresVersion || matrix.postgres.base_tag }}${{ inputs.imageLabel != '' && format('-{0}', inputs.imageLabel) || '' }}_${{ matrix.arch }}
+          target: ${{ matrix.postgres.target }}
+          tags: supabase/postgres:${{ inputs.postgresVersion != '' && format('{0}-multigres', inputs.postgresVersion) || matrix.postgres.tag }}${{ inputs.imageLabel != '' && format('-{0}', inputs.imageLabel) || '' }}_${{ matrix.arch }}
+          platforms: linux/${{ matrix.arch }}
+          cache-from: type=gha,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ github.ref_name }}-latest-${{ matrix.arch }}
+          file: ${{ matrix.postgres.dockerfile }}
+
   merge_manifest:
-    needs: [prepare, build, build_release_image]
+    if: always() && !failure() && !cancelled()
+    needs: [prepare, build, build_base_images, build_layered_images]
     strategy:
       matrix:
         include: ${{ fromJson(needs.prepare.outputs.matrix_config).include }}
@@ -177,20 +177,7 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
       - name: Get image tag
         id: get_version
-        run: |
-          nix run nixpkgs#nushell -- -c '
-            let version = "${{ matrix.version }}"
-            let is_multigres = ($version | str starts-with "multigres-")
-            let base_version = if $is_multigres { $version | str replace "multigres-" "" } else { $version }
-            let release_key = if ($base_version | str contains "orioledb") {
-              $"postgresorioledb-17"
-            } else {
-              $"postgres($base_version)"
-            }
-            let base_tag = (open ansible/vars.yml | get postgres_release | get $release_key | str trim)
-            let pg_version = if $is_multigres { $"($base_tag)-multigres" } else { $base_tag }
-            $"pg_version=supabase/postgres:($pg_version)" | save --append $env.GITHUB_OUTPUT
-          '
+        run: echo "pg_version=supabase/postgres:${{ inputs.postgresVersion != '' && (contains(matrix.tag, '-multigres') && format('{0}-multigres', inputs.postgresVersion) || inputs.postgresVersion) || matrix.tag }}${{ inputs.imageLabel != '' && format('-{0}', inputs.imageLabel) || '' }}" >> $GITHUB_OUTPUT
       - name: Output version
         id: output_version
         run: |
@@ -202,7 +189,7 @@ jobs:
       - name: Upload Results Artifact
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
-          name: merge_results-${{ matrix.version }}
+          name: merge_results-${{ matrix.tag }}
           path: results.txt
           if-no-files-found: warn
       - name: Merge multi-arch manifests
@@ -211,6 +198,7 @@ jobs:
           ${{ steps.get_version.outputs.pg_version }}_amd64 \
           ${{ steps.get_version.outputs.pg_version }}_arm64
   combine_results:
+    if: always() && !failure() && !cancelled()
     needs: [prepare, merge_manifest]
     runs-on: blacksmith-8vcpu-ubuntu-2404
     steps:
@@ -228,12 +216,12 @@ jobs:
           nix run nixpkgs#nushell -- -c '
             # Parse the matrix configuration directly
             let matrix_config = (${{ toJson(needs.prepare.outputs.matrix_config) }} | from json)
-      
+
             # Get versions directly from include array
-            let versions = ($matrix_config.include | get version)
-      
+            let versions = ($matrix_config.include | get tag)
+
             echo "Versions: $versions"
-      
+
             # Convert the versions to a comma-separated string
             let versions_str = ($versions | str join ",")
             $"versions=$versions_str" | save --append $env.GITHUB_ENV
@@ -249,7 +237,7 @@ jobs:
             # Get all results files and process them in one go
             let files = (ls **/results.txt | get name)
             echo $"Found files: ($files)"
-            
+
             let matrix = {
               include: (
                 $files
@@ -257,17 +245,17 @@ jobs:
                 | each { |content| $content | lines }  # Split into lines
                 | flatten                             # Flatten the nested lists
                 | where { |line| $line != "" }        # Filter empty lines
-                | each { |line| 
+                | each { |line|
                     # Extract just the version part after the last colon
                     let version = ($line | parse "supabase/postgres:{version}" | get version.0)
                     {version: $version}
                 }
               )
             }
-            
+
             let json_output = ($matrix | to json -r)  # -r for raw output
             echo $"Debug output: ($json_output)"
-            
+
             $"matrix=($json_output)" | save --append $env.GITHUB_OUTPUT
           '
       - name: Debug Combined Results
@@ -285,5 +273,5 @@ jobs:
         matrix: ${{ fromJson(needs.combine_results.outputs.matrix) }}
       uses: ./.github/workflows/mirror.yml
       with:
-        version: ${{ inputs.postgresVersion != '' && (contains(matrix.version, '-multigres') && format('{0}-multigres', inputs.postgresVersion) || inputs.postgresVersion) || matrix.version }}
+        version: ${{ matrix.version }}
       secrets: inherit
--- a/Dockerfile-multigres
+++ b/Dockerfile-multigres
@@ -1,14 +1,28 @@
-# syntax=docker/dockerfile:1.6
-# Multigres PostgreSQL 17 variants — vanilla and OrioleDB
+# syntax=docker/dockerfile:1.4
+# 1.4: minimum version for heredoc support in RUN (used for nix config below); also covers COPY --chmod (1.2+)
+# Multigres PostgreSQL image — layered on top of the supabase base image.
 #
-# Build targets:
-#   docker build -f Dockerfile-multigres --target variant-17 -t multigres-17 .
-#   docker build -f Dockerfile-multigres --target variant-orioledb-17 -t multigres-orioledb-17 .
+# Put all ARGs up top for better discoverability.
+# Use non-versioned `ARG THE_ARG` in the stage that uses it to get the ref to this one.
+#
+# The supabase image must be built first:
+#   docker build -f Dockerfile-supabase -t supabase-postgres:17 .
+#   docker build -f Dockerfile-multigres -t multigres:17 .
+#
+# Override the base image at build time:
+#   docker build -f Dockerfile-multigres \
+#       --build-arg SUPABASE_IMAGE=registry.example.com/supabase-postgres:17 \
+#       -t multigres:17 .
+
+ARG ALPINE_VERSION=3.23
+ARG GOLANG_VERSION=1.26
+ARG PGCTLD_REV=1e3bad798972600778ee27eb08ab7d34cc8be8e9 # Pinned to the commit that introduced --pg-initdb-sql-dirs (MUL-484)
+ARG SUPABASE_IMAGE=supabase-postgres:17
 
 ####################
-# Stage 0: Nix base — shared Alpine + Nix setup for all builders
+# Stage 1: pgBackRest nix builder
 ####################
-FROM alpine:3.23 AS nix-base
+FROM alpine:${ALPINE_VERSION} AS pgbackrest-builder
 
 RUN apk add --no-cache \
     bash \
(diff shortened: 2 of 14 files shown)


Dockerfile-supabase
```diff
@@ -0,0 +1,219 @@
...
+
+# Set permissions
+RUN chown -R postgres:postgres /usr/lib/postgresql && \
+    chown -R postgres:postgres /usr/share/postgresql
+
+# Setup configs
+COPY --chown=postgres:postgres ansible/files/postgresql_config/postgresql.conf.j2 /etc/postgresql/postgresql.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_hba.conf.j2 /etc/postgresql/pg_hba.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_ident.conf.j2 /etc/postgresql/pg_ident.conf
+COPY --chown=postgres:postgres ansible/files/postgresql_config/conf.d /etc/postgresql-custom/conf.d
```
mkindahl: Since we are doing refactoring anyway, we might want to follow the Debian convention and name this `/etc/postgresql/postgresql.conf.d` and rewrite the `/etc/postgresql/postgresql.conf` apropriately. To deal with backward compatibility, we could add a symbolic link.

mkindahl: Updated this.

docker/pgctld/pgctld-layered.sh
mkindahl, line 1: Previous code sent Postgres output to stdout and this was mixed with pgctld output. 

This patch changes that and now only pgctld output is written to stdout. This means that we need to capture the log output by reading the file or mounting a volume where we can read the data.

I am unsure how logs are collected, so we might either send all output to stdout again (which will mix pgctld and postgres output), or send one of them to stdout (depending on how the log collectors operate) and require the other to be read explicitly.

mmlb:
This looks pretty good to me, I've got some follow ups that I think make sense but better to get this PR in and do so later (mostly minor nits, but I suspect we can drop all the Dockerfiles-* and have just the one multi-stage Dockerfile and use `--target` for multigres vs plain, this way we don't need to worry about remembering to build base before multigres)

mmlb:
@mkindahl I tried restarting the failed CI builds and noticed that the failures are real issues in Dockerfile-supabase (for example syntax should be 1.4 not 1.3 because the heredoc syntax was in 1.3-labs and graduated in 1.4, maybe some other issues too).

I tried fixing it and then I ended up trying out putting everything all into one Dockerfile and fell down the rabbit hole here. It's not too bad IMO. I'm going to drop it into here so you can take a quick look if nothing stands out I can throw it into this PR. This one Dockerfile can be used to build PG15, PG17, PG17-orioledb, PG17-multigres, PG17-orioledb-multigres and there's no issue with running 2 commands for multigres.

Unfortunately I think that means some of your changes need to backed off since we'd want to make use of `--target` again.

<details>
<summary>Dockerfile</summary>

```dockerfile
# syntax=docker/dockerfile:1.4
# 1.4: minimum version for heredoc support in RUN (used for nix config below)
# Supabase PostgreSQL image with Nix extensions — parameterised by major version.
#
# Build (defaults to PostgreSQL 17):
#   docker build --build-arg=PG_VERSION=17 --target=postgres -t supabase-postgres:17 .
#
# Build for a different PostgreSQL version:
#   docker build --build-arg=PG_VERSION=15 --target=postgres -t supabase-postgres:15 .
#
# Build for Multigres:
#   docker build --build-arg=PG_VERSION=17 --target=multigres -t supabase-multigres:17 .

# Put all ARGs up top for better discoverability
# Use non-versioned `ARG THE_ARG` in the layer that uses it to get the ref to this one
ARG ALPINE_VERSION=3.23
ARG GOSU_VERSION=1.19
ARG PGCTLD_REV=1e3bad798972600778ee27eb08ab7d34cc8be8e9 # Pinned to the commit that introduced --pg-initdb-sql-dirs (MUL-484)
ARG PG_VERSION=17
ARG VARIANT

###################################
#         Postgres + Exts         #
###################################
FROM alpine:${ALPINE_VERSION} AS pg
ARG PG_VERSION
ARG VARIANT

RUN <<EOF
case $PG_VERSION:$VARIANT in
 15: | 17:) ;;
 17:orioledb) VARIANT=$VARIANT-;;
 15:orioledb) echo "OrioleDB not supported on PG17" >&2 && exit 1;;
 17:*) echo "Unknown VARIANT ($VARIANT) only orioledb supported" >&2 && exit 1;;
 *) echo "Unknown PG_VERSION ($PG_VERSION) only 15,17 supported" >&2 && exit 1;;
esac
EOF

# Install dependencies for nix installer (coreutils for GNU cp, sudo for installer)
RUN apk add --no-cache \
    bash \
    coreutils \
    curl \
    shadow \
    sudo \
    xz

# Create nix config
RUN cat >/tmp/extra-nix.conf <<EOF
extra-experimental-features = nix-command flakes
extra-substituters = https://nix-postgres-artifacts.s3.amazonaws.com
extra-trusted-public-keys = nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI=
EOF
RUN curl -L https://releases.nixos.org/nix/nix-2.34.6/install | sh -s -- --daemon --no-channel-add --yes --nix-extra-conf-file /tmp/extra-nix.conf
ENV PATH="$PATH:/nix/var/nix/profiles/default/bin"
RUN nix --version

WORKDIR /nixpg
COPY . .

# Build PostgreSQL with extensions
RUN attr=psql_$VARIANT${PG_VERSION}_slim/bin && \
    echo Building $attr && \
    nix profile add path:.#$attr

# Build groonga and copy plugins
RUN nix profile add path:.#supabase-groonga && \
    mkdir -p /tmp/groonga-plugins && \
    cp -r /nix/var/nix/profiles/default/lib/groonga/plugins /tmp/groonga-plugins/

RUN nix store gc && nix store optimise

################################################
#                      Gosu                    #
################################################
FROM golang:1.26-alpine${ALPINE_VERSION} AS gosu
ARG GOSU_VERSION

RUN apk add --no-cache curl git
# Build gosu from source
RUN git clone --depth 1 --branch "${GOSU_VERSION}" https://github.com/tianon/gosu.git /gosu && \
    cd /gosu && \
    CGO_ENABLED=0 go build -ldflags="-s -w" -o /usr/local/bin/gosu . && \
    chmod +x /usr/local/bin/gosu && \
    /usr/local/bin/gosu --version

#########################################
#            Production Image           #
#########################################
FROM alpine:${ALPINE_VERSION} AS postgres
ARG ALPINE_VERSION
ARG PG_VERSION

# Install minimal runtime dependencies
RUN apk add --no-cache \
    bash \
    curl \
    shadow \
    su-exec \
    tzdata \
    musl-locales \
    musl-locales-lang \
    && rm -rf /var/cache/apk/*

# Create postgres user/group
RUN addgroup -S postgres && \
    adduser -S -G postgres -h /var/lib/postgresql -s /bin/bash postgres && \
    addgroup -S wal-g && \
    adduser -S -G wal-g -s /bin/bash wal-g && \
    adduser postgres wal-g

# Copy Nix store and profiles from builder (profile already created by nix profile install)
COPY --from=pg /nix /nix

# Copy groonga plugins
COPY --from=pg /tmp/groonga-plugins/plugins /usr/lib/groonga/plugins

# Copy gosu
COPY --from=gosu /usr/local/bin/gosu /usr/local/bin/gosu

# Setup PostgreSQL directories
RUN mkdir -p /usr/lib/postgresql/bin \
    /usr/lib/postgresql/share/postgresql \
    /usr/share/postgresql \
    /var/lib/postgresql/data \
    /var/run/postgresql \
    && chown -R postgres:postgres /usr/lib/postgresql \
    && chown -R postgres:postgres /var/lib/postgresql \
    && chown -R postgres:postgres /usr/share/postgresql \
    && chown -R postgres:postgres /var/run/postgresql

# Create symbolic links for binaries
RUN for f in /nix/var/nix/profiles/default/bin/*; do \
        ln -sf "$f" /usr/lib/postgresql/bin/ 2>/dev/null || true; \
        ln -sf "$f" /usr/bin/ 2>/dev/null || true; \
    done

# Create symbolic links for PostgreSQL shares
RUN ln -sf /nix/var/nix/profiles/default/share/postgresql/* /usr/lib/postgresql/share/postgresql/ 2>/dev/null || true && \
    ln -sf /nix/var/nix/profiles/default/share/postgresql/* /usr/share/postgresql/ 2>/dev/null || true && \
    ln -sf /usr/lib/postgresql/share/postgresql/timezonesets /usr/share/postgresql/timezonesets 2>/dev/null || true

# Set permissions
RUN chown -R postgres:postgres /usr/lib/postgresql && \
    chown -R postgres:postgres /usr/share/postgresql

# Setup configs
COPY --chown=postgres:postgres ansible/files/postgresql_config/postgresql.conf.j2 /etc/postgresql/postgresql.conf
COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_hba.conf.j2 /etc/postgresql/pg_hba.conf
COPY --chown=postgres:postgres ansible/files/postgresql_config/pg_ident.conf.j2 /etc/postgresql/pg_ident.conf
COPY --chown=postgres:postgres ansible/files/postgresql_config/conf.d /etc/postgresql/postgresql.conf.d
COPY --chown=postgres:postgres ansible/files/postgresql_config/postgresql-stdout-log.conf /etc/postgresql/logging.conf
COPY --chown=postgres:postgres ansible/files/postgresql_config/supautils.conf.j2 /etc/postgresql-custom/supautils.conf
COPY --chown=postgres:postgres ansible/files/postgresql_extension_custom_scripts /etc/postgresql-custom/extension-custom-scripts
COPY --chown=postgres:postgres ansible/files/pgsodium_getkey_urandom.sh.j2 /usr/lib/postgresql/bin/pgsodium_getkey.sh
COPY --chown=postgres:postgres ansible/files/postgresql_config/custom_walg.conf /etc/postgresql-custom/wal-g.conf
COPY --chown=postgres:postgres ansible/files/postgresql_config/custom_read_replica.conf /etc/postgresql-custom/read-replica.conf
COPY --chown=postgres:postgres ansible/files/walg_helper_scripts/wal_fetch.sh /home/postgres/wal_fetch.sh
COPY ansible/files/walg_helper_scripts/wal_change_ownership.sh /root/wal_change_ownership.sh

# Configure PostgreSQL settings
RUN sed -i \
    -e "s|#unix_socket_directories = '/tmp'|unix_socket_directories = '/var/run/postgresql'|g" \
    -e "s|#session_preload_libraries = ''|session_preload_libraries = 'supautils'|g" \
    -e "s|#include = '/etc/postgresql-custom/supautils.conf'|include = '/etc/postgresql-custom/supautils.conf'|g" \
    -e "s|#include = '/etc/postgresql-custom/wal-g.conf'|include = '/etc/postgresql-custom/wal-g.conf'|g" /etc/postgresql/postgresql.conf && \
    echo "pgsodium.getkey_script= '/usr/lib/postgresql/bin/pgsodium_getkey.sh'" >> /etc/postgresql/postgresql.conf && \
    echo "vault.getkey_script= '/usr/lib/postgresql/bin/pgsodium_getkey.sh'" >> /etc/postgresql/postgresql.conf && \
    chown -R postgres:postgres /etc/postgresql-custom && \
    ln -s /etc/postgresql/postgresql.conf.d /etc/postgresql-custom/conf.d

# pg17+ does not ship timescaledb or plv8; db_user_namespace was removed in pg17.
# Applied conditionally so the same Dockerfile works for pg15 (where these are valid).
RUN if [[ "${PG_VERSION#*-}" -ge 17 ]]; then \
        sed -i 's/ timescaledb,//g; s/ plv8,//g' /etc/postgresql/postgresql.conf && \
        sed -i 's/db_user_namespace = off/#db_user_namespace = off/g' /etc/postgresql/postgresql.conf && \
        sed -i 's/ timescaledb,//g; s/ plv8,//g' /etc/postgresql-custom/supautils.conf; \
    fi

# OrioleDB stuff if building it
RUN [[ "$VARIANT-" != orioledb ]] && exit 0; \
    sed -i 's/\(shared_preload_libraries.*\)'\''\(.*\)$/\1, orioledb'\''\2/' "/etc/postgresql/postgresql.conf" && \
    echo "default_table_access_method = 'orioledb'" >> "/etc/postgresql/postgresql.conf" && \
    sed -i 's/ postgis,//g; s/ pgrouting,//g' "/etc/postgresql-custom/supautils.conf" && \
    echo "CREATE EXTENSION orioledb;" > /docker-entrypoint-initdb.d/init-scripts/00-pre-init.sql && \
    chown postgres:postgres /docker-entrypoint-initdb.d/init-scripts/00-pre-init.sql && \
    true

# Include schema migrations
COPY migrations/db /docker-entrypoint-initdb.d/
COPY ansible/files/pgbouncer_config/pgbouncer_auth_schema.sql /docker-entrypoint-initdb.d/init-scripts/00-schema.sql
COPY ansible/files/stat_extension.sql /docker-entrypoint-initdb.d/migrations/00-extension.sql

ADD --chmod=0755 \
    https://raw.githubusercontent.com/docker-library/postgres/6edb0a8c4def40c371514b34aef9037ec82d9110/${PG_VERSION}/alpine${ALPINE_VERSION}/docker-entrypoint.sh \
    /usr/local/bin/docker-entrypoint.sh

# Setup pgsodium key script
RUN mkdir -p /usr/share/postgresql/extension/ && \
    ln -s /usr/lib/postgresql/bin/pgsodium_getkey.sh /usr/share/postgresql/extension/pgsodium_getkey && \
    chmod +x /usr/lib/postgresql/bin/pgsodium_getkey.sh

# Environment variables
ENV PATH="/nix/var/nix/profiles/default/bin:/usr/lib/postgresql/bin:${PATH}"
ENV PGDATA=/var/lib/postgresql/data
ENV POSTGRES_HOST=/var/run/postgresql
ENV POSTGRES_USER=supabase_admin
ENV POSTGRES_DB=postgres
ENV POSTGRES_INITDB_ARGS="--allow-group-access --locale-provider=icu --encoding=UTF-8 --icu-locale=en_US.UTF-8"
ENV LANG=en_US.UTF-8
ENV LANGUAGE=en_US:en
ENV LC_ALL=en_US.UTF-8
ENV GRN_PLUGINS_DIR=/usr/lib/groonga/plugins
# Point to minimal glibc locales included in slim Nix package for initdb locale support
ENV LOCALE_ARCHIVE=/nix/var/nix/profiles/default/lib/locale/locale-archive

# Marks the container unhealthy after 10 failed pg_isready probes, which blocks
# dependent services in Docker Compose (depends_on: condition: service_healthy).
# Kubernetes ignores Docker HEALTHCHECK entirely — use readinessProbe in the Pod spec.
HEALTHCHECK --interval=2s --timeout=2s --retries=10 CMD pg_isready -U postgres -h localhost

# SIGINT triggers PostgreSQL smart shutdown: waits for active sessions to finish
# before stopping. This avoids interrupting in-flight transactions but can delay
# pod termination if long-running queries are active.
# Consider SIGTERM (fast shutdown) to disconnect clients immediately, which
# respects Kubernetes terminationGracePeriodSeconds more predictably.
STOPSIGNAL SIGINT
EXPOSE 5432

# No USER directive: the entrypoint starts as root to fix volume ownership and
# set up permissions, then drops to the postgres user via gosu before exec'ing
# the postgres process. This follows the standard official PostgreSQL image pattern.
ENTRYPOINT ["docker-entrypoint.sh"]
CMD ["postgres", "-D", "/etc/postgresql"]

##################################################
##                     pgctld                   ##
##################################################
FROM golang:1.25-alpine${ALPINE_VERSION} AS pgctld
ARG PGCTLD_REV

RUN apk add --no-cache git

RUN git clone https://github.com/multigres/multigres.git /multigres && \
    cd /multigres && \
    git checkout ${PGCTLD_REV} && \
    # Copy pico CSS assets before build (mirrors pgctld.nix preBuild step)
    cp external/pico/pico.* go/common/web/templates/css/ 2>/dev/null || true && \
    CGO_ENABLED=0 go build -ldflags="-s -w" -o /usr/local/bin/pgctld ./go/cmd/pgctld


##########################
##       MULTIGRES      ##
##########################
FROM postgres AS multigres

RUN if [ "$(postgres --version | awk -F. '{print $1}')" != 'postgres (PostgreSQL) 17' ]; then \
      echo "Multigres only supports PG17, got -->$(postgres --version)<--" >&2; \
      exit 1; \
    fi

# Install pgbackrest (available in Alpine community repo)
RUN apk add --no-cache pgbackrest

# Copy pgctld binary; keep it separate so the wrapper script can reference it cleanly
COPY --from=pgctld /usr/local/bin/pgctld /usr/local/bin/pgctld-bin

# pgctld config template — /etc/pgctld is a mount point in k8s so use a custom dir
COPY docker/pgctld/postgresql.conf.tmpl /etc/pgctld-custom/postgresql.conf.tmpl

# Wrapper: injects --postgres-config-template on every pgctld call so unmodified
# k8s manifests and local provisioner commands work without extra flags
COPY --chmod=0755 docker/pgctld/pgctld /usr/local/bin/pgctld

# /etc/postgresql/postgresql.conf is not modified here: pgctld renders its own
# config from postgresql.conf.tmpl and passes it directly to PostgreSQL, so the
# supabase base config (including wal-g and data_directory settings) is never loaded.

# wal-g is not used in Multigres (pgbackrest handles backups); remove inherited files.
RUN rm -f \
    /etc/postgresql-custom/wal-g.conf \
    /home/postgres/wal_fetch.sh \
    /root/wal_change_ownership.sh

# No HEALTHCHECK defined here: inherits the probe from the supabase base image.
# Kubernetes ignores Docker HEALTHCHECK entirely — use readinessProbe in the Pod spec.
# STOPSIGNAL inherited from supabase base image (SIGINT — smart shutdown).
USER postgres

# pgctld is the cluster lifecycle manager for Multigres: it handles initdb,
# config templating, replication setup, and coordinated restarts. Running it
# as PID 1 ensures it receives stop signals directly and can shut down
# PostgreSQL cleanly before the container exits.
ENTRYPOINT ["/usr/local/bin/pgctld"]

##########################################
##                USELESS               ##
##########################################
FROM scratch
RUN echo "You have tried to build this image without specifying --target" >&2 && \
    echo "Either postgres or multigres is required" >&2 && \
    echo "P.S.: You can specify PG version with --build-arg=PG_VERSION=15|17" >&2 && \
    echo "P.P.S: You can build orioledb varian with --build-arg=VARIANT=orioledb" >&2 && \
    exit 1

```
</details>

<details>

<summary>diff Dockerfile-supabase Dockerfile</summary>

```diff
--- Dockerfile-supabase	2026-05-27 19:26:43.324333422 -0400
+++ Dockerfile	2026-05-27 19:31:23.248698133 -0400
@@ -1,21 +1,43 @@
-# syntax=docker/dockerfile:1.3
-# 1.3: minimum version for heredoc support in RUN (used for nix config below)
+# syntax=docker/dockerfile:1.4
+# 1.4: minimum version for heredoc support in RUN (used for nix config below)
 # Supabase PostgreSQL image with Nix extensions — parameterised by major version.
 #
 # Build (defaults to PostgreSQL 17):
-#   docker build -f Dockerfile-supabase -t supabase-postgres:17 .
+#   docker build --target=postgres -t supabase-postgres:17 .
 #
 # Build for a different PostgreSQL version:
-#   docker build -f Dockerfile-supabase --build-arg PG_VERSION=15 -t supabase-postgres:15 .
+#   docker build --build-arg=PG_VERSION=15 --target=postgres -t supabase-postgres:15 .
+#
+# Build for OrioleDB:
+#   docker build --build-arg=PG_VERSION=17 --build-arg=VARIANT=orioledb --target=postgres -t supabase-orioledb:17 .
+#
+# Build for Multigres:
+#   docker build --build-arg=PG_VERSION=17 --target=multigres -t supabase-multigres:17 .
 
+# Put all ARGs up top for better discoverability
+# Use non-versioned `ARG THE_ARG` in the layer that uses it to get the ref to this one
+ARG ALPINE_VERSION=3.23
+ARG GOSU_VERSION=1.19
+ARG PGCTLD_REV=1e3bad798972600778ee27eb08ab7d34cc8be8e9 # Pinned to the commit that introduced --pg-initdb-sql-dirs (MUL-484)
 ARG PG_VERSION=17
+ARG VARIANT
 
-####################
-# Stage 1: Nix builder
-####################
-FROM alpine:3.23 AS nix-builder
-
+###################################
+#         Postgres + Exts         #
+###################################
+FROM alpine:${ALPINE_VERSION} AS pg
 ARG PG_VERSION
+ARG VARIANT
+
+RUN <<EOF
+case $PG_VERSION:$VARIANT in
+ 15: | 17:) ;;
+ 17:orioledb) VARIANT=$VARIANT-;;
+ 15:orioledb) echo "OrioleDB not supported on PG17" >&2 && exit 1;;
+ 17:*) echo "Unknown VARIANT ($VARIANT) only orioledb supported" >&2 && exit 1;;
+ *) echo "Unknown PG_VERSION ($PG_VERSION) only 15,17 supported" >&2 && exit 1;;
+esac
+EOF
 
 # Install dependencies for nix installer (coreutils for GNU cp, sudo for installer)
 RUN apk add --no-cache \
@@ -26,63 +48,50 @@
     sudo \
     xz
 
-# Create users (Alpine syntax)
-RUN addgroup -S postgres && \
-    adduser -S -h /var/lib/postgresql -s /bin/bash -G postgres postgres && \
-    addgroup -S wal-g && \
-    adduser -S -s /bin/bash -G wal-g wal-g
-
 # Create nix config
-RUN cat <<EOF > /tmp/extra-nix.conf
+RUN cat >/tmp/extra-nix.conf <<EOF
 extra-experimental-features = nix-command flakes
 extra-substituters = https://nix-postgres-artifacts.s3.amazonaws.com
 extra-trusted-public-keys = nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI=
 EOF
 RUN curl -L https://releases.nixos.org/nix/nix-2.34.6/install | sh -s -- --daemon --no-channel-add --yes --nix-extra-conf-file /tmp/extra-nix.conf
-ENV PATH="${PATH}:/nix/var/nix/profiles/default/bin"
+ENV PATH="$PATH:/nix/var/nix/profiles/default/bin"
 RUN nix --version
 
 WORKDIR /nixpg
 COPY . .
 
 # Build PostgreSQL with extensions
-RUN nix profile add path:.#psql_${PG_VERSION}_slim/bin
-
-RUN nix store gc
+RUN attr=psql_$VARIANT${PG_VERSION}_slim/bin && \
+    echo Building $attr && \
+    nix profile add path:.#$attr
 
 # Build groonga and copy plugins
 RUN nix profile add path:.#supabase-groonga && \
     mkdir -p /tmp/groonga-plugins && \
     cp -r /nix/var/nix/profiles/default/lib/groonga/plugins /tmp/groonga-plugins/
 
-RUN nix store gc
+RUN nix store gc && nix store optimise
 
-####################
-# Stage 2: Gosu builder
-####################
-FROM alpine:3.23 AS gosu-builder
-
-ARG TARGETARCH
-ARG GOSU_VERSION=1.19
-ARG GO_VERSION=1.26.1
+################################################
+#                      Gosu                    #
+################################################
+FROM golang:1.26-alpine${ALPINE_VERSION} AS gosu
+ARG GOSU_VERSION
 
 RUN apk add --no-cache curl git
-
-# Install Go
-RUN curl -fsSL "https://go.dev/dl/go${GO_VERSION}.linux-${TARGETARCH}.tar.gz" | tar -C /usr/local -xz
-ENV PATH="/usr/local/go/bin:${PATH}"
-
 # Build gosu from source
 RUN git clone --depth 1 --branch "${GOSU_VERSION}" https://github.com/tianon/gosu.git /gosu && \
     cd /gosu && \
     CGO_ENABLED=0 go build -ldflags="-s -w" -o /usr/local/bin/gosu . && \
-    chmod +x /usr/local/bin/gosu
-
-####################
-# Stage 3: Final production image
-####################
-FROM alpine:3.23 AS production
+    chmod +x /usr/local/bin/gosu && \
+    /usr/local/bin/gosu --version
 
+#########################################
+#            Production Image           #
+#########################################
+FROM alpine:${ALPINE_VERSION} AS postgres
+ARG ALPINE_VERSION
 ARG PG_VERSION
 
 # Install minimal runtime dependencies
@@ -104,13 +113,13 @@
     adduser postgres wal-g
 
 # Copy Nix store and profiles from builder (profile already created by nix profile install)
-COPY --from=nix-builder /nix /nix
+COPY --from=pg /nix /nix
 
 # Copy groonga plugins
-COPY --from=nix-builder /tmp/groonga-plugins/plugins /usr/lib/groonga/plugins
+COPY --from=pg /tmp/groonga-plugins/plugins /usr/lib/groonga/plugins
 
 # Copy gosu
-COPY --from=gosu-builder /usr/local/bin/gosu /usr/local/bin/gosu
+COPY --from=gosu /usr/local/bin/gosu /usr/local/bin/gosu
 
 # Setup PostgreSQL directories
 RUN mkdir -p /usr/lib/postgresql/bin \
@@ -165,20 +174,28 @@
 
 # pg17+ does not ship timescaledb or plv8; db_user_namespace was removed in pg17.
 # Applied conditionally so the same Dockerfile works for pg15 (where these are valid).
-RUN if [ "${PG_VERSION}" -ge 17 ]; then \
+RUN if [ "$PG_VERSION" -ge 17 ]; then \
         sed -i 's/ timescaledb,//g; s/ plv8,//g' /etc/postgresql/postgresql.conf && \
         sed -i 's/db_user_namespace = off/#db_user_namespace = off/g' /etc/postgresql/postgresql.conf && \
         sed -i 's/ timescaledb,//g; s/ plv8,//g' /etc/postgresql-custom/supautils.conf; \
     fi
 
+# OrioleDB stuff if building it
+RUN [[ "$VARIANT-" != orioledb ]] && exit 0; \
+    sed -i 's/\(shared_preload_libraries.*\)'\''\(.*\)$/\1, orioledb'\''\2/' "/etc/postgresql/postgresql.conf" && \
+    echo "default_table_access_method = 'orioledb'" >> "/etc/postgresql/postgresql.conf" && \
+    sed -i 's/ postgis,//g; s/ pgrouting,//g' "/etc/postgresql-custom/supautils.conf" && \
+    echo "CREATE EXTENSION orioledb;" > /docker-entrypoint-initdb.d/init-scripts/00-pre-init.sql && \
+    chown postgres:postgres /docker-entrypoint-initdb.d/init-scripts/00-pre-init.sql && \
+    true
+
 # Include schema migrations
 COPY migrations/db /docker-entrypoint-initdb.d/
 COPY ansible/files/pgbouncer_config/pgbouncer_auth_schema.sql /docker-entrypoint-initdb.d/init-scripts/00-schema.sql
 COPY ansible/files/stat_extension.sql /docker-entrypoint-initdb.d/migrations/00-extension.sql
 
-# Add entrypoint script from the official postgres Docker library (version-matched)
 ADD --chmod=0755 \
-    https://raw.githubusercontent.com/docker-library/postgres/6edb0a8c4def40c371514b34aef9037ec82d9110/${PG_VERSION}/alpine3.23/docker-entrypoint.sh \
+    https://raw.githubusercontent.com/docker-library/postgres/6edb0a8c4def40c371514b34aef9037ec82d9110/${PG_VERSION}/alpine${ALPINE_VERSION}/docker-entrypoint.sh \
     /usr/local/bin/docker-entrypoint.sh
 
 # Setup pgsodium key script
@@ -218,3 +235,73 @@
 # the postgres process. This follows the standard official PostgreSQL image pattern.
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["postgres", "-D", "/etc/postgresql"]
+
+##################################################
+##                     pgctld                   ##
+##################################################
+FROM golang:1.25-alpine${ALPINE_VERSION} AS pgctld
+ARG PGCTLD_REV
+
+RUN apk add --no-cache git
+
+RUN git clone https://github.com/multigres/multigres.git /multigres && \
+    cd /multigres && \
+    git checkout ${PGCTLD_REV} && \
+    # Copy pico CSS assets before build (mirrors pgctld.nix preBuild step)
+    cp external/pico/pico.* go/common/web/templates/css/ 2>/dev/null || true && \
+    CGO_ENABLED=0 go build -ldflags="-s -w" -o /usr/local/bin/pgctld ./go/cmd/pgctld
+
+
+##########################
+##       MULTIGRES      ##
+##########################
+FROM postgres AS multigres
+
+RUN if [ "$(postgres --version | awk -F. '{print $1}')" != 'postgres (PostgreSQL) 17' ]; then \
+      echo "Multigres only supports PG17, got -->$(postgres --version)<--" >&2; \
+      exit 1; \
+    fi
+
+# Install pgbackrest (available in Alpine community repo)
+RUN apk add --no-cache pgbackrest
+
+# Copy pgctld binary; keep it separate so the wrapper script can reference it cleanly
+COPY --from=pgctld /usr/local/bin/pgctld /usr/local/bin/pgctld-bin
+
+# pgctld config template — /etc/pgctld is a mount point in k8s so use a custom dir
+COPY docker/pgctld/postgresql.conf.tmpl /etc/pgctld-custom/postgresql.conf.tmpl
+
+# Wrapper: injects --postgres-config-template on every pgctld call so unmodified
+# k8s manifests and local provisioner commands work without extra flags
+COPY --chmod=0755 docker/pgctld/pgctld /usr/local/bin/pgctld
+
+# /etc/postgresql/postgresql.conf is not modified here: pgctld renders its own
+# config from postgresql.conf.tmpl and passes it directly to PostgreSQL, so the
+# supabase base config (including wal-g and data_directory settings) is never loaded.
+
+# wal-g is not used in Multigres (pgbackrest handles backups); remove inherited files.
+RUN rm -f \
+    /etc/postgresql-custom/wal-g.conf \
+    /home/postgres/wal_fetch.sh \
+    /root/wal_change_ownership.sh
+
+# No HEALTHCHECK defined here: inherits the probe from the supabase base image.
+# Kubernetes ignores Docker HEALTHCHECK entirely — use readinessProbe in the Pod spec.
+# STOPSIGNAL inherited from supabase base image (SIGINT — smart shutdown).
+USER postgres
+
+# pgctld is the cluster lifecycle manager for Multigres: it handles initdb,
+# config templating, replication setup, and coordinated restarts. Running it
+# as PID 1 ensures it receives stop signals directly and can shut down
+# PostgreSQL cleanly before the container exits.
+ENTRYPOINT ["/usr/local/bin/pgctld"]
+
+##########################################
+##                USELESS               ##
+##########################################
+FROM scratch
+RUN echo "You have tried to build this image without specifying --target" >&2 && \
+    echo "Either postgres or multigres is required" >&2 && \
+    echo "P.S.: You can specify PG version with --build-arg=PG_VERSION=15|17" >&2 && \
+    echo "P.P.S: You can build orioledb varian with --build-arg=VARIANT=orioledb" >&2 && \
+    exit 1
```
</details>

mkindahl:
> This looks pretty good to me, I've got some follow ups that I think make sense but better to get this PR in and do so later (mostly minor nits, but I suspect we can drop all the Dockerfiles-* and have just the one multi-stage Dockerfile and use `--target` for multigres vs plain, this way we don't need to worry about remembering to build base before multigres)

Ah. The main reason for separating this into its own Dockerfile was precisely to have a more streamlined workflow. If a single Dockerfile is created for all the different images, it means we have a monolith that is hard to maintain (in my opinion, of course.)

My thinking was that if we have a good "base", then other projects should be able to build on top of that, even if they are not in the same repo, and having a separate Dockerfile for Multigres creates a good example for how this could be done.

mkindahl:
> @mkindahl I tried restarting the failed CI builds and noticed that the failures are real issues in Dockerfile-supabase (for example syntax should be 1.4 not 1.3 because the heredoc syntax was in 1.3-labs and graduated in 1.4, maybe some other issues too).

Thanks for testing this. I'll update the version.

> I tried fixing it and then I ended up trying out putting everything all into one Dockerfile and fell down the rabbit hole here. It's not too bad IMO. I'm going to drop it into here so you can take a quick look if nothing stands out I can throw it into this PR.

Thanks! I'll take a look at it.

> This one Dockerfile can be used to build PG15, PG17, PG17-orioledb, PG17-multigres, PG17-orioledb-multigres and there's no issue with running 2 commands for multigres.
> 
> Unfortunately I think that means some of your changes need to backed off since we'd want to make use of `--target` again.

My main concern was the duplication of code in all the files. Even if we end up with a single Dockerfile with multiple targets, it is a significant improvement.

mmlb:
> > This one Dockerfile can be used to build PG15, PG17, PG17-orioledb, PG17-multigres, PG17-orioledb-multigres and there's no issue with running 2 commands for multigres.
> > Unfortunately I think that means some of your changes need to backed off since we'd want to make use of `--target` again.
> 
> My main concern was the duplication of code in all the files. Even if we end up with a single Dockerfile with multiple targets, it is a significant improvement.

Yeah I'd drop all the other Dockerfiles since they would now be unnecessary. I'd be concerned about the monolithic Dockerfile being hard to maintain but both orioledb and multigres were pretty tiny additions and not as bad as I would have thought. I suspect building the images with nix might not be terrible either (after other cleanups) I'd also expect it to be faster overall since nix's caching is better than docker's.

mkindahl:
> > @mkindahl I tried restarting the failed CI builds and noticed that the failures are real issues in Dockerfile-supabase (for example syntax should be 1.4 not 1.3 because the heredoc syntax was in 1.3-labs and graduated in 1.4, maybe some other issues too).
> 
> Thanks for testing this. I'll update the version.

I updated the version and ensured that all tests pass (I hope, some are still running).

The current changes does *not* include the single Dockerfile that you proposed: I just wanted to ensure that we have covered everything else before we discuss which approach to take.

mmlb:
> > > @mkindahl I tried restarting the failed CI builds and noticed that the failures are real issues in Dockerfile-supabase (for example syntax should be 1.4 not 1.3 because the heredoc syntax was in 1.3-labs and graduated in 1.4, maybe some other issues too).
> > 
> > 
> > Thanks for testing this. I'll update the version.
> 
> I updated the version and ensured that all tests pass (I hope, some are still running).
> 
> The current changes does _not_ include the single Dockerfile that you proposed: I just wanted to ensure that we have covered everything else before we discuss which approach to take.

I like the one dockerfile-for-everything approach because it really isn't too much extra effort. Multigres add 2 stages, pgctl build and final creation. Not too much to separate out into its own file especially considering that orioledb makes a lot of sense to keep in main Dockerfile, so multigres will odd-build-out.

Keeping it all in one file also makes for simpler build in CI IMO. I think we can remove a bunch of the code by having 3 build steps (static or introspecting vars.yml), one each for pg15, pg17, orioledb. Multigres builds took ~30s on my machine on top its base. Thats likely much less time than getting a new CI VM, checking out code, fetching base image... So we'd have a build for each pg, then if version >15 we also build multigres. That should cut down on the matrix generation code and base/layered/suffix stuff too.

<details>
<summary>bench</summary>
Note I clean between each run because otherwise pgctld is cached and the orioledb flavored multigres finishes in 5s :D

```shell
./bench.sh
+ docker system prune -af --volumes
+ docker pull alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+ docker pull golang:1.25-alpine3.23@sha256:8d22e29d960bc50cd025d93d5b7c7d220b1ee9aa7a239b3c8f55a57e987e8d45
+ docker pull golang:1.26-alpine3.23@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d
+ PG_VERSION=15
+ docker build --build-arg=PG_VERSION=15 --target=postgres -t supabase-15 .
+ tee pg-15.log.swp

real    1m41.767s
user    0m0.697s
sys     0m0.411s
+ docker system prune -af --volumes
+ docker pull alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+ docker pull golang:1.25-alpine3.23@sha256:8d22e29d960bc50cd025d93d5b7c7d220b1ee9aa7a239b3c8f55a57e987e8d45
+ docker pull golang:1.26-alpine3.23@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d
+ PG_VERSION=17
+ docker build --build-arg=PG_VERSION=17 --target=postgres -t supabase-17 .
+ tee pg-17.log.swp

real    1m36.863s
user    0m0.686s
sys     0m0.374s
+ docker build --build-arg=PG_VERSION=17 --target=multigres -t supabase-17-multigres .
+ tee pg-17-multigres.log.swp

real    0m20.832s
user    0m0.140s
sys     0m0.139s
+ docker system prune -af --volumes
+ docker pull alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+ docker pull golang:1.25-alpine3.23@sha256:8d22e29d960bc50cd025d93d5b7c7d220b1ee9aa7a239b3c8f55a57e987e8d45
+ docker pull golang:1.26-alpine3.23@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d
+ docker build --build-arg=PG_VERSION=17 --build-arg=VARIANT=orioledb --target=postgres -t supabase-17-orioledb .
+ tee pg-17-orioledb.log.swp

real    1m37.622s
user    0m0.715s
sys     0m0.414s
+ docker build --build-arg=PG_VERSION=17 --build-arg=VARIANT=orioledb --target=multigres -t supabase-17-orioledb-multigres .
+ tee pg-17-orioledb-multigres.log.swp

real    0m22.151s
user    0m0.210s
sys     0m0.129
```
</details>

mmlb:
@mkindahl I decided to try my thoughts off of develop instead of basing on this branch and pushed it to https://github.com/supabase/postgres/compare/push-nlprpslluupn so we have something to compare against better. I've got more of the CI stuff to handle for a better comparison but wanted to get it up and out for :eye: and so you don't think I've ignored this PR :)

mmlb:
Meeting recap: I'd like to investigate moving images to be built by nix instead of docker, which should make it easier to get the checks/tests run automatically. Would benefit from some refactors to the repo first though. No point continuing going down the route I was taking if that comes to pass so lets get this merged as its an improvement anyway. Pending images get pushed by CI and are deemed working.

.github/workflows/manual-docker-release.yml
```diff
@@ -88,7 +92,7 @@ jobs:
             ${{ needs.build.outputs.build_args }}
             PG_VERSION=${{ matrix.postgres.pg_version }}
           target: ${{ matrix.postgres.target }}
-          tags: supabase/postgres:${{ inputs.postgresVersion || matrix.postgres.tag }}_${{ matrix.arch }}
+          tags: supabase/postgres:${{ inputs.postgresVersion || matrix.postgres.tag }}-${{ needs.prepare.outputs.short_sha }}_${{ matrix.arch }}
```
mmlb: Why not add this as a docker label instead? Don't get me wrong I'm not a fan of our current versioning scheme but I'd prefer to tackle them all at once instead of having different ways per output type.

docker/pgctld/pgctld-layered.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 exec /usr/local/bin/pgctld-bin \
-  --postgres-config-template /etc/pgctld-custom/postgresql.conf.tmpl \
-  --pg-initdb-sql-dirs supabase_admin:/etc/pgctld-custom/pre-init \
-  --pg-initdb-sql-dirs postgres:/docker-entrypoint-initdb.d/init-scripts \
-  --pg-initdb-sql-dirs supabase_admin:/docker-entrypoint-initdb.d/migrations \
-  "$@"
+	--postgres-config-template /etc/pgctld-custom/postgresql.conf.tmpl \
```
mmlb, line 3: Could you squash this into the first commit, no sense in having fixup commits instead of just fixing the thing :D

.github/workflows/manual-docker-release.yml
mkindahl: Good point. Intention here is to be able to generate a dedicated image for testing purposes that does not override the existing image. I have traditionally used the output of `git describe` since that provides you with a version and a hash allowing you to distinguish "post-version builds", but having an explicit label with a proper default is just as good.

mkindahl: This modification is done as a separate commit and intention was to drop it before merging.

docker/pgctld/pgctld-layered.sh
mmlb: Nevermind, I just saw that we squash commits to develop from the merge queue

.github/workflows/manual-docker-release.yml
mmlb: oh if you mean to drop it before merge then there's nothing to worry about. I like to be loud about those commits prepending `DNM:` or `DO-NOT-MERGE` to the commit subject as a signal to reviewers (and myself hopefully..., could add a CI check for that too actually)

Dockerfile-multigres
```diff
@@ -1,321 +1,91 @@
...
 
-RUN for f in /nix/var/nix/profiles/default/bin/*; do \
-        ln -sf "$f" /usr/lib/postgresql/bin/ 2>/dev/null || true; \
-        ln -sf "$f" /usr/bin/ 2>/dev/null || true; \
-    done && \
-    ln -sf /nix/var/nix/profiles/default/share/postgresql/* /usr/lib/postgresql/share/postgresql/ 2>/dev/null || true && \
-    ln -sf /nix/var/nix/profiles/default/share/postgresql/* /usr/share/postgresql/ 2>/dev/null || true && \
-    ln -sf /usr/lib/postgresql/share/postgresql/timezonesets /usr/share/postgresql/timezonesets 2>/dev/null || true
+# Install pgbackrest (available in Alpine community repo)
+RUN apk add --no-cache pgbackrest
```
samrose: @mkindahl hi! 

Once we had some discussion about the needs for multigres to keep pgbackrest in sync with the rest of supabase platform, we're suggesting to install it from our build with a command 

`RUN nix profile add path:.#psql_17_slim/bin path:.#pg-backrest`

fwiw: We may build multiple versions and make them available, which could require the command to change later on for instance to adding version number in package name. But for now this is the supported way to do it. Thanks

samrose (dismissed):

Dockerfile-multigres
mkindahl: Thanks @samrose!

The command above would only work if we have a running nix daemon, but there is none in the Multigres image, so I added a separate build stage to build `pgbackrest` for the Multigres image.

An alternative would be to add `pgbackrest` to the supabase image, but there is no such addition yet. If there are plans to add pgbackrest to supabase images as well, we could add it there.

nix/packages/docker-image-test.nix
```diff
@@ -319,20 +334,37 @@ writeShellApplication {
...
-            log_error "  pgctld init failed: $init_out"
+        # 3. /usr/local/bin/pgctld must be a wrapper script (not the raw binary)
+        local first_line
+        first_line=$(docker exec "$container" sh -c "head -1 /usr/local/bin/pgctld")
+        if [[ "$first_line" != "#!/bin/sh" ]]; then
+            log_error "  /usr/local/bin/pgctld is not a wrapper script (first line: $first_line)"
+            exit 1
+        fi
+        log_info "  ✓ /usr/local/bin/pgctld is a wrapper script"
+
```
mmlb, line 345: Curious whats the reasoning behind this check?

nix/packages/docker-image-test.nix
```diff
@@ -319,20 +334,37 @@ writeShellApplication {
...
+        # 4. pgctld init: initializes PGDATA and runs SQL init scripts (via wrapper's
+        #    --pg-initdb-sql-dirs). Uses pooler-dir for socket dir and pgBackRest config.
+        local init_out init_rc=0
+        init_out=$(docker exec \
+            -e POSTGRES_PASSWORD="$POSTGRES_PASSWORD" \
+            "$container" \
+            sh -c "/usr/local/bin/pgctld init --pooler-dir $pooler_dir 2>&1") || init_rc=$?
+        if [[ $init_rc -ne 0 ]] || ! echo "$init_out" | grep -qE "initialized successfully|already initialized"; then
```
mmlb: Why isn't `already initialized` an error? Shouldn't we be expecting an uninitialized db and encountering one means our context/environment is wrong?

nix/packages/docker-image-test.nix
```diff
@@ -319,20 +334,37 @@ writeShellApplication {
...
-        if ! echo "$start_out" | grep -q "started successfully"; then
-            log_error "  pgctld start failed: $start_out"
+        # 5. pgctld start: starts PostgreSQL (pg_ctl start). Postgres will listen on
+        #    port 5432 inside the container and on the pooler-dir unix socket.
+        local start_out start_rc=0
+        start_out=$(docker exec \
+            -e POSTGRES_PASSWORD="$POSTGRES_PASSWORD" \
+            "$container" \
+            sh -c "/usr/local/bin/pgctld start --pooler-dir $pooler_dir 2>&1") || start_rc=$?
+        if [[ $start_rc -ne 0 ]] || ! echo "$start_out" | grep -qE "started successfully|already running"; then
```
mmlb: ditto `already running`

nix/packages/docker-image-test.nix
```diff
@@ -524,11 +535,10 @@ writeShellApplication {
             -p "$PORT:5432" \
             "$IMAGE_TAG"
 
-        # Multigres images use "tail -f /dev/null" as their entrypoint — postgres must be
-        # started manually before we can run tests against them.
+        # Multigres: pgctld server (PID 1) does not auto-start PostgreSQL.
+        # Call pgctld init + start via docker exec before waiting for postgres.
```
mmlb: This line indicates an action we should be taking but its already done by `verify_pgctld_integration` so can be confusing. Suggest reword to something like:

```suggestion
        # Multigres: pgctld server (PID 1) does not auto-start PostgreSQL.
        # verify_pgctld_integration starts Postgres and ensures its listening for us.
```

nix/packages/image-size-analyzer.nix
```diff
@@ -74,6 +77,14 @@ runCommand "image-size-analyzer"
           NO_BUILD=true
           shift
           ;;
+        --pg-version)
+          if [[ -z "$2" ]]; then
```
mmlb:
```suggestion
          if [[ -z "${2:-}" ]]; then
```
Since this script runs with `set -u` we'd get a bash error before `-z` can do the check so the error message won't be shown.

nix/packages/image-size-analyzer.nix
```diff
@@ -141,8 +152,16 @@ runCommand "image-size-analyzer"
     # Get tag name from Dockerfile name
     get_tag() {
       local dockerfile=$1
-      local suffix=''${dockerfile#Dockerfile-}
-      echo "supabase-postgres:$suffix-analyze"
+      if [[ "$dockerfile" == "Dockerfile-supabase" ]]; then
+        if [[ -z "$PG_VERSION" ]]; then
```
mmlb: ditto

Dockerfile-multigres
mmlb: @mkindahl I'm not seeing the pgbackrest install from this repo, did you forget to push?

--- 

nm I see it now. Guess I was an a previous push or something...

nix/packages/docker-image-test.nix
mkindahl: If the raw pgctld binary is called, it will miss some options that are needed for it to work correctly. In particular, `--postgres-config-template` and `--pg-initdb-sql-dirs`, so this is just an extra check that nothing is wrong and the binary was accidentally copied.

nix/packages/image-size-analyzer.nix
mkindahl: Changed.

Dockerfile-multigres
mkindahl: Ok.

nix/packages/docker-image-test.nix
mkindahl: It should be an error. Fixed.

nix/packages/docker-image-test.nix
mkindahl: Same here. Fixed.

.github/workflows/manual-docker-release.yml
mkindahl: Now I keep it instead and check this in the workflow. It defaults to no label, but you can add one when you trigger the workflow.

mkindahl (commented):
Thanks for the review. I fixed this issues and pushed a new version.

nix/packages/image-size-analyzer.nix
mkindahl: Fixed.

mmlb (approved):
LGTM, can't wait to delete this tho ;)

PR #2167, opened 2026-05-26
chore: bump jit gatekeeper version

staaldraad:
Bumps to latest version.

* adds support for scoped PATs

### DIFF
--- a/nix/packages/gatekeeper.nix
+++ b/nix/packages/gatekeeper.nix
@@ -6,12 +6,12 @@ let
 
   upstream-gatekeeper = buildGoModule124 {
     pname = "jit-db-gatekeeper";
-    version = "1.0.2";
+    version = "1.0.5";
     src = pkgs.fetchFromGitHub {
       owner = "supabase";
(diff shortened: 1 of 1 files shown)


brainrake (approved):
Perfect, I had this as a task. The deployed version is already updated to 1.0.5 via other means, but this will be the source of truth soon.

ansible/vars.yml
```diff
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.086-orioledb"
-  postgres17: "17.6.1.129"
-  postgres15: "15.14.1.129"
+  postgresorioledb-17: "17.6.0.087-orioledb"
+  postgres17: "17.6.1.130"
+  postgres15: "15.14.1.130"
```
brainrake: no need to bump version since gatekeeper 1.0.5 is already deployed in production

brainrake (approved):

PR #2174, opened 2026-05-27
Update privileged role allowed configs in supautils.conf.j2

TheOtherBrian1:
Allows users to change the following superuser configs at the role level:

- ~~log_connections~~
- ~~log_disconnections~~
- log_duration
- log_min_error_statement

Exposes log settings that do not interfere with security. Although some of these logs may be expensive, there are imminent plans to force users to pay for log pricing:
- https://linear.app/supabase/issue/DEBUG-124/blog-pricing-for-supabase-logs

As a result, exposing these logs does not pose a long-term cost issue.

Users cannot change these configs via supatutils

## What is the new behavior?

Users can change these configs via supautils

## Additional context

A Linear issue: "[Allow users to change log settings via the MGMT API](https://linear.app/supabase/issue/PSQL-1218/allow-users-to-change-log-settings-via-the-mgmt-api)" seeks to expose certain log configs. It seemed odd not to also expose them at the role level, too, so I added the supautils compatible settings to the list


ansible/files/postgresql_config/supautils.conf.j2
```diff
@@ -10,7 +10,7 @@ supautils.privileged_extensions = 'address_standardizer, address_standardizer_da
...
+supautils.privileged_role_allowed_configs = 'auto_explain.*, deadlock_timeout, log_connections, log_disconnections, log_duration, log_lock_waits, log_min_duration_statement, log_min_error_statement, log_min_messages, log_parameter_max_length, log_replication_commands, log_statement, log_temp_files, pg_net.batch_size, pg_net.ttl, pg_stat_statements.*, pgaudit.log, pgaudit.log_catalog, pgaudit.log_client, pgaudit.log_level, pgaudit.log_relation, pgaudit.log_rows, pgaudit.log_statement, pgaudit.log_statement_once, pgaudit.role, pgrst.*, plan_filter.*, safeupdate.enabled, session_replication_role, track_functions, track_io_timing, wal_compression'
```
soedirgo: I think `log_connections` and `log_disconnections` we need to support in your Mgmt API PR. I tested on my own project and it doesn't work via supautils. It seems only `superuser` GUCs make sense to be in `supautils.privileged_role_allowed_configs`

TheOtherBrian1: I used to think that only sighup+ commands were incompatible with supautils. Apparently `superuser-backend` is incompatible indirectly. It can't be changed in an active session, which functionally means it must be updated directly in the `postgres.conf` file. I'll remove the log_connections settings from the PR

soedirgo (approved):

soedirgo:
Can you update here as well? https://github.com/supabase/salt/blob/808653fd5661796c92a3dec9abb02cd1df72784c/salt/state/formula/supautils/files/supautils.conf#L13

PR #2155, opened 2026-05-14
feat: upgrade 15.16 -> 15.19, 17.6 -> 17.11

samrose:
## What kind of change does this PR introduce?

Bump Postgres 15.14 → 15.19 and 17.6 → 17.11 (upstream security releases of 2026-08-13), absorbing five upstream minors per major — 44 CVEs total, including CVE-2026-19385, CVE-2026-16239, CVE-2026-14669 (all CVSS 8.8).

## Breaking change analysis

Full analysis in [PSQL-1110](https://linear.app/supabase/issue/PSQL-1110)

PR #2175, opened 2026-05-28
fix(tests): split prime.sql so non-superuser harnesses can source it

utkarash2991:
## What

Split `nix/tests/prime.sql` into:
- `prime.sql` — extensions any privileged-extensions-gated session can create. Sourced by all 4 consumers.
- `prime-superuser.sql` (new) — the 7 extensions excluded from supautils' `privileged_extensions` list. Sourced only by superuser-context harnesses.

## Why

`prime.sql` is a dual-use file:

| Consumer | Connects as | Can create "may be unsafe" exts? |
|----------|-------------|----------------------------------|
| pg_regress (`nix/checks.nix`) | `supabase_admin` | yes |
| docker-image-test | `supabase_admin` | yes |
| migrate-tool | superuser (vanilla PG) | yes |
| supadev `test-postgres-engines-with-smoke.sh` | `postgres` (hosted) | **no** |

The 7 supautils-gated extensions (`amcheck, file_fdw, lo, pageinspect, pg_freespacemap, pg_surgery, pg_visibility`) cannot be created by a non-superuser. Pre-[supadev@c6be2cb](https://github.com/supabase/supadev/commit/c6be2cb7cbbc0a8ef6f5569305fa3d4c932a9235) (2026-05-14, fail-on-error by default) the error was silently swallowed; post that change, psql halts at the first failing extension (amcheck, alphabetically first) and engines-with-smoke fails.

First surfaced in the PG 15.18 / 17.10 release smoke run (against [supabase/postgres#2155](https://github.com/supabase/postgres/pull/2155)) — the engines-with-smoke test failed on all 3 engines (PG15, PG17, PG17-oriole) with `permission denied to create extension "amcheck"`:

- Failing run: https://github.com/supabase/supadev/actions/runs/26536859604


nix/tests/prime-superuser.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- These extensions are excluded from `supautils.privileged_extensions` (see
+-- the "omitted because may be unsafe" comment in
+-- `ansible/files/postgresql_config/supautils.conf.j2`). Hosted Supabase
+-- projects cannot install them via non-superuser sessions, so this file is
+-- loaded only by superuser-context harnesses: pg_regress (nix/checks.nix),
+-- the docker-image-test, and the local migrate-tool. supadev's hosted
+-- engines-with-smoke test sources `prime.sql` only.
+--
+-- Keep this list in sync with the "may be unsafe" list in supautils.conf.j2.
```
mmlb: absent an automated way to ensure this, I usually like to mirror this sort of comment in the other file so that *hopefully* when it gets updated someone sees it and will do the right thing here (🤞)

mmlb (changes requested):
lgtm besides the one nit

nix/tests/prime-superuser.sql
utkarash2991: @mmlb Thanks. Taken care in latest revision.

mmlb: well thats pretty terrible now that I look at it. Was hoping for an `exclude` list or something :D. Guess we'll just deal with it for now.

utkarash2991:
@mmlb thought about this — none of the directions I tried felt clean enough for this PR:

- CI validator that diffs the two lists → catches drift but doesn't fix the duplication, just hides it behind a check
- Refactor `supautils.conf.j2` to compute `privileged_extensions` from a structured Jinja dict → meaningful, but it's a production config-rendering change that needs its own design + review
- Codegen `prime-superuser.sql` from the same Jinja source → bigger still

Punting for now. Your "exclude list" instinct lines up most closely with the second option — happy to spin up a follow-up ticket if you'd like to pursue that direction.

mmlb:
> @mmlb thought about this — none of the directions I tried felt clean enough for this PR:
> 
>     * CI validator that diffs the two lists → catches drift but doesn't fix the duplication, just hides it behind a check
This isn't the worse option imo, but not necessary for this PR. I think a small refactor to the .j2 file would make sense too. Create some vars for the various lists then actually use jinja to create the final privileged_extensions. Not quite the next bullet point, simpler version of it. Then the check could be pretty simple too. Again not needed for this PR anyway.
> 
>     * Refactor `supautils.conf.j2` to compute `privileged_extensions` from a structured Jinja dict → meaningful, but it's a production config-rendering change that needs its own design + review
> 
>     * Codegen `prime-superuser.sql` from the same Jinja source → bigger still
> 
> 
> Punting for now. Your "exclude list" instinct lines up most closely with the second option — happy to spin up a follow-up ticket if you'd like to pursue that direction.

nix/tests/prime-superuser.sql
```diff
@@ -0,0 +1,21 @@
...
+
+set client_min_messages = warning;
+
```
mmlb, line 20: looking at j2 file it has:
```
# omitted because may be unsafe:             adminpack, amcheck, file_fdw, lo, old_snapshot, pageinspect, pg_freespacemap, pg_surgery, pg_visibility
```

You're missing `adminpack` and `old_snapshot` here, was that on purpose or didn't notice or something else?

utkarash2991: Both `adminpack` and `old_snapshot` are not part of PG17/contrib, hence the sql doesn't have them. I have updated the comment in `prime-superuser.sql` to document the asymmetry rather than introduce a `DO ... EXCEPTION` workaround. Let me know if you'd prefer the silently-skip version.

utkarash2991: @mmlb 

> I think a small refactor to the .j2 file would make sense too. Create some vars for the various lists then actually use jinja to create the final privileged_extensions. Not quite the next bullet point, simpler version of it. Then the check could be pretty simple too.

Yeah thats cool to do. I have created a Linear task to take that up in next PR: https://linear.app/supabase/issue/PSQL-1265/

mmlb (approved):
Thanks, LGTM

PR #2178, opened 2026-05-29
[codex] Add AMI config precedence smoke test

jgoux:
## What changed

Adds an AMI-level testinfra regression test that verifies CLI-managed PostgreSQL overrides in `/etc/postgresql-custom/custom-overrides.conf` take precedence over generated optimizations in `/etc/postgresql-custom/conf.d/generated-optimizations.conf` after PostgreSQL restarts.

The test writes conflicting `max_connections` values, restarts PostgreSQL, and asserts `pg_settings` reports the value and source file from `custom-overrides.conf`. It backs up and restores the touched config files so later AMI tests are not left with mutated state.

## Why

A hosted AMI incident showed generated optimization config could override CLI-managed Postgres config changes. This smoke test captures that precedence contract at the AMI layer, where the regression occurred, without coupling the test to the CLI or Admin API.

## Validation

- `python3 -m py_compile testinfra/test_ami_nix.py`
- `git diff --check`

I also attempted `nix develop --accept-flake-config -c python -m pytest --collect-only testinfra/test_ami_nix.py`, but the dev shell failed while fetching `crate-cbitset-0.2.0.tar.gz` from crates.io due SSL/403 download errors before pytest collection could run.


samrose (approved):
looks good thank you!

brainrake:
@jgoux this is approved — anything blocking merge, or shall we rebase and land it?

samrose:
set merge when ready just needs a rebase

PR #2181, opened 2026-05-30
refactor(nix/cargo-pgrx): build cargo-pgrx by overriding upstream pkg

Kranzes:
Replace the vendored and outdated cargo-pgrx expression with an override of the upstream nixpkgs package, dropping the manual dependency wiring, custom check flags, and meta block.

Remove the cargo-pgrx overlay and call the package set directly via `callPackages` in the package list. Using the overlay shadows the upstream package, and in addition just adds another layer of unneeded indirection.

### DIFF
--- a/nix/cargo-pgrx/default.nix
+++ b/nix/cargo-pgrx/default.nix
@@ -1,10 +1,8 @@
 {
   lib,
+  cargo-pgrx,
   fetchCrate,
-  openssl,
-  pkg-config,
   makeRustPlatform,
-  stdenv,
   rust-bin,
   rustVersion ? "1.85.1",
 }:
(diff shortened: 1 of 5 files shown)


samrose:
@Kranzes thanks very much for this contribution. We may not use it directly (not because of any problem with it), but we'll definitely consider this direction as we're evaluating this part of our package set right now.

nix/cargo-pgrx/default.nix
```diff
@@ -14,73 +12,56 @@ let
...
-      auditable = false;
-      inherit pname;
-      inherit version;
-      src = fetchCrate { inherit version pname hash; };
-      inherit cargoHash;
-      nativeBuildInputs = lib.optionals stdenv.hostPlatform.isLinux [ pkg-config ];
-      buildInputs = lib.optionals stdenv.hostPlatform.isLinux [ openssl ];
+    args:
+    (cargo-pgrx.override { inherit rustPlatform; }).overrideAttrs rec {
+      pname = if lib.versionOlder version "0.7.4" then "cargo-pgx" else "cargo-pgrx";
```
brainrake, line 17: this might not be needed since oldest version in the collectionis 0.10.2

brainrake:
Good entropy reduction.

PR #2183, opened 2026-06-02
packer: set xtrace for clean up scripts

mmlb:

Simple commit for better CI logs

This was all in service of fixing AMI build failures, for example: https://github.com/supabase/postgres/actions/runs/26757788681/job/78981563652#step:9:6046. Similar to #2177, but looks like its still biting.

## What is the new behavior?

I was trying to debug the issue and wanted to have `set -x` in the clean up scripts but when I tried building the AMI again it all worked. So I want to leave this in since its helpful if/when there's an error in the script again.

### DIFF
--- a/scripts/90-cleanup-qemu.sh
+++ b/scripts/90-cleanup-qemu.sh
@@ -4,7 +4,7 @@
 # © 2021 DigitalOcean LLC.
 # This code is licensed under Apache 2.0 license (see LICENSE.md for details)
 
-set -o errexit
+set -ex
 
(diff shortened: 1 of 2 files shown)


mmlb:
Best reviewed each-commit-at-a-time.

mmlb:
:tada: my ami build fine, see action run: https://github.com/supabase/postgres/actions/runs/27021781886?pr=2183

Dropping the pg versions bump and going to try to get some review for merging.

samrose (approved):

PR #2185, opened 2026-06-05
Internalize PostgREST mirror workflow

jgoux:
## Summary
- replace the `supabase/cli` reusable mirror workflow in `mirror-postgrest.yml`
- mirror `docker.io/postgrest/postgrest:v<version>` directly from this repo
- push the mirrored tag to `public.ecr.aws/supabase/postgrest:v<version>` and `ghcr.io/supabase/postgrest:v<version>`

## Notes
The Postgres image release workflows already call this repo’s local `mirror.yml`. This change removes the remaining cross-repo mirror dependency for the PostgREST image.

## Validation
- parsed `.github/workflows/mirror-postgrest.yml` as YAML
- ran `git diff --check`

### DIFF
--- a/.github/workflows/mirror-postgrest.yml
+++ b/.github/workflows/mirror-postgrest.yml
@@ -27,11 +27,28 @@ jobs:
   mirror:
     needs:
       - version
+    runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
       id-token: write
-    uses: supabase/cli/.github/workflows/mirror-image.yml@main
-    with:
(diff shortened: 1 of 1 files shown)


jgoux:
Actually I'm not sure it's the best move, it would be better to have this process directly internalized in PostgREST just after they deploy their docker image instead of waiting for someone to update PostgREST version in here to trigger this workflow.
On the CLI side, dependabot is updating the local stack image versions and we get failures because dependabot is faster than the human updating the PostgREST version in this repo so the image of PostgREST isn't mirrored yet when our CI runs against the dependabot's PR.

PR #2188, opened 2026-06-05
packer: Run scripts with stdout going to stderr

mmlb:

Bug fix/debugging.

Still seeing `exited 123` errors even after my last PR went in (which is not surprising as I didn't actually change any logic). Not sure why my PR and AMI build didn't trip up the error but its definitely happening still https://github.com/supabase/postgres/actions/runs/27033322500/job/79809778592. Looking at the logs that we now have the last line printed is `umount /mnt` and claude-code guesses that it has to do with salt being updated at the end in `90-cleanup.sh` and running something from post-install hooks/scripts.

## What is the new behavior?

I've set it so stdout goes to stderr for the _main_ scripts run by packer so that we don't have unbuffered stderr showing up in the middle of buffered stdout in a point that appears way too early to make sense. Now they'll both go to stderr which is unbuffered by default and should be in sync (this way we also won't lose any logs to buffers in case of OOM or something else).

## Additional context

Holding of on marking salt packages as held to see if I can trigger this once.

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.090-orioledb"
-  postgres17: "17.6.1.133"
-  postgres15: "15.14.1.133"
+  postgresorioledb-17: "17.6.0.090-orioledb-mmlb-err123-1"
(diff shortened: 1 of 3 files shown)


samrose (approved):

mmlb:
No luck reproducing the AMI build, I'm starting to think its only on merge-queue builds...

PR #2195, opened 2026-06-09
test: MERGE serialization fix with an isolation spec

utkarash2991:
Adds a MERGE serialization isolation spec on top of the `pg_isolation_regress` harness (#2189).

The spec pins the upstream fix (PG 15.16 / 17.8): `MERGE` must raise a serialization failure (`40001`) under `REPEATABLE READ` when its target row was updated and committed by a concurrent transaction after the MERGE's snapshot. Pre-fix, MERGE silently skipped this concurrent-update check.

Ref: PSQL-1277


mmlb:
This is going to come in as part of #2155, so closing.

PR #2198, opened 2026-06-09
fix(ansible): refactor kernel parameters and fix tuned integration

hunleyd:
This PR refactors how kernel parameters and `tuned` profiles are handled to ensure reliability and prevent command-line corruption.

#### 1. Tuned & Kernel Parameter Refactoring
- **Ansible:** Fixed `tuned` integration in `setup-tuned.yml`.
    - Removed the `[bootloader]` section from the `postgresql` tuned profile. This section was identified as dangerous as it could lead to corruption of the kernel command line during profile activation.
    - Moved `tcp_keepalive_intvl` and `tcp_keepalive_time` settings from `setup-system.yml` into the `postgresql` tuned profile's `[sysctl]` section.
    - Added guards (`when: stage2_nix`) to `tuned` service activation to avoid failures when running in restricted environments (like chroot during AMI builds).
- **GRUB/Bootstrap:** 
    - Moved the application of critical boot-time kernel parameters to `ebssurrogate/scripts/chroot-bootstrap-nix.sh`.
    - Restored `zswap` parameters (`zswap.enabled=1 zswap.zpool=zsmalloc zswap.compressor=zstd zswap.max_pool_percent=10`) to `GRUB_CMDLINE_LINUX_DEFAULT`.
    - Added `transparent_hugepage=never` to `GRUB_CMDLINE_LINUX_DEFAULT`.
    - Included documentation in the bootstrap script clarifying that unknown kernel parameters are safely ignored, ensuring compatibility across different kernel versions.

#### 2. Systemd & Cleanup
- **PostgreSQL:** Corrected the systemd override path in `ansible/tasks/setup-postgres.yml` from `override.conf` to `overrides.conf` for the `tuned.service` dependency.
- **Cleanup:** Removed redundant `net.ipv4.tcp_keepalive_*` sysctl tasks from `setup-system.yml`.

### Why?
- The `tuned` bootloader plugin is known to be fragile in certain environments and was causing issues with command-line management.
- Applying kernel parameters via GRUB is more robust for settings that should be active from early boot.
- Consolidating PostgreSQL-related sysctls into the `tuned` profile makes the configuration more modular and easier to manage.


ansible/tasks/setup-system.yml
```diff
@@ -161,14 +161,6 @@
...
         state: 'present'
 
-    - name: Set net.ipv4.tcp_keepalive_intvl=60
-      ansible.builtin.sysctl:
-        name: 'net.ipv4.tcp_keepalive_intvl'
-        value: 60
-        state: 'present'
-
-    # postgres_exporter runs on port 9187 and postgresT occasionlly chooses it as random srcport
-    # adminapi for 8085
```
romanhx, line 164: seem like useful comments to keep

romanhx (changes requested):
v minor request

ansible/tasks/setup-system.yml
romanhx: unless this is duplicated elsewhere?

hunleyd: fair, but also, this task goes away in #2199 :)

romanhx: true!

romanhx (approved):

ebssurrogate/scripts/chroot-bootstrap-nix.sh
```diff
@@ -329,12 +329,15 @@ function setup_apparmor {
 }
 
 function setup_grub_conf {
+	# Note: Unknown kernel parameters (like zswap settings on kernels without zswap support)
+	# are safely ignored by the kernel and passed to user-space. This allows us to
+	# include them here without risking boot failures on older or incompatible kernels.
```
romanhx, line 334: neat!

ebssurrogate/scripts/chroot-bootstrap-nix.sh
```diff
@@ -329,12 +329,15 @@ function setup_apparmor {
...
 	cat <<EOF >/etc/default/grub
 GRUB_DEFAULT=0
 GRUB_TIMEOUT=0
 GRUB_TIMEOUT_STYLE="hidden"
 GRUB_DISTRIBUTOR="Supabase postgresql"
-GRUB_CMDLINE_LINUX_DEFAULT="nomodeset console=tty1 console=ttyS0 ipv6.disable=0"
+GRUB_CMDLINE_LINUX_DEFAULT="nomodeset console=tty1 console=ttyS0 ipv6.disable=0 transparent_hugepage=never zswap.enabled=1 zswap.zpool=zsmalloc zswap.compressor=zstd zswap.max_pool_percent=10"
```
romanhx, line 340: Just wanna make sure im following, but we don't do the check `ansible_facts['swaptotal_mb'] ` like we used to because the tuned task only runs when it's set later, yes?

romanhx (approved):

ebssurrogate/scripts/chroot-bootstrap-nix.sh
hunleyd: si

PR #2203, opened 2026-06-09
ansible: move ip_local_port_range to tuned profile

hunleyd:
Relocates net.ipv4.ip_local_port_range from setup-system.yml to the postgresql tuned profile in setup-tuned.yml.

### DIFF
--- a/ansible/tasks/setup-tuned.yml
+++ b/ansible/tasks/setup-tuned.yml
@@ -135,6 +135,19 @@
         state: 'present'
         value: '3000,3001,8085,9122,9187,9999'
 
+    - name: tuned - Set ip_local_port_range
+      become: true
+      community.general.ini_file:
+        create: true
+        group: 'root'
+        mode: '0644'
+        no_extra_spaces: true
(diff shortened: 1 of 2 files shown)


ansible/tasks/setup-tuned.yml
```diff
@@ -111,6 +111,19 @@
...
+          community.general.ini_file:
+            create: true
+            group: 'root'
+            mode: '0644'
+            no_extra_spaces: true
+            option: 'net.ipv4.ip_local_port_range'
+            path: '/etc/tuned/profiles/postgresql/tuned.conf'
+            section: 'sysctl'
+            state: 'present'
+            value: '1025 65499'
```
romanhx: Just wanna confirm you intended on `65499` instead of the original `65000`

hunleyd: yes, its currently tweaked elsewhere to this value, so this adopts that value. you can see the 'consolidated from everywhere else' values in INDATA-387

romanhx (approved):

PR #2205, opened 2026-06-10
chore: clean up .gitignore file and most of docker/ subdir

mmlb:

Refactor/clean up

.gitignore has old and unnecessary rules.
We also used differently named files for keeping empty directories (.keep, .gitkeep).
We also still have many unused files in docker/ dir that can go away.

## What is the new behavior?

Leaner .gitignore file, standardize on .gitkeep for tracking empty dirs, delete unused files/dirs in docker/ dir.


hunleyd:
@claude review

hunleyd (approved):

PR #2207, opened 2026-06-10
testinfra: Fix InstanceType and startup error handling

mmlb:
This is used to get logs from journalctl but when the service name is wrong there's nothing to report and thus is very unhelpful.

Buf fix 

If testinfra times out waiting on a healthy postgres service it will use journalctl to show last few lines of log, but the service name is wrong therefore the logs are empty. For example see https://github.com/supabase/postgres/actions/runs/27230790541/job/80420986600#step:12:161

## What is the new behavior?

Correct systemd service name should give us logs next time we hit this.


hunleyd:
@claude review

hunleyd (approved):

PR #2214, opened 2026-06-12
ci: Use build-ami action in Release AMI Nix workflow too

mmlb:
Release AMI Nix gained this recently and we should have had that applied here too, doing so now.

## What is the new behavior?

## Additional context


mmlb:
nah, this is going to take more cleaning up than I have time for atm and maybe it'll all go away soon enough.

PR #2216, opened 2026-06-12
nix: Avoid unnecessary nixpkgs git checkout

mmlb:

DevEx (minor)

I noticed we had an extra nixpkgs in flake.lock, was easy to miss because its actually something that nix-eval-jobs should have done on its own (by propagating the nixpkgs input to its treefmt-nix). This does have a slight/tiny effect on build times since we waste time fetching a nixpkgs tree and copying into /nix/store only for it to never be used.

## What is the new behavior?

No time wasted fetching unnecessary nixpkgs tree, no time wasted copying into /nix/store.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -265,18 +265,14 @@
     },
     "nixpkgs-oldstable": {
       "locked": {
-        "lastModified": 1712666087,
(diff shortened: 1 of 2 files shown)


mmlb:
Dropping this. Forcing nix-eval-job's treefmt-nix to use our nixpkgs causes an issue with missing `rumdl`. Next time we update nixpkgs we should update nix-eval-jobs too which already does the nixpkgs follows internally.

PR #2217, opened 2026-06-14
ci: nix-build: Drop dependencies on nix-build-checks-*

mmlb:

DevEx improvement by increasing PR performance

The `run-testinfra` and `run-tests` jobs are depending on the `nix-build-checks-*` jobs. They each take somewhere between 2-6mins.

## What is the new behavior?

 These `run-test*` jobs don't actually depend on anything from those `check` jobs (which are mostly formatter/linter checks anyway) so no need to wait ~6m to get real checks to run.

## Additional Context

I also cleaned up the `if:`s since they were written as if the dependents were showing up skipped which is not currently the case:

```
gh run view 27444894989 --json jobs --jq '.jobs[] | select(.name == "no packages to build (aarch64-darwin)") | .status, .conclusion'
completed
success
```

So this is a `kaizen` clean up that makes for a more readable file.

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -61,7 +61,9 @@ jobs:
     name: >-
       ${{ matrix.name }}${{ matrix.postgresql_version && format(' - Postgres {0}', matrix.postgresql_version) || '' }}
       (aarch64-linux)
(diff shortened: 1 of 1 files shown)


hunleyd (approved):

samrose (approved):
reviewed do not see any issue.

PR #2220, opened 2026-06-15
fix(pg_upgrade): retry nix-store -r to avoid hangs on stalled S3 fetches

jfreeland:
## What

Wrap the binary-cache `nix-store -r` step in the pg_upgrade initiate script with a per-attempt timeout and a small retry loop.

## Why

`nix-store -r` can stall indefinitely on a dropped S3 connection without erroring out, its own download timeout doesn't reliably fire. When that happens the upgrade hangs forever reporting `running`, eventually exhausting the caller's poll budget with no useful error.

## How

- Each attempt runs under `timeout` (3 × 120s); `nix-store` is resumable, so a retry only re-fetches the paths that didn't finish.
- On exhaustion it fails as a normal command (not `exit 1`) so the existing `ERR` trap runs cleanup and records `failed`, turning an indefinite hang into a fast, reported failure.
- Worst case (360s) stays within the nix-fetch window the upgrade poller already allows, so no caller-side timeout changes are needed.

### DIFF
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -358,12 +358,12 @@ EXTRA_NIX_CONF
 		# and record "failed" instead of hanging.
 		nix_store_ok="false"
 		for attempt in 1 2 3; do
-			if timeout -k 10 120 nix-store -r "$STORE_PATH"; then
+			if timeout -k 10s 120s nix-store -r "$STORE_PATH"; then
 				nix_store_ok="true"
 				break
 			fi
 			if [ "$attempt" -lt 3 ]; then
-				echo "WARNING: nix-store -r attempt ${attempt}/3 for $STORE_PATH failed or stalled (>120s); retrying"
+				echo "WARNING: nix-store -r attempt ${attempt}/3 for $STORE_PATH failed or stalled (>=120s + up to 10s kill grace); retrying"
 			else
 				echo "ERROR: nix-store -r failed after 3 attempts for $STORE_PATH"
 			fi
--- a/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
+++ b/ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
@@ -348,8 +348,22 @@ EXTRA_NIX_CONF
 
 		echo "Store path: $STORE_PATH"
 
-		# Realize from binary cache (no nix evaluation needed!)
-		nix-store -r "$STORE_PATH"
+		# Realize the closure from the binary cache.
+		#
+		# nix-store -r can stall indefinitely on a dropped S3 connection without
+		# erroring out (its own download timeout doesn't reliably fire), so guard each
+		# attempt with a timeout and retry. nix-store is resumable, so a retry only
+		# re-fetches the unfinished paths. Failing as a normal command (not exit 1)
+		# lets the ERR trap run cleanup and record "failed" instead of hanging.
+		nix_store_ok="false"
+		for attempt in 1 2 3; do
+			if timeout -k 10 120 nix-store -r "$STORE_PATH"; then
+				nix_store_ok="true"
+				break
+			fi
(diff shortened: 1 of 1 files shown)


ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -348,8 +348,22 @@ EXTRA_NIX_CONF
...
 		echo "Store path: $STORE_PATH"
 
-		# Realize from binary cache (no nix evaluation needed!)
-		nix-store -r "$STORE_PATH"
+		# Realize the closure from the binary cache.
+		#
+		# nix-store -r can stall indefinitely on a dropped S3 connection without
+		# erroring out (its own download timeout doesn't reliably fire), so guard each
+		# attempt with a timeout and retry. nix-store is resumable, so a retry only
+		# re-fetches the unfinished paths. Failing as a normal command (not exit 1)
```
samrose: Resumability is per-path, not per-byte. Nix realisation iterates the closure, and each individual NAR is downloaded to a temp location and only registered as a valid store path after it completes and its hash verifies. Registration is transactional, so a SIGKILL mid-download leaves the store consistent and that path simply invalid. The next nix-store -r invocation checks `isValidPath` per path and skips the ones already registered. That's why retries work. But the path that was in flight when timeout fired restarts from 0% on the next attempt; there's no byte-level resume within a single NAR. So if your closure contains one NAR that takes longer than 120s to fetch over the link you're worried about, every attempt dies at 120s mid-fetch on that path and you get three guaranteed failures with zero forward progress. You have to just make sure 120s comfortably exceeds your single largest NAR's transfer time at your worst expected S3 throughput, or split/raise the budget accordingly. Everything smaller than that converges fine across attempts.

samrose (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -348,8 +348,22 @@ EXTRA_NIX_CONF
...
+		# lets the ERR trap run cleanup and record "failed" instead of hanging.
+		nix_store_ok="false"
+		for attempt in 1 2 3; do
+			if timeout -k 10 120 nix-store -r "$STORE_PATH"; then
+				nix_store_ok="true"
+				break
+			fi
+			echo "WARNING: nix-store -r attempt ${attempt}/3 for $STORE_PATH failed or stalled (>120s); retrying"
```
Crispy1975: The only thing I can same here as a small nit is that on the final iteration the log will say _retrying_ when it will not. Maybe we can have a final _failed_ log line? It might be a bit clearer to operators logging in to check?

Crispy1975 (commented):
LGTM, but I will defer to @samrose on nix things.

Crispy1975 (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
jfreeland: added a very very brief version of this so it's at least kind of captured and people can find the PR for this context in the future. :)

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
jfreeland: updated to be more clear

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -348,8 +348,27 @@ EXTRA_NIX_CONF
...
+		# nix-store -r can stall indefinitely on a dropped S3 connection without
+		# erroring out (its own download timeout doesn't reliably fire), so guard each
+		# attempt with a timeout and retry. Each path downloads atomically: already-
+		# registered paths are skipped on retry but an in-flight NAR restarts from
+		# 0%. Failing as a normal command (not exit 1) lets the ERR trap run cleanup
+		# and record "failed" instead of hanging.
+		nix_store_ok="false"
+		for attempt in 1 2 3; do
+			if timeout -k 10s 120s nix-store -r "$STORE_PATH"; then
+				nix_store_ok="true"
```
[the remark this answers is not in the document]
jfreeland: caller polls for longer

PR #2229, opened 2026-06-18
Fix amd64 AMI Builds

mmlb:

Bug fix and refactors.

Flaky AMI builds.

## What is the new behavior?

No more flaky AMI builds.

## Additional context

We've been seeing issues with building the AMI, specifically the `Release AMI Nix` workflow. Took quite some debugging to figure out that we were hit with https://github.com/hashicorp/packer-plugin-amazon/issues/676 ultimately due to packer not supporting lock files and always chasing the newest release possible.

The issue boiled down to packer_plugin_amazon causing ~600% cpu load (6 cpus @ 100%) on the blacksmith machines which ended up causing the network to backup until the connection is forced closed by a middleware/hv and the build fail. Looks like arm machines were managing the load well enough to continue passing but we were also affected by the bug there too.

I'm still not sure why `Release AMI Nix` seemed to trigger this bug more reliably even after re-working it to use the same `buid-ami` action and thus `build-ami` nix package but in the meantime I was able to get a reproducer and narrow down on the issue. I blamed the blacksmith networking (because if its not DNS, its the network) but that was only another symptom (would be nice to have a better indicator/behavior though). Luckily support pointed out that the cpus were at 100% and that caused the network to drop. I don't know how I didn't notice that myself on the runner's metrics page. I recall focusing on network io instead. Once I saw the cpu usage I tried to replicate on my own machine to factor out blacksmith possibly throttling/slow cpu but saw the same behavior which finally led me the bug in question.

Fix is to pin to 1.8.0 for now. I'll be working on the CI in general next and will enable renovate once I get it so we aren't running long builds for unrelated changes and are sure that testinfra and co are required to pass for PRs to be mergeable. I am keeping all the "unrelated" cleanups though because they are good and kaizen.

### DIFF
--- a/nix/packages/build-ami.nix
+++ b/nix/packages/build-ami.nix
@@ -1,10 +1,11 @@
 {
   lib,
   stdenv,
-  writeShellApplication,
-  packer,
   awscli2,
+  coreutils,
   jq,
+  packer,
+  writeShellApplication,
   ...
 }:
 
@@ -41,14 +42,13 @@ writeShellApplication {
   name = "build-ami";
 
   runtimeInputs = [
-    packer
     awscli2
+    coreutils
     jq
+    packer
   ];
 
   text = ''
-    set -euo pipefail
-
     set -x
 
     # Parse required parameters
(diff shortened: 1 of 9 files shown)


nix/packages/build-ami.nix
```diff
@@ -63,22 +63,48 @@ writeShellApplication {
...
+    INPUT_HASH=''${INPUT_HASH#/nix/store/}
+    INPUT_HASH=''${INPUT_HASH%%-*}
     shift 2
 
+    export PACKER_LOG=''${PACKER_LOG:-''${RUNNER_DEBUG:-0}}
+    on_error=ask
+    if ''${CI:-false}; then
+      echo "::notice::Setting packer build -on-error=abort since this is CI, this is different than non-CI runs!"
+      on_error=abort
+    elif ! [[ -t 1 ]]; then
```
mmlb: This should be `-t 0` actually

samrose (approved):
Don't know for sure but Potential issue: the release workflow’s path filter no longer covers all files used by the workflow, so some future AMI-build changes may not trigger this release workflow automatically.

mmlb:
> Don't know for sure but Potential issue: the release workflow’s path filter no longer covers all files used by the workflow, so some future AMI-build changes may not trigger this release workflow automatically.

Should be no worse than today, I only dropped the common- packer file which isn't in git so would never be part of a push anyway. I'll be back here soonish anyway taking a look at all of these paths and workflows.

PR #2230, opened 2026-06-18
feat: handle pg_net persistence mismatch for PG14 upgrade

cfunkhouser:
Execute `ALTER TABLE net.http_request_queue SET LOGGED` on PG14 databases before upgrade.

There is a state mismatch in which `net.http_request_queue_id_seq` sequence is logged, but the table `net.http_request_queue` is not. This mismatch prevents `pg_upgrade` from proceeding. This change forces that alteration to occur at upgrade initiation time, before any of the other upgrade process has occurred.


Crispy1975 (approved):
LGTM; we have a ConfigCat `allowPg14LaunchOrgs` flag which allows an org slug to be added. This mean we can spin up a PG14 project and then do surgery to mimic the conditions we're gonna fix.

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -210,7 +210,27 @@ EOF
...
 }
 
+# handle_upgrade_prerequisites applies changes to the database to prevent known
+# migration errors between pg or extension versions, and must be run before any
+# other changes.
+function handle_upgrade_prerequisites {
+
+	# pg_net's net.http_request_queue was created UNLOGGED on PG14. Newer pg_net
+	# defines it as LOGGED, so the persistence mismatch breaks pg_upgrade.
+	# Convert it to LOGGED before making any other changes to the source server.
```
steve-chavez: Confused with the above as it's not true, the table is UNLOGGED since [pg_net v0.5](https://github.com/supabase/pg_net/releases/tag/v0.5) and has been kept that way until v0.20.3 now, see:

https://github.com/supabase/pg_net/blob/feaef7600103af5858ee7b99cf0b6700bb39f3ef/sql/pg_net.sql#L12-L19

steve-chavez (changes requested):

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
cfunkhouser: Okay, interesting. We do have cases where the table is unlogged, but the sequence is logged. I'm assuming this is because unlogged sequences were introduced in PG15, so the sequence being logged is sticky from before upgrade.

Also, since unlogged is default for pg_net, setting the table to logged and leaving it that way is likely not the whole solution. Likely, we should be setting table and sequence back to unlogged after upgrading to versions which support that. Does that make sense?

I'll make that change, and will scrub that incorrect comment.

cfunkhouser: I have adjusted the comment.

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
```diff
@@ -210,7 +210,27 @@ EOF
...
+# migration errors between pg or extension versions, and must be run before any
+# other changes.
+function handle_upgrade_prerequisites {
+	PG_NET_ENABLED=$(run_sql -A -t -c "SELECT EXISTS(SELECT 1 FROM pg_extension WHERE extname = 'pg_net');")
+
+	# Even though the pg_net tables are created UNLOGGED, PG14 does not support
+	# unlogged sequences, and so they are created LOGGED. This mismatch causes
+	# pg_upgrade trouble. To work around it, SET LOGGED before the upgrade
+	# process. It (and the sequence) are then SET UNLOGGED after the upgrade
+	# succeeds (in complete.sh).
```
steve-chavez, line 223: Q: Since the `http_request_queue` could be big (in some rare cases), would setting it to `LOGGED` have a perf impact?

IIUC, LOGGED involves a table rewrite + I/O + WAL generation.

steve-chavez:
Could you clarify why is pg_upgrade a problem?

If the table is UNLOGGED + sequence LOGGED, this should run fine on pg 15.

Or is the problem restoring from pg 15, where table + sequence are UNLOGGED, to pg 14 (which doesn't supported UNLOGGED tables)?

steve-chavez:
Also how do we decide if an upgrade script goes into ansible or in the Nix build itself?

For example we have some upgrade testing inside the Nix expression here:

https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/nix/ext/tests/postgis.nix#L99-L100

And in some cases we patch the extensions like on:

https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/nix/ext/pg_cron/pg_cron-1.3.1-pg15.patch#L16

IMO it looks it'd be better to handle this pg_net case inside the Nix expression itself since we already have a testing framework in place (and this can be ran locally).

cc @samrose

ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh
steve-chavez: If we patch the extension as [mentioned below](https://github.com/supabase/postgres/pull/2230#issuecomment-4773598997) perhaps we could avoid this step?

cfunkhouser: Yeah, I would definitely expect a perf impact for this. The changes here are scoped only to upgrade time, to get us out of the breaking situation where the persistence mismatch is not accepted during the upgrade, so no database should be online and serving traffic normally with `net.http_request_queue` logged.

cfunkhouser:
After some... prolonged internal discussion, I'm going to close this pull request. The TL;DR is, we're not actually sure how to reproduce the issue, and therefor cannot test that this fix does anything more than complicate these scripts. Will resurrect if we are able to repro.

cfunkhouser:
@steve-chavez to not leave you hanging on your question: this fix was for an issue affecting migrations from 14 to something more recent. 14 predates Nix, so for this particular concern, ansible is the only lever we have to pull. It's a good question to consider for other cases, once we get past 14 deprecation.

samrose:
> Also how do we decide if an upgrade script goes into ansible or in the Nix build itself?
> 
> For example we have some upgrade testing inside the Nix expression here:
> 
> https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/nix/ext/tests/postgis.nix#L99-L100
> 
> And in some cases we patch the extensions like on:
> 
> https://github.com/supabase/postgres/blob/e1570998e1ccaa05a5f49a560d2a4dc74b5908f1/nix/ext/pg_cron/pg_cron-1.3.1-pg15.patch#L16
> 
> IMO it looks it'd be better to handle this pg_net case inside the Nix expression itself since we already have a testing framework in place (and this can be ran locally).
> 
> cc @samrose

agree with handle in pg_net packaging in nix if it otherwise doesn't make sense generically in the extension itself and seems unique to an end user case (like supabase)

PR #2238, opened 2026-06-25
fix(security): insecure nix trusted-users config recommendation

### LINKED ISSUE #1622: Nix cache is not documented
## Problem

There's no doc explaining how to use the Nix cache, when you do `nix develop` it will compile everything.

One has to hunt the `echo` commands from:

https://github.com/supabase/postgres/blob/1772a43e063d4c763b868fbcf88e95eeaf9c2d6b/docker/nix/Dockerfile#L3-L5

Then `sudo pkill nix-daemon` for Nix to pick the updated settings ([ref](https://docs.cachix.org/faq#why-is-nix-not-picking-up-on-any-of-the-pre-built-artifacts))

## Solution

Document this

brainrake:

Fix insecure config recommendation. Add nix binary cache documentation.

`trusted-users` is recommended

## What is the new behavior?

`trusted-users` is warned against

## Additional context

https://nix.dev/manual/nix/stable/command-ref/conf-file.html#conf-trusted-users
https://github.com/NixOS/nix/issues/9649#issuecomment-1868001568

Fixes https://github.com/supabase/postgres/issues/1622
Related: https://linear.app/supabase/issue/PSQL-1323/insecure-nix-cache-config-recommendation-in-docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ The project uses Nix as its build system, which provides:
 
 ## Common Tasks
 
+To skip hours of building and download instead, configure the Supabase Postgres Nix binary cache: [nix/docs/binary-cache.md](nix/docs/binary-cache.md).
+
 ### Building Locally
 
 To build PostgreSQL with extensions locally:
--- a/docs/multigres-image.md
+++ b/docs/multigres-image.md
@@ -77,43 +77,33 @@ docker build -f Dockerfile-multigres --target variant-orioledb-17 -t pg-docker-t
 nix run .#docker-image-test -- --no-build --target variant-orioledb-17 Dockerfile-multigres
 ```
 
-### Optional: install nix
+### Install nix
 
 ## Install Nix (Fresh Installation)
 
 We'll use the official Nix installer with a custom configuration that includes our build caches and settings. This works on many platforms, including **aarch64 Linux**, **x86_64 Linux**, and **macOS**.
 
-### Step 1: Create nix.conf
+### Step 1: Create nix.conf.extra
 
-First, create a file named `nix.conf` with the following content:
+First, create a file named `nix.conf.extra` with the following content:
 
 ```
-allowed-users = *
-always-allow-substitutes = true
-auto-optimise-store = false
-build-users-group = nixbld
-builders-use-substitutes = true
-cores = 0
 experimental-features = nix-command flakes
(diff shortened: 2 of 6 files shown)


samrose (approved):

samrose (approved):

nix/docs/nix-binary-cache.md
```diff
@@ -0,0 +1,53 @@
...
+You need to edit this file: 
+
+```text
+/etc/nix/nix.conf
+```
+
+Add or merge this config:
+
+```conf
+always-allow-substitutes = true
```
mmlb: I'm not sure this is appropriate for a dev machine. The fix for our CI was great but probably not going to be too useful for a persistent dev machine and can make things take longer fetching these trivial files.

nix/docs/nix-binary-cache.md
```diff
@@ -0,0 +1,53 @@
...
+
+## NixOS or nix-darwin
+
+Add this to your system configuration:
+
+```nix
+{
+  nix.settings.extra-substituters = [
+    "https://nix-postgres-artifacts.s3.amazonaws.com"
+    "https://postgrest.cachix.org"
```
mmlb: we don't need the postgrest substituters, we currently install postgrest via ansible

nix/docs/nix-binary-cache.md
brainrake: Good catch. I thought it was necessary for nix-darwin linux-builder but on closer look it's not.

nix/docs/nix-binary-cache.md
brainrake: and we'll cache it anyway

mmlb: yep

brainrake:
For context: [PSQL-1408](https://linear.app/supabase/issue/PSQL-1408/ami-pre-populate-site-profile-at-bake-drop-pam-store-symlink) will have the AMI bake fetch `site-env` from the binary cache at bake time, so this trusted-users fix is directly relevant to that path too.

brainrake:
Note: [PSQL-1408](https://linear.app/supabase/issue/PSQL-1408) will rely on nix trust config from this PR when baking the `site` profile at AMI build time.

PR #2240, opened 2026-06-26
feat: enable customer installation of amcheck (PSQL-1327)

brainrake:
Closes [PSQL-1327](https://linear.app/supabase/issue/PSQL-1327/investigate-enabling-customer-installation-of-amcheck)

Moves the **amcheck** contrib extension into `supautils.privileged_extensions` so
customers can install it, pinned to the `pg_catalog` schema.

## Additional context

amcheck was on supautils' "may be unsafe" list based on [upstream
caution](https://www.postgresql.org/docs/current/amcheck.html) around physical data
structure exposure. Enabling it turned out to require the schema pin:

- supautils creates privileged extensions as `supabase_admin`, whose search_path
  starts with `public`.
- `ALTER DEFAULT PRIVILEGES ... IN SCHEMA public` (`initial-schema.sql`) grants
  EXECUTE to `postgres, anon, authenticated, service_role`. Those are explicit
  grants, so amcheck's `REVOKE ALL FROM PUBLIC` does not strip them.
- amcheck performs no permission check of its own upstream `verify_nbtree.c`:
  *"Intentionally not checking permissions"* so EXECUTE means "check any index in
  the database".

Verified before the pin: `set role anon; select bt_index_check('auth.users_pkey')`
**succeeded**. That would have exposed `bt_index_parent_check()` (ShareLock, blocks
writes) and `verify_heapam()` to unauthenticated PostgREST callers. In `pg_catalog`
the equivalent rule grants to `postgres` only  with grant option, so customers can
still share it deliberately.

## Tests

- `nix/tests/sql/amcheck.sql` — placement, override enforcement (`with schema public`
  is redirected), the four-role privilege matrix, and the functional path.
- `nix/ext/tests/amcheck-upgrade.nix`  15→17 `pg_upgrade` preserves schema and
  grants, including through the 1.3→1.4 extension update that creates new functions.

## Action Items

- [x] **New extension releases** checked for breaking changes
- [ ] **Extensions compatibility** — `supadev trigger-smoke-tests <PR> engines extensions`
- [ ] **Backup and Restore** — `upgrade` target covers pause-on-old/restore-on-new;
      single-file backup/restore still manual


brainrake:
https://github.com/supabase/postgres/pull/2240#issuecomment-4807768165
Perhaps this should also mention changes in the list of extensions.

hanefi:
Backup Testing section in current PR description has an item that does not sound reasonable to me.

>Pause on new Postgres version, restore on new Postgres version

Assuming old and new correspond to Postgres 15 and 17, this is not an easy task. I do not yet know what options do our users have for restoring but the most probable methods do not work.

Postgres 15 was shipped with amcheck `1.3` and 17 was shipped with `1.4`. `pg_dump` & `pg_restore` fails as the dump has explicit versions and the same version is not available on the older Postgres version.

Official support for downgrading to an older major version is explicitly not supported by PostgreSQL via standard restore tools. One can always try to manually fix issues in the dump and attempt to restore. It is highly likely that many iterations of edits and retries are required until you can actually downgrade your database version.

brainrake:
I just used the issue template. I only wrote the first two paragraphs.

brainrake:
@hanefi can you review wether amcheck is safe to enable for users?

samrose:
I am taking over the review and testing of this PR

samrose:
Will document next step of supadev smoke testing prior requesting review

ansible/files/postgresql_config/supautils.conf.j2
```diff
@@ -1,13 +1,21 @@
-supautils.extensions_parameter_overrides = '{"pg_cron":{"schema":"pg_catalog"}}'
+# amcheck must stay pinned to extensions. supautils creates privileged extensions as
+# supabase_admin, whose search_path starts with public, where ALTER DEFAULT PRIVILEGES
+# (init-scripts/00000000000000-initial-schema.sql) grants EXECUTE to anon, authenticated
+# and service_role, i.e. to unauthenticated PostgREST callers. amcheck does no permission
+# check of its own (upstream verify_nbtree.c: "Intentionally not checking permissions"),
+# so EXECUTE alone lets a role check any index in the database, including auth's. In
+# extensions the equivalent rule grants to postgres only.
+# See PSQL-1327 and nix/tests/sql/amcheck.sql.
+supautils.extensions_parameter_overrides = '{"pg_cron":{"schema":"pg_catalog"},"amcheck":{"schema":"extensions"}}'
```
soedirgo: Comment is unnecessary; also let's pin the schema to `pg_catalog` instead, users can drop the `extensions` schema

samrose:
latest changes as of this writing in this PR confirmed in staging

```
 | check                                                      | value                                                                 |
  | ---------------------------------------------------------- | --------------------------------------------------------------------- |
  | acl: anon                                                  | all=false  any=false                                                  |
  | acl: authenticated                                         | all=false  any=false                                                  |
  | acl: postgres                                              | all=true  any=true                                                    |
  | acl: service_role                                          | all=false  any=false                                                  |
  | amcheck_schema (want pg_catalog)                           | pg_catalog                                                            |
  | bt_index_check on an index postgres does NOT own (want ok) | ok                                                                    |
  | bt_index_check on owned index (want ok)                    | ok                                                                    |
  | pin                                                        | {"pg_cron":{"schema":"pg_catalog"},"amcheck":{"schema":"pg_catalog"}} |
  | postgres_is_superuser (want f)                             | false                                                                 |
```

imor (approved):
LGTM

ansible/files/postgresql_extension_custom_scripts/amcheck/after-create.sql
```diff
@@ -0,0 +1,19 @@
...
+  perform set_config('search_path', '', true);
+
+  for r in
+    select p.oid::regprocedure as sig
+      from pg_depend d
+      join pg_proc p on p.oid = d.objid and d.classid = 'pg_proc'::regclass
+      join pg_extension e on e.oid = d.refobjid
+     where e.extname = 'amcheck' and d.deptype = 'e'
+  loop
+    execute format('grant execute on function %s to postgres', r.sig);
```
soedirgo: This may not be causing problems now but it doesn't handle overloaded functions. Also, use `on routine` instead of `on function` to handle procedures

```suggestion
    select p.oid::regprocedure as sig,
      pg_get_function_identity_arguments(p.oid) as args
      from pg_depend d
      join pg_proc p on p.oid = d.objid and d.classid = 'pg_proc'::regclass
      join pg_extension e on e.oid = d.refobjid
     where e.extname = 'amcheck' and d.deptype = 'e'
  loop
    execute format('grant execute on routine %s(%s) to postgres', r.sig, r.args);
```

ansible/vars.yml
```diff
@@ -11,9 +11,9 @@ postgres_major:
 # This is the source of truth for Postgres versions used in the Dockerfiles, and
 # is used to derive image tags and base images in the release matrix.
 postgres_release:
-  postgresorioledb-17: "17.9.0.019-orioledb"
-  postgres17: "17.6.1.166"
-  postgres15: "15.14.1.166"
+  postgresorioledb-17: "17.9.0.018-orioledb-amcheck-2"
+  postgres17: "17.6.1.165-amcheck-2"
+  postgres15: "15.14.1.165-amcheck-2"
```
soedirgo: Remember to update this

ansible/files/postgresql_extension_custom_scripts/amcheck/after-create.sql
soedirgo: This is not causing issues for now, but use `routine` to make it work for both functions and procedures:

```suggestion
    execute format('grant execute on routine %s to postgres', r.sig);
```

soedirgo (approved):

ansible/vars.yml
samrose: done

samrose:
his issue with be handed off to managed postgres team to make decisions about how to handle upgrades, pause/restore etc cc @KevinBrolly

brainrake:
Taking over drive to merge this. Reviewing and testing upgrade.

brainrake:
Pause/restore risk is low: Supabase doesn't support restoring a project onto a different major Postgres version, so the amcheck 1.3↔1.4 version-pin issue (dump/restore across 15↔17) doesn't apply there — pause/resume is same-version stop/start only. The pg_upgrade path is the real risk; tracked separately.

brainrake:
Upgrade risk: PG15 ships amcheck 1.3, PG17 ships 1.4. pg_upgrade runs `ALTER EXTENSION amcheck UPDATE`, which adds new 1.4-only function overloads (e.g. `bt_index_check(regclass, boolean, boolean)`). `after-create.sql` only grants EXECUTE to `postgres` at `CREATE EXTENSION` time — supautils has no after-update hook — so those new functions are ungranted after an upgrade, silently defeating the fix for exactly the customers who just upgraded.

Fix: add `amcheck` to `EXTENSIONS_TO_DISABLE` in `initiate.sh` so it's dropped pre-upgrade and recreated fresh post-upgrade (same mechanism already used for pg_graphql/pg_stat_monitor/etc.), instead of surviving via in-place update.

brainrake:
Restored `amcheck-upgrade.nix` (and its after-create.sql comment) and turned the printed known-limitation note into a hard assertion. Ran locally: it reproduces the regression — after a real 15→17 `pg_upgrade`, `postgres` gets `permission denied` calling amcheck 1.4's new `bt_index_check(regclass, boolean, boolean)` overload. Expected red until the `EXTENSIONS_TO_DISABLE` fix lands (though this harness runs a bare `pg_upgrade`, so that specific fix won't turn it green — tracked separately).

brainrake:
Fix pushed: added `amcheck` to `EXTENSIONS_TO_DISABLE` in `initiate.sh`, same drop-before/recreate-after mechanism already used for pg_graphql etc. Verified locally in `amcheck-upgrade.nix` (mirroring the same drop/recreate logic). Closes the upgrade permission gap.

PR #2241, opened 2026-06-26
Bandaid for darwin runners being disconnected

mmlb:

CI env fix

self hosted bare metal macos darwin runners are showing up as offline.

## What is the new behavior?

Lets use blacksmith runners for anything that doesn't explicitly required virtualization. Less management/headaches this way.

## Additional context

Still need to get these fixed up for the "kvm" tests that will come up at some point, also waiting on info from Blacksmith about running these virt cases on their runners.


samrose (approved):

mmlb:
Going ahead with the merge since `run-testinfra/test-ami-nix` jobs aren't blocking anything and really shouldn't even be running in this case since they aren't really affected.

PR #2242, opened 2026-06-29
feat: add `USING heap` to pg_cron installation script

za-arthur:

Fix `pg_cron`'s installation script and add `USING heap` to create heap tables. Otherwise smoke tests fails for orioledb tests with the error:
```
2026-06-27T12:00:03.943209Z ERROR supadev::run:execute_sql_with_connection{allow_failures=false conn.label=direct conn.host=db.xvnsysuyfaosipddgvhm.supabase.red conn.port=5432}: supadev::psql: error=psql failed with code 3: ERROR:  attempted to delete invisible tuple
```

ORI-172

## What is the new behavior?

The installation script will have the following statement:
```
CREATE TABLE cron.job (
	jobid bigint primary key default pg_catalog.nextval('cron.jobid_seq'),
	schedule text not null,
	command text not null,
	nodename text not null default 'localhost',
	nodeport int not null default pg_catalog.inet_server_port(),
	database text not null default pg_catalog.current_database(),
	username text not null default current_user
) USING heap;
```

## Additional context

### DIFF
--- a/ansible/vars.yml
+++ b/ansible/vars.yml
@@ -10,9 +10,9 @@ postgres_major:
 
 # Full version strings for each major version
 postgres_release:
-  postgresorioledb-17: "17.6.0.099-orioledb"
-  postgres17: "17.6.1.142"
-  postgres15: "15.14.1.142"
+  postgresorioledb-17: "17.6.0.100-orioledb"
+  postgres17: "17.6.1.143"
+  postgres15: "15.14.1.143"
 
 # Non Postgres Extensions
 pgbouncer_release: 1.25.1
--- a/nix/ext/pg_cron/pg_cron-heap-tables.patch
+++ b/nix/ext/pg_cron/pg_cron-heap-tables.patch
@@ -1,0 +1,26 @@
+diff --git a/pg_cron--1.2--1.3.sql b/pg_cron--1.2--1.3.sql
+index b1e6a68..f4b94e7 100644
+--- a/pg_cron--1.2--1.3.sql
++++ b/pg_cron--1.2--1.3.sql
+@@ -12,7 +12,7 @@ CREATE TABLE cron.job_run_details (
+ 	return_message text,
+ 	start_time timestamptz,
+ 	end_time timestamptz
+-);
++) USING heap;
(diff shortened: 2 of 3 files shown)


nix/ext/pg_cron/default.nix
```diff
@@ -63,6 +63,9 @@ let
           mv $out/share/postgresql/extension/pg_cron--1.5--1.6.sql $out/share/postgresql/extension/pg_cron--1.5.2--1.6.4.sql
         fi
 
+        find $out/share/postgresql/extension -name "*.sql" \
```
samrose: @za-arthur I am going to set this as changes requested to block merging just while I investigate if this should cause any problem at all to do this exactly here. Once I confirm one way or the other will respond here and change review accordingly

samrose (changes requested):

samrose:
If we make this change to the targeted pg_cron tables, It's now pinned to heap forever. If orioledb later fixes the DELETE bug and you'd want these tables on orioledb, the install SQL hard-codes heap and an upgrade won't move them. Migrating an existing table off heap later requires an explicit ALTER TABLE ... SET ACCESS METHOD.

I think we should research if anyone already applies TAM in our existing running projects before we make this change and release it. Possibly not, but I believe it is worth checking so that we don't have to fix it later if it turns out they are.

za-arthur:
Smoke tests managed to finish:
- [test-postgres-composite.sh](https://github.com/supabase/supadev/actions/runs/28572700404)
- [test-postgres-disk-pressure.sh](https://github.com/supabase/supadev/actions/runs/28572701463)
- [test-postgres-engines-with-smoke.sh](https://github.com/supabase/supadev/actions/runs/28572703678)
- [test-postgres-pitr.sh](https://github.com/supabase/supadev/actions/runs/28572705957)
- [test-postgres-replication.sh](https://github.com/supabase/supadev/actions/runs/28572707811)
- [test-postgres-upgrade-with-smoke.sh](https://github.com/supabase/supadev/actions/runs/28572709560)

samrose (approved):

PR #2245, opened 2026-06-29
feat(ansible): add net.core sysctl tuning params to tuned profile

hunleyd:
## What

Adds the following `net.core.*` parameters to the `sysctl` section of the PostgreSQL tuned profile (`ansible/tasks/setup-tuned.yml`):

| option | value |
| --- | --- |
| `net.core.netdev_budget` | `1024` |
| `net.core.netdev_max_backlog` | `10000` |
| `net.core.rmem_default` | `262144` |
| `net.core.rmem_max` | `104857600` |
| `net.core.somaxconn` | `16384` |
| `net.core.wmem_default` | `262144` |
| `net.core.wmem_max` | `104857600` |

It also adds a tracked `.yamllint` config (and a `.gitignore` negation so it survives the repo's `.*` dotfile ignore) so that standalone `yamllint` agrees with the yamllint profile `ansible-lint` already applies internally.

## Why / Benefit to PostgreSQL

These kernel network-stack settings improve how Postgres handles connections and high-throughput client traffic:

- **`net.core.somaxconn` (16384)** — raises the ceiling on the listen backlog so bursts of incoming connections aren't dropped during connection storms (e.g. poolers reconnecting, many clients starting at once). Pairs with Postgres's own `listen` backlog.
- **`net.core.netdev_max_backlog` (10000)** — allows more incoming packets to queue when the NIC receives faster than the kernel can process, reducing packet drops under heavy network load and the retransmits/latency spikes that follow.
- **`net.core.netdev_budget` (1024)** — lets the kernel process more packets per softirq poll cycle, improving network throughput on busy database servers handling many concurrent connections.
- **`net.core.rmem_default` / `rmem_max` (256 KiB / 100 MiB)** — larger receive socket buffers let Postgres backends absorb large inbound payloads (bulk INSERT/COPY, large query parameters) without stalling on slow/high-latency links.
- **`net.core.wmem_default` / `wmem_max` (256 KiB / 100 MiB)** — larger send socket buffers let backends ship large result sets (big SELECTs, COPY TO, replication streams) without blocking, improving throughput over high-bandwidth-delay-product networks.

Together these raise the buffer and queue headroom of the network stack so a high-connection, high-throughput Postgres instance is less likely to drop connections or bottleneck on the kernel under load.

## Notes

- `net.core.somaxconn` was previously set to `16834` — corrected to `16384` (transposed-digit typo).
- The `.yamllint` config mirrors `ansible-lint`'s embedded yamllint profile (line-length off, `document-start` disabled, `braces`/`truthy` relaxed for Ansible). With it in place, `setup-tuned.yml` passes both `ansible-lint` and `yamllint` with no playbook content changes. Pre-existing lint debt in other, untouched task files is intentionally left out of scope.

INDATA-378


hunleyd:
@claude review

matthudsonatx (approved):

ansible/tasks/setup-tuned.yml
```diff
@@ -123,8 +123,20 @@
...
         - option: 'kernel.shmmni'
           value: '4096'
+        - option: 'net.core.netdev_budget'
+          value: '1024'
+        - option: 'net.core.netdev_max_backlog'
+          value: '10000'
+        - option: 'net.core.rmem_default'
+          value: '262144'
+        - option: 'net.core.rmem_max'
+          value: '104857600'
```
jfreeland, line 133: this seems awfully high for free tier?  cc: @danielmitterdorfer for review from #team-perf-eng

hunleyd: you're probably not wrong, but as I said on Slack, i'm just consolidating the values from elsewhere in our repos into tuned. once it's all in tuned, then we can talk instance sizing 🤞🏻

PR #2255, opened 2026-07-03
INDATA-841: restrict to system journal for pgbouncer logs

pinaraf:
Instead of checking all journal files, restrict to the system journal to reduce the system load from fail2ban.


pinaraf:
(double-checking everything, sorry for the noise)

PR #2256, opened 2026-07-06
update nixpkgs

brainrake:

Update nixpkgs package set.

Package set is not up to date.

## What is the new behavior?

Package set is up to date.

## Additional context

TODO: make weekly scheduled github action to create pr

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1764011051,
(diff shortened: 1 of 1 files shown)


samrose:
Good to see but also want to caution that this kind of change will need extensive testing until enough of a robust testing regime exists that would allow it to be a reliable operation to happen on a regular basis.

brainrake:
there is a scheduled workflow for this, i'll trigger that instead

brainrake:
workflow is broken, will check back

PR #2261, opened 2026-07-07
fix: make pg_net background worker run as postgres

imor:
This PR updates `pg_net.username` GUC to `postgres` to ensure the pg_net background worker runs as the `postgres` user and not as the `supabase_admin` user. It also adds a test confirming the behaviour after updating the GUC.


nix/tests/sql/pg_net_worker_privileges.sql
```diff
@@ -0,0 +1,76 @@
...
+-- so literally any role -- including a plain `authenticated` app user -- can attach
+-- a trigger to them. Since a normal (non SECURITY DEFINER) trigger runs with the
+-- privileges of whoever performs the triggering statement, an `authenticated` user
+-- could make such a trigger run arbitrary SQL with the pg_net worker's privileges
+-- the next time it processed a queued request.
+--
+-- This test plants such a trigger as `authenticated` and has it record which role
+-- and privilege level it actually ran with, confirming the worker's DML now runs
+-- as the unprivileged `postgres` role instead of the superuser `supabase_admin`.
+
```
za-arthur, line 22: Can you add to the test a check that the worker executes triggers as indeed the `postgres` role?

imor: It was mainly the comment that mentioned the wrong role (`authenticated`), the test actually did verify that the triggers ran as `postgres` in line 69. Fixed it in a new commit.

steve-chavez (approved):

PR #2267, opened 2026-07-12
fix(nix): use nixpkgs pgbackrest instead of a from-source build

hunleyd:
## Summary

Adds `pg-backrest` to the Linux-only CI build-check list in `nix/checks.nix`, so a break in the pgbackrest flake output is caught at CI time instead of first surfacing during an AMI build. The `pg-backrest` package itself already landed on `develop` independently via #1714; this PR only adds the CI check for it.

---

<details>
<summary>Details</summary>

- The `pg-backrest` package itself (`nix/packages/default.nix`, a plain passthrough of `nixpkgs.legacyPackages.<system>.pgbackrest`) already landed on `develop` independently via #1714. This PR does not change it.
- This PR originally added a from-source meson build of pgbackrest (`nix/packages/pgbackrest.nix`) to pin an exact version and mesonFlags; that was reverted in favor of the nixpkgs passthrough, so no from-source recipe remains.
- Version consistency across a cluster is handled independently of pgbackrest's own version pin: `ansible/tasks/setup-pgbackrest.yml` installs via `nix profile install github:supabase/postgres/{{ git_commit_sha }}#pg-backrest`, locked to the exact postgres repo commit. `Dockerfile-multigres` consumes the same flake output by name.
- The nixpkgs pin currently resolves to pgbackrest 2.57.0. Bumping to 2.58.0 rides on the `flake.lock` update automation that #2281 (PSQL-1388) is fixing, and is tracked separately in [MUL-129](https://linear.app/supabase/issue/MUL-129/upgrade-to-pgbackrest-2580); informational only, not a merge blocker for this PR, since this diff makes no `flake.lock` change.
- `nix-eval / eval` is a required check that evaluates the `checks` output this PR adds `pg-backrest` to. If a future nixpkgs bump ever breaks pgbackrest at eval time (not just build time), that shows up as `nix-eval` going red on an unrelated PR. This repo has hit that failure mode once already (see `c261fda8`, this branch's own revert of a nixpkgs bump for the same reason).

</details>

---

<details>
<summary>Testing</summary>

- [x] `nix build .#pg-backrest` succeeds
- [x] `nix build .#checks.<linux-system>.pg-backrest` succeeds

</details>

---

<details>
<summary>Misc</summary>

Resolves [INDATA-520](https://linear.app/supabase/issue/INDATA-520/create-a-supabase-nix-pkg-for-pgbackrest), using the plain nixpkgs passthrough rather than a from-source build.

</details>


brainrake (changes requested):
https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/pg/pgbackrest/package.nix

- Latest pgbackrest 2.58.0 is in nixpkgs. Use this please.
- If not good and need eg. different build options, lease override nixpkgs package instead of copying it verbatim, unless there is a blocker. Lower maintenance burden.
- Add pgbackrest src repo as a non-flake input to the flake. This way hashes are managed using default tools in the lockfile instead of manually.

brainrake:
If we need to update nixpkgs, lets prioritize that, it needs to be done in any case. Doing the chores is preferable to deep special casing. Special casing will only make it more difficult to stay up to date, which we need to do anyway to bring in stack wide security updates.

hunleyd:
Depends on #2281 (PSQL-1388): the nixpkgs bump this PR's pgbackrest pin rides on needs the `update-flake-lock` workflow fixed first. Leaving this in draft until #2281 lands.

hunleyd:
@claude review

PR #2271, opened 2026-07-14
Check for 2 GB free space before PG upgrade

hunleyd:
## Summary
- `initiate.sh` realizes the new pg_upgrade Nix store path onto `/` before touching the mounted data disk, so a tight root partition fills mid-upgrade instead of failing fast at the start.
- Adds `check_free_space` to `common.sh`: checks available space on `/` via `df` and returns non-zero if below a threshold.
- Calls it at the top of `initiate_upgrade()` in `initiate.sh` with a 2 GiB threshold, before any mkdir/nix work. A failure there hits the existing `trap cleanup ERR` path, same as every other failure in this script — postgresql restarts and status is set to `failed`.

Resolves INDATA-641

## Test plan
- [x] `bash -n` syntax check on both modified scripts
- [x] Manually exercised `check_free_space` with a trivial threshold (pass) and an absurd threshold (fail) against real `df` output
- [x] Run the existing pg_upgrade integration test suite (`tests/pg_upgrade`)


ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
```diff
@@ -44,6 +44,27 @@ function ship_logs {
...
+	# -P forces POSIX single-line output; plain `df -k` wraps onto a second
+	# line when the source device name is long, which would misalign NR==2.
+	available_kb=$(df -Pk / | awk 'NR==2 {print $4}')
+
+	if ! [[ $available_kb =~ ^[0-9]+$ ]]; then
+		echo "ERROR: could not determine free space on /; aborting upgrade."
+		return 1
+	fi
+
+	if [ "$available_kb" -lt "$required_kb" ]; then
```
mmlb:
```suggestion
	if (( available_kb < required_kb )); then
```

Lets use bash math mode since its nicer to read and this script is specified to be bash so not required to be posix sh.

mmlb (dismissed):

mmlb (commented):
Yep LGTM

mmlb (approved):

mmlb:
@hunleyd do we still want to merge this in lieu of the linear update? Will leave it to you to merge if so or close otherwise.

hunleyd:
@mmlb yes, still want this merged. Dug into it: admin-api's \`getFreeDiskSpace\` only gates the download step (requires free space ≥ 2x the object being downloaded), not the actual upgrade — the failure mode this PR targets is \`nix-store\` realizing new derivations onto \`/\` *during* \`pg_upgrade\`, well after the download completes. Different check, different point in the pipeline, no overlap. The Linear closure on INDATA-641 was specifically about not vendoring a copy of these scripts into Salt (supabase/salt#685) — this PR (the actual source of truth) still stands.

Also pushed a fix: \`cleanup()\`'s \`retry 3 umount $MOUNT_POINT\` had no \`|| true\`, so when \`check_free_space\` fails before the disk is ever mounted, the ERR trap's own umount call was aborting under set -e before writing the status file — leaving it stuck at "running" forever on exactly the failure path this PR adds. Fixed and reproduced/verified with a minimal harness.

Enabling auto-merge now that CI's running clean.

PR #2272, opened 2026-07-14
Whole lot of ansible clean up

mmlb:

Maintenance

Lots of dead weight from dead code such as debpkg_mode, qemu_mode.
Using 2 variables nixpkg_mode and stage2_nix to represent 2 states when 1 variable will do.
Different ways to conditionally run ansible tasks.
Bunch of unverified downloads... :clock12: :bomb: if you ask me...

## What is the new behavior?

Dead code deleted
Lots of clean ups of tasks themselves
All downloads are verified via checksum or signatures (including d/l of signature files)
Added script with 2 modes: update artifact version & checksum, verify checksums are all still good

## Additional context

More stuff to come in future PRs.
Closes PSQL-1539

### DIFF
--- a/ansible/files/adminapi.service.j2
+++ b/ansible/files/adminapi.service.j2
@@ -14,7 +14,7 @@ Restart=always
 RestartSec=3
 TimeoutStopSec=10
 Environment="AWS_USE_DUALSTACK_ENDPOINT=true"
-{% if qemu_mode is defined and qemu_mode %}
+{% if qemu %}
 Environment="AWS_SDK_LOAD_CONFIG=true"
 {% endif %}
 StandardOutput=journal
--- a/ansible/files/gotrue.service.j2
+++ b/ansible/files/gotrue.service.j2
@@ -90,7 +90,7 @@ Environment=GOTRUE_RELOADING_POLLER_ENABLED=false
 # ensures only 1 reload operation occurs during a burst of config updates.
 Environment=GOTRUE_RELOADING_GRACE_PERIOD_INTERVAL=2s
 
-{% if qemu_mode is defined and qemu_mode %}
+{% if qemu %}
 # v3 does not use filesystem notifications for config reloads.
 Environment=GOTRUE_RELOADING_NOTIFY_ENABLED=false
 {% else %}
--- a/ansible/files/postgres_exporter.service.j2
+++ b/ansible/files/postgres_exporter.service.j2
@@ -6,7 +6,7 @@ StartLimitBurst=0
 
 [Service]
 Type=simple
-ExecStart=/opt/postgres_exporter/postgres_exporter --disable-settings-metrics --extend.query-path="/opt/postgres_exporter/queries.yml" --disable-default-metrics --no-collector.locks --no-collector.replication --no-collector.replication_slot --no-collector.stat_bgwriter --no-collector.stat_database --no-collector.stat_user_tables --no-collector.statio_user_tables --no-collector.wal {% if qemu_mode is defined and qemu_mode %}--no-collector.database {% endif %}
+ExecStart=/opt/postgres_exporter/postgres_exporter --disable-settings-metrics --extend.query-path="/opt/postgres_exporter/queries.yml" --disable-default-metrics --no-collector.locks --no-collector.replication --no-collector.replication_slot --no-collector.stat_bgwriter --no-collector.stat_database --no-collector.stat_user_tables --no-collector.statio_user_tables --no-collector.wal {% if qemu %}--no-collector.database {% endif %}
 
 User=postgres
 Group=postgres
--- a/ansible/files/supabase_admin_agent_config/supabase-admin-agent_salt.service.j2
+++ b/ansible/files/supabase_admin_agent_config/supabase-admin-agent_salt.service.j2
@@ -6,7 +6,7 @@ Requires=local-fs.target
 
 [Service]
 Type=oneshot
-{% if qemu_mode is defined and qemu_mode %}
+{% if qemu %}
 ExecStart=/opt/supabase-admin-agent/supabase-admin-agent --config /opt/supabase-admin-agent/config.yaml salt --apply --store-result --targets-key  targetsv3-main.yaml
 User=root
 Group=root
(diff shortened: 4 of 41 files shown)


hunleyd (approved):

mmlb:
Turns out qemu mode is still necessary, its used to build the multigres k8s images. Will rebase this PR once #2274 is merged. Funnily enough I bet it'll be easier to re-introduce the qemu ansible tasks on top of these commits instead of never removing it. Will see...

hunleyd:
## Review

### Overview

This PR cleans up the AMI-build Ansible playbook:
- Collapses two overlapping boolean flags (`nixpkg_mode`, `stage2_nix`) into one (`stage2`), plus a separate `qemu` flag, and drops the `is defined` indirection everywhere.
- Moves every `when:` condition from the bottom of each task to the top (pure style/readability change, no behavior difference).
- Centralizes every downloaded artifact's URL + checksum into a single `*_artifacts` table in `ansible/vars.yml`, converts every checksum to sha256, and adds checksums to several downloads that previously had none (nginx, PostgREST's GPG key, Vector, admin-mgr, adminapi, AWS CLI).
- Adds `nix/tools/update-ansible-artifacts`, a script (exposed as a flake app) that can bump a tool's version + refetch its checksum, or verify all checksums in `vars.yml` still match upstream.
- Removes several redundant `set_fact: arch` tasks now that `platform` (amd64/arm64) is used directly as a dict key into the new artifacts table.

### Issues found

**1. Likely bug: `surrogate-bootstrap-nix.sh` now passes `qemu: true` for what looks like the real (non-QEMU) AMI build path.**
`ebssurrogate/scripts/surrogate-bootstrap-nix.sh:344` — this script builds the actual production AMI using the ebssurrogate/chroot technique (as opposed to `qemu-bootstrap-nix.sh`, which is the QEMU-based build/test path). Before this PR, this invocation never set `qemu_mode` at all, so every `qemu_mode is defined` check evaluated false. The new line sets `"qemu":true` — same value used by the actual QEMU script. If unintentional, this flips several conditionals on real AMI builds: skips installing `ec2-instance-connect` (`when: not qemu`), skips the IPv6 NDisc fix, and later (in the stage-2 pass) would skip the `/data` production mount task and the supascan baseline validation, both gated on `not qemu`. Worth confirming whether this was intentional before merge.

**2. Dead code left behind, contradicting the PR's own stated goal.**
`ansible/tasks/internal/supabase-admin-agent.yml` still has:
```yaml
- name: Setting arch (amd64)
  when: platform == "amd64"
  set_fact:
    arch: "amd64"

- name: Setting arch (arm)
  when: platform == "arm64"
  set_fact:
    arch: "arm64"
```
`arch` is no longer used anywhere in this file — the symlink task that used to reference `{{ arch }}` was updated to use `{{ platform }}` directly. The identical redundant `set_fact` pair was correctly removed from `admin-api.yml` and `admin-mgr.yml` in this same PR, but this one was missed.

**3. Removed the `update` / `update-only` tags from the platform-detection task — possible breakage for tag-scoped runs.**
The "Set the platform arch as a fact" task moved from `tasks/setup-system.yml` (tagged `update`, `update-only`) into `playbook.yml`'s `pre_tasks`, untagged. Ansible tag filtering applies to `pre_tasks` the same as any other task — an untagged task is skipped whenever `--tags <something>` is passed (unless it carries the special `always` tag). No task in this repo consumes that tag itself, which suggests it exists for an external caller doing fleet-wide config pushes (e.g. via `supabase-admin-agent`/`admin-mgr`) using `ansible-playbook --tags update-only`. If so, `platform` would come back undefined on every tag-scoped run post-merge. Worth confirming with whoever owns that fleet-update path.

**4. Two checksums are excluded from the new verification tooling.**
`ansible/tasks/setup-postgrest.yml`'s PostgreSQL APT GPG key and `ansible/tasks/internal/install-salt.yml`'s Salt GPG key both got inline `checksum: sha256:...` added (good — previously unverified), but they weren't added to `vars.yml`'s `*_artifacts` structure. `nix/tools/update-ansible-artifacts/update-ansible-artifacts.sh`'s `verify` mode only walks `vars.yml` for maps with `url` + `checksum` keys, so these two checksums are invisible to the new "verify everything still matches upstream" tool.

**5. Minor inconsistency: `postgrest_release` embeds a `v` prefix; every sibling variable doesn't.**
Every other `*_release` var (`adminapi_release: 0.112.1`, `gotrue_release: 2.190.0`, etc.) is a bare version string, with `v` added inside the URL template. `postgrest_release: v14.15` bakes the `v` into the value itself. Anyone running `nix run .#update-ansible-artifacts postgrest 14.16` will get a broken URL unless they remember this one tool needs `v14.16` instead.

### What's good here

- The core motivation — unverified downloads — is fully addressed: every `get_url` task in the diff now carries a checksum, and the sha1→sha256 conversions plus the new `verify` mode give this real staying power rather than a one-time fix.
- Collapsing `nixpkg_mode`/`stage2_nix` into one `stage2` boolean removes a real footgun.
- Moving `when:` to the top of each task is a pure convention change and applied consistently.

mmlb:
> ### Issues found
> 
> **1. Likely bug: `surrogate-bootstrap-nix.sh` now passes `qemu: true` for what looks like the real (non-QEMU) AMI build path.**

copy pasted and forgot to change to false, fixed now.
 
> **2. Dead code left behind, contradicting the PR's own stated goal.** 

Rebase bug, fixed now.

> **3. Removed the `update` / `update-only` tags from the platform-detection task — possible breakage for tag-scoped runs.** The "Set the platform arch as a fact" task moved from `tasks/setup-system.yml` (tagged `update`, `update-only`) into `playbook.yml`'s `pre_tasks`, untagged. Ansible tag filtering applies to `pre_tasks` the same as any other task — an untagged task is skipped whenever `--tags <something>` is passed (unless it carries the special `always` tag). No task in this repo consumes that tag itself, which suggests it exists for an external caller doing fleet-wide config pushes (e.g. via `supabase-admin-agent`/`admin-mgr`) using `ansible-playbook --tags update-only`. If so, `platform` would come back undefined on every tag-scoped run post-merge. Worth confirming with whoever owns that fleet-update path.

This is fine, nothing uses ansible at runtime and we don't ever pass any tags when calling ansible from this repo. This was added ~5y ago and no longer relevant.

> **4. Two checksums are excluded from the new verification tooling.** 

Fixed.
 
> **5. Minor inconsistency: `postgrest_release` embeds a `v` prefix; every sibling variable doesn't.** 

I wanted to avoid the `keep this a string` by having something to definitely force it to string instead of float but decided best to just keep the `"..."` and comment.

hunleyd (approved):

PR #2274, opened 2026-07-14
Fix qemu/k8s image builds

mmlb:

Bug fix

Qemu build fails to build due to requiring /data mount, which does not exist during qemu builds.

## What is the new behavior?

Couple of new things:
1. Don't require the /data mount for qemu builds
2. Always run the qemu workflow whenever its inputs change, but only push under same conditions as previous runs.


qemu-arm64-nix.pkr.hcl
```diff
@@ -1,50 +1,22 @@
...
 
 variable "postgres-version" {
-  type = string
+  type    = string
   default = ""
 }
 
-variable "postgres-major-version" {
-  type = string
+variable "postgres_major_version" {
```
brainrake, line 10: This variable was renamed to `postgres_major_version` (underscore), but the qemu workflow still writes the old hyphenated name into `common-nix.vars.pkr.hcl`:

```yaml
# .github/workflows/qemu-image-build.yml:73
echo 'postgres-major-version = "'$POSTGRES_MAJOR_VERSION'"' >> common-nix.vars.pkr.hcl
```

Packer treats `postgres-major-version` as an **undeclared** variable, so `postgres_major_version` keeps its `""` default and `POSTGRES_MAJOR_VERSION` reaches `qemu-bootstrap-nix.sh` empty (`postgresql_version=postgresql_`, `psql_version=psql_`, …) — the qemu image ends up built against no PG major version. Since the sibling `postgres-version` var and the workflow both use hyphens, simplest is to keep this one hyphenated too. Revert here **and** at the reference on line 105 (`${var.postgres-major-version}`). Alternatively, update the workflow's line 73 to the underscore name.

```suggestion
variable "postgres-major-version" {
```

ansible/manifest-playbook.yml
```diff
@@ -1,120 +0,0 @@
-- hosts: localhost
```
brainrake, line 1: This deletes software-manifest generation entirely — `manifest-playbook.yml` plus the "Upload software manifest to s3" steps in `ami-release-nix.yml` (and the removed `ami-release-nix-single.yml`). Those uploaded `software-manifest-{arch}.json` to `s3://…/manifests/postgres-<version>/`. Can you confirm nothing downstream consumes those manifests (release tooling, upgrade pipeline, dashboards) before dropping them? It's an externally-visible artifact going away, separate from the `debpkg_mode` cleanup.

mmlb: Yep this hasn't been used internally for almost 2 years now.

qemu-arm64-nix.pkr.hcl
mmlb: Thanks for pointing this out, I've been running this locally and it works on my machine and I was going to start looking at why. I changed things up so it runs from a nix env (like build-ami does) and that did the right thing. Will leave that for a separate PR though.

ansible/manifest-playbook.yml
mmlb: To give some more info here, I pinged @darora internally and he confirmed that it was used by just one tool/process but has not been used since ~ end of July 2024. My own github search turned up empty.

samrose (changes requested):
qemu workflow's ungated AWS steps

.github/workflows/qemu-image-build.yml now triggers on push to any branch (path-filtered), where before it was gated to develop + release/*. The stated intent is "always build to test, only push under the old conditions." The two docker push steps are correctly gated, but the steps before them are not:

- configure aws credentials - staging (assumes CONTROL_PLANE_DEV_ROLE)
- Login to Amazon ECR (dev)
- configure aws credentials - prod (assumes CONTROL_PLANE_PROD_ROLE)
- Login to Amazon ECR (prod)

These four now run on every changed-path push to any feature branch. I confirmed they are not needed for the build itself: Dockerfile-kubernetes is FROM alpine:3.23 and just ADDs the qemu output (./output-cloudimg/packer-cloudimg) — no private-ECR pull. So AWS is only needed for the push.

Consequences on a non-develop/release branch push:
- If the OIDC role trust policies are branch-scoped (common), configure aws credentials fails → the whole job fails → this defeats the PR's own goal of "make sure qemu builds keep working via CI test builds" precisely on the feature branches meant to benefit.
- If the trust policies are only repo-scoped, it works but needlessly assumes the prod role on arbitrary branches — a mild security/noise smell.

Suggested fix: put the same gate on the four AWS/ECR steps (or lift auth+push into a separate job gated at the job level):

      - name: configure aws credentials - staging
        if: ${{ github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/heads/release/') }}

samrose (approved):

mmlb:
> qemu workflow's ungated AWS steps
> 
> .github/workflows/qemu-image-build.yml now triggers on push to any branch (path-filtered), where before it was gated to develop + release/*. The stated intent is "always build to test, only push under the old conditions." The two docker push steps are correctly gated, but the steps before them are not:
> 
>     * configure aws credentials - staging (assumes CONTROL_PLANE_DEV_ROLE)
> 
>     * Login to Amazon ECR (dev)
> 
>     * configure aws credentials - prod (assumes CONTROL_PLANE_PROD_ROLE)
> 
>     * Login to Amazon ECR (prod)
> 
> 
> These four now run on every changed-path push to any feature branch. I confirmed they are not needed for the build itself: Dockerfile-kubernetes is FROM alpine:3.23 and just ADDs the qemu output (./output-cloudimg/packer-cloudimg) — no private-ECR pull. So AWS is only needed for the push.
> 
> Consequences on a non-develop/release branch push:
> 
>     * If the OIDC role trust policies are branch-scoped (common), configure aws credentials fails → the whole job fails → this defeats the PR's own goal of "make sure qemu builds keep working via CI test builds" precisely on the feature branches meant to benefit.
> 
>     * If the trust policies are only repo-scoped, it works but needlessly assumes the prod role on arbitrary branches — a mild security/noise smell.
> 
> 
> Suggested fix: put the same gate on the four AWS/ECR steps (or lift auth+push into a separate job gated at the job level):
> 
> ```
>   - name: configure aws credentials - staging
>     if: ${{ github.event_name == 'workflow_dispatch' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/heads/release/') }}
> ```

Fixed

PR #2281, opened 2026-07-15
ci: fix update-flake-lock workflow (app token + update-flake-lock action)

brainrake:
The `update-flake-lock` workflow has failed on every run. Two issues:

1. **Duplicate `Authorization` header → HTTP 400.** `actions/checkout@v6` persists a credential auth header that collides with the one the PR-creation step sets.
2. **`GitHub Actions is not permitted to create or approve pull requests`.** The org disables that policy, so `GITHUB_TOKEN` can't open the PR regardless of workflow `permissions`.

## Fix

- Fix auth header.
- Create Supabase-owned GitHub App with **Contents: R/W** and **Pull requests: R/W**,
- Authenticate with a **GitHub App token** via `actions/create-github-app-token`. This also means PRs trigger CI (unlike `GITHUB_TOKEN`-created PRs).

## Extra

- Switch to `DeterminateSystems/update-flake-lock` for better ergonomics (handles the `nix flake update` + PR in one step).


brainrake:
Test run: https://github.com/supabase/postgres/actions/runs/29878589209
Resulting PR: https://github.com/supabase/postgres/pull/2297

brainrake:
@hunleyd we still need to actually update nixpkgs in https://github.com/supabase/postgres/pull/2281 to close PSQL-1388, this pr only fixes the auto update pr bot.

brainrake:
Latest workflow run: https://github.com/supabase/postgres/actions/runs/30595099779
Resulting PR:
https://github.com/supabase/postgres/pull/2328

mmlb:
I don't think we should run this workflow if we aren't confident in the update from CI tests its just going to be noise until we run manual tests and merge.

brainrake:
Agreed, I'll make it manually triggered instead of scheduled until then.

brainrake:
Ok it''s not scheduled. Please re-review @mmlb 
Although it only updates the existing PR so we know what we're up against when we open it. It's fine.

.github/workflows/update-flake-lock.yml
```diff
@@ -12,34 +13,30 @@ jobs:
...
-            Automated monthly update of critical flake inputs:
-            - `nixpkgs` (nixos-unstable channel)
-            - `rust-overlay`
-            - `multigres`
-            - `nix-eval-jobs`
-          branch: auto-update-flake-lock
-          base: develop
-          labels: |
+          token: ${{ steps.app-token.outputs.token }}
+          inputs: nixpkgs rust-overlay multigres nix-eval-jobs
```
mmlb: do we need to specify inputs?

brainrake: It was in the past and I left it in place thinking defensively. But that's probably the wrong approach. This workflow should update all inputs and we should explicitly pin any inputs we need to. Fixing.

brainrake:
@mmlb good to go

mmlb (approved):

PR #2292, opened 2026-07-21
feat(pgbackrest): schedule process-max tuning via systemd timer

hunleyd:
## Summary

- `pgbackrest tune` recomputes process-max from the host's live vCPU count but was only ever invoked at `pgbackrest enable`/`setup-replica` time, so values went stale after a compute resize.
- Adds a `supabase-admin-agent_pgbackrest_tune` timer/service pair (10-minute cadence, cloned from the existing salt timer) that re-runs tune on a schedule.
- Service is gated with `ConditionPathExists=/etc/pgbackrest/conf.d/pgbackrest.conf` so it skips cleanly (not failed) on instances without pgBackRest enabled — no coordination needed with whatever enables/disables the feature. Timer is enabled by default since that condition is the actual safety gate.
- Updates the `prod-deployed` audit-spec baselines for the two new unit files.

The originally proposed approach (call `pgbackrest tune` directly from the `admin-mgr` resize-compute flow) doesn't apply: `admin-mgr` has no resize-compute code, and the actual flow (`platform`'s worker) has no remote-exec path into the instance to invoke an instance-local command. A scheduled timer sidesteps that entirely and also covers the "cron fallback" question raised in the ticket — it's both the trigger and the fallback.

Resolves INDATA-997

## Test plan

- [ ] Build a test AMI/QEMU image via this repo's build process
- [ ] Boot with pgBackRest enabled; confirm `supabase-admin-agent_pgbackrest_tune.timer` is `active (waiting)` and `computed_globals.conf` matches the host's actual vCPU count
- [ ] Boot without pgBackRest enabled; confirm the service reports condition-skipped, not failed
- [ ] Run a resize-compute flow against a real instance (`/smokeit`) and confirm `computed_globals.conf` updates within one timer tick after the post-resize reboot

### DIFF
--- a/ansible/files/supabase_admin_agent_config/supabase-admin-agent_pgbackrest_tune.service.j2
+++ b/ansible/files/supabase_admin_agent_config/supabase-admin-agent_pgbackrest_tune.service.j2
@@ -1,0 +1,17 @@
+[Unit]
+Description=Recompute pgBackRest process-max via supabase-admin-agent pgbackrest tune
+After=network.target local-fs.target tmp.mount
+Requires=local-fs.target
+# Skips cleanly (not a failure) on instances where pgBackRest isn't enabled.
+ConditionPathExists=/etc/pgbackrest/conf.d/pgbackrest.conf
+
+[Service]
+Type=oneshot
+ExecStart=/opt/supabase-admin-agent/supabase-admin-agent --config /opt/supabase-admin-agent/config.yaml pgbackrest tune
+User=supabase-admin-agent
+Group=supabase-admin-agent
+StandardOutput=journal
+StandardError=journal
+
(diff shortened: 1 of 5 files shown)


hunleyd:
Closing: the systemd-timer approach here only ever reaches instances built from a new AMI going forward. It doesn't touch the existing fleet, which is the actual population with stale pgBackRest process-max — that's a much bigger population than a new-AMI-only fix helps. The real fleet-wide convergence mechanism already exists (the salt state.apply loop pulled from S3 every 10 minutes, running on both v2 and v3), so the fix belongs there instead: a salt formula that shells out to the existing `pgbackrest tune` command. Tracking that as the real INDATA-997 fix.

PR #2299, opened 2026-07-22
ci: add package version diff comment on PRs

brainrake:
Adds a package version diff comment to the `nix-build.yml` (`Nix CI`) workflow.

For each of `x86_64-linux`, `aarch64-linux`, `aarch64-darwin`: evaluates every derivation's `drvPath` under `legacyPackages.<system>` on the PR's head and base commits (pure evaluation — no network, no build). If nothing differs, that's reported directly with zero further work. If something differs, only the changed/added/removed attrs are built (substituted from the shared cache) and diffed with `nvd` — the rest of the closure (typically 700+ paths) is never touched.

Results are posted as a single collapsed comment per PR (one `<details>` per system), updated in place on subsequent pushes.

**Example — no differences:** see this PR's own comment below.
**Example — with a real diff (added package):** https://github.com/supabase/postgres/pull/2306 (closed test PR, kept for reference; also used to verify timing — see PR discussion for the before/after breakdown)

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -283,17 +283,18 @@ jobs:
         uses: ./head/.github/actions/nix-install-self-hosted
 
       # nix-eval's packages_matrix only lists *uncached* packages, so it
-      # cannot be used to decide what to diff. Instead, collect every
-      # derivation under legacyPackages.<system> (recursing into attrsets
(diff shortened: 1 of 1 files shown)


brainrake:
Closing. Too much complexity added, not enough value over existing tools.

PR #2300, opened 2026-07-22
fix: cover pg_cron legacy extversion strings with alignment migration and alias scripts

brainrake:
## Summary
Supersedes #1790 

- Adds a `legacyVersions` table (`nix/ext/pg_cron/default.nix`) mapping every bare extversion string pg_cron has shipped under (`1.0`–`1.6`) to its schema-identical canonical version, generating one empty alignment migration per entry — restoring a resolvable `ALTER EXTENSION pg_cron UPDATE` path for projects still recording one of these old strings.
- The nixos test consumes that same table via `passthru` and asserts each legacy string reaches latest, rather than discovering upgrade edges from disk.

Root cause / customer impact: [supabase/supabase#46669](https://github.com/supabase/supabase/issues/46669) — in-place PG upgrades fail at the completion step with `extension "pg_cron" has no update path from version "1.6" to "1.6.4"` for any project whose pg_cron extension is still recorded at the bare `1.6` version.

## Test plan
- [x] ci
- [x] new nixos tests testing upgrades from all starting versions

Fixes PSQL-1414


brainrake:
Replaced 21 `ln -s` aliases with a single legacyVersions table (legacy version -> semver), generating one empty alignment migration per entry. The table is shared via `passthru` so the nixos test consumes it directly instead of rediscovering the alias graph from the filesystem — `check_all_upgrade_edges` (discovery + reachability + two-pass exerciser) is replaced by `check_legacy_versions_upgrade_to_latest`, which patches `pg_extension.extversion` to each legacy string and asserts a bare `ALTER EXTENSION UPDATE` lands on latest.

Removed legacy->legacy and semver->legacy migration versions since those should never be used, only `legacy->semver`.

Also fixes a real bug the old renames introduced: `1.3--1.4` was renamed to `1.3.1--1.4.2`, silently dropping the `cron.jobid_seq/runid_seq GRANT`s that only exist in upstream's `1.4--1.4-1` script. Now renames to `1.3.1--1.4.0` with a new `1.4.1--1.4.2` connector reconnecting the chain.

imor (approved):

PR #2302, opened 2026-07-22
feat(nix): expose per-version extension derivations via passthru

brainrake:
Adds `passthru.perVersion` (`version -> single-version derivation`) to each multi-version extension (18 files, one line each).

The per-version attrset already existed inline as the input to each extension's `buildEnv` — this just surfaces it instead of discarding it. No change to any `buildEnv` output or extension identity; `exts.<name>` is still the same derivation, now carrying `.perVersion."<version>"` whose `outPath` is the lean single-version store path.

This is the prerequisite for the extension catalog in the profile-based version-switching work — it lets the catalog reference a single `(extension, version)` store path directly, with no buildEnv copy and no IFD.

Part of [PSQL-1530](https://linear.app/supabase/issue/PSQL-1530/use-profiles-for-multi-version-extensions).

### DIFF
--- a/nix/ext/hypopg.nix
+++ b/nix/ext/hypopg.nix
@@ -98,6 +98,7 @@ buildEnv {
   '';
 
   passthru = {
+    perVersion = lib.mapAttrs (name: value: build name value.hash) versionsToUse;
     versions = versionsBuilt;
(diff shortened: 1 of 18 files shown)


brainrake:
Merged into #2317 (retargeted to develop, now includes this commit).

PR #2306, opened 2026-07-22
testing: version-diff

brainrake:
DO NOT MERGE, TESTING.

Verifies the drvPath-scoped version-diff job (only realizes/diffs attrs whose drvPath actually changed, instead of the whole `legacyPackages` closure). Adds a trivial dummy package to trigger a real "added" diff.

Base is `ci/nix-version-diff` so the version-diff workflow changes are present without duplicating that diff here.

### DIFF
--- a/nix/packages/postgres.nix
+++ b/nix/packages/postgres.nix
@@ -250,6 +250,13 @@
     in
     {
       packages = binPackages;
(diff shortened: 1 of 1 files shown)


brainrake:
Test PR done — drvPath-scoped version-diff verified working, ~55% faster than the full-closure approach. Closing.

PR #2308, opened 2026-07-23
chore: remove nixConfig from flake.nix

brainrake:
## Why
`nixConfig.extra-substituters` only applies to trusted users, so everyone else sees:

> warning: ignoring untrusted substituter '...', you are not a trusted user

That warning invites people to add themselves to `trusted-users`, which is actually a local privilege-escalation vector ([Nix manual](https://nix.dev/manual/nix/stable/command-ref/conf-file#conf-trusted-users), see also SEC-894), not a config gap to fix.

## What
Remove `nixConfig` from `flake.nix`. (Configure substituters via `nix.conf`/nix-darwin instead.)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,10 +1,9 @@
 {
-  description = "Prototype tooling for deploying PostgreSQL";
-  nixConfig = {
-    # Skip rebuilding all the packages, download instead.
-    # See nix/docs/binary-cache.nix to set it up.
-    extra-substituters = [ "https://nix-postgres-artifacts.s3.amazonaws.com" ];
-  };
+  description = "Tools for deploying PostgreSQL and extensions";
+
+  # Skip rebuilding all the packages, download instead.
+  # See nix/docs/binary-cache.nix to set it up.
+
   inputs = {
     devshell.url = "github:numtide/devshell";
     devshell.inputs.nixpkgs.follows = "nixpkgs";
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ The project uses Nix as its build system, which provides:
 
 ## Common Tasks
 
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -1,10 +1,5 @@
 {
   description = "Prototype tooling for deploying PostgreSQL";
-  nixConfig = {
-    # Skip rebuilding all the packages, download instead.
-    # See nix/docs/binary-cache.nix to set it up.
```
mmlb, line 5: Might be good to keep the comment, it makes sense to have the doc pointer here since its pretty likely to have someone read the flake at some point.

Maybe we do:
```suggestion
  # Skip rebuilding all the packages, download instead.
  # See nix/docs/binary-cache.nix to set it up.
  # nixConfig.extra-substituters = [ "https://nix-postgres-artifacts.s3.amazonaws.com" ];
  # nixConfig.extra-trusted-public-keys = [ "nix-postgres-artifacts:dGZlQOvKcNEjvT7QEAJbcV6b6uk7VF/hWMjhYleiaLI=" ];
```

mmlb (approved):
Just a quick comment, will approve either way as this is definitely a better setup.

brainrake:
@mmlb i added the comment back. It feels out of place. It' already in the readme. I guess we can remove it later if our docs get smaller/better.

mmlb:
> @mmlb i added the comment back. It feels out of place. It' already in the readme. I guess we can remove it later if our docs get smaller/better.

@brainrake yeah that was the issue in my mind. Right now the docs are too much of a wall of text with too much info so I figured its best to make sure we have pointers to docs in places people would likely be looking at.

PR #2309, opened 2026-07-23
ci: install AMI postgres profile via nix-env --set

brainrake:
Builds on #2283 (site-profiles).

## Summary
- Add `postgres-env-{15,17,orioledb-17}` packages (`nix/packages/postgres-env.nix`) that bundle everything the AMI build installs for the postgres user: `psql_<v>/bin`, `pg_prove`, `supabase-groonga`, `postgresql_<v>_src`, `postgresql_<v>_debug` (Linux only), and `gatekeeper` (Linux, non-PG15) for `pam_jit_pg.so`.
  - Terminology: a postgres **env** is a fixed derivation with one set of packages and no history (nix-env sense). It gets installed into the postgres user's nix **profile** via `nix-env --set`, which is what gives it history/generations and a "latest" pointer.
- `ansible/tasks/stage2-setup-postgres.yml`: replace the 5-item `nix profile install` loop with `nix build --print-out-paths` (registered) followed by `nix-env --set` to install the resolved env into the profile.

Switching to `nix-env --set` avoids mixing `nix profile install`'s manifest format with `nix-env`'s on the same default profile path.

## Test plan
- [x] `nix eval .#packages.aarch64-darwin.postgres-env-{15,17,orioledb-17}.name` — all three packages evaluate
- [x] `nix build .#postgres-env-{15,17,orioledb-17}` — all three build successfully; verified `bin/psql`, `bin/pg_prove` present, `gatekeeper`'s `pam_jit_pg.so` present for 17/orioledb-17
- [x] `nix-env --profile <scratch> --set <resolved store path>` — profile correctly set to a single new generation, `psql --version` works from it
- [x] CI: `test_jit_pam_module_installed` (testinfra) passes for PG17/orioledb-17
- [x] Real AMI build (packer/ansible)
- [ ] Smoke tests (supadev, in progress)


brainrake:
Note: gatekeeper is baked into `postgres-env` here as a stopgap to keep CI green. Per [PSQL-1408](https://linear.app/supabase/issue/PSQL-1408/ami-pre-populate-site-profile-at-bake-drop-pam-store-symlink), gatekeeper (and the PAM store-symlink hack) should ultimately move to the `site-env` profile instead — worth reconciling once that lands.

ansible/tasks/stage2-setup-postgres.yml
```diff
@@ -132,11 +135,6 @@
         - name: Install gatekeeper if not pg15
```
mmlb, line 135:
```suggestion
        - name: Setup gatekeeper
```

ansible/tasks/stage2-setup-postgres.yml
```diff
@@ -62,17 +62,20 @@
   when: stage2
   become: true
   block:
-    - name: Install packages from nix binary cache
+    - name: Resolve postgres env store path
```
mmlb, line 65:
```suggestion
    - name: Resolve postgres env from nix binary cache
```

mmlb (approved):
couple of small nits that would be nice to have but not worth blocking, otherwise lgtm

PR #2315, opened 2026-07-23
chore: enable extension version restriction (warn) and cut AMIs

utkarash2991:
- Sets `supautils.restrict_extension_versions = 'warn'` in `supautils.conf.j2` (GUC shipped in supautils v3.4.0, already on develop via #2311): CREATE/ALTER EXTENSION version clauses from non-superuser roles are ignored with a WARNING and `default_version` is installed. Superusers and `supabase_admin` (via `privileged_extensions_superuser`) are exempt.
- Bumps `postgres_release` to cut fresh AMIs.
- **Tests the conf in the same PR**: `nix/tests/sql/supautils_restrict_versions.sql` runs in the pg_regress harness, which loads the rendered `supautils.conf.j2` with migrations applied (via `start-postgres-server`) — asserts the warn behavior as the non-superuser `postgres` role and the warning-free exemption for `supabase_admin`. (Skipped on the CLI variant, which lacks the `postgres` role.) The supautils regress suite covers the GUC logic itself; this covers the platform wiring.

`warn` is rollout phase 1: fleet log warnings measure explicit-version usage before the separately-announced `error` flip. Public announcement (effective 2026-08-05): https://github.com/supabase/changelog/pull/128 — companion supabase/salt#717 covers the existing fleet.

PSQL-1159

PR #2316, opened 2026-07-24
test: platform-policy test harness improvements

utkarash2991:
Two improvements for testing platform behavior (supautils policy against the rendered config and real roles). Independent of the extension-version-restriction rollout (#2315), but gives tests like its future policy test a proper home.

**1. testinfra split.** `test_ami_nix.py`  → `conftest.py` (shared session-scoped EC2 fixture + helpers) and six domain modules (postgrest / postgres / jit_pam / overrides / mitigations / apparmor), all 27 tests preserved. One EC2 instance still serves the whole run. The workflow now runs `pytest testinfra/`, so new AMI tests get their own module instead of growing one file.

**2. Platform-policy test pattern in nix/tests.** `nix/tests/README.md` documents what was implicit: the pg_regress suite runs against the rendered `supautils.conf.j2` with real migrations/roles (via `start-postgres-server`), so supautils policy is testable as the actual platform roles with `SET ROLE` — conf changes and their tests can travel in the same PR. `supautils_platform_policy.sql` seeds the pattern (privileged-extension delegation as the non-superuser `postgres` role, previously uncovered); `supautils_reserved_roles.sql` was the existing precedent.

Draft until the seed test's expected output is confirmed against CI (hand-written; will promote from `regression_output` if it diffs).

PSQL-1159 follow-up context: the warn-mode policy test lands as a follow-up PR on this pattern once this and #2315 merge.


utkarash2991:
Scratching this in favor of folding the policy test directly into #2315, so the conf change and its test merge together. The testinfra split and docs are preserved on this branch if we want them later.

PR #2317, opened 2026-07-27
feat(nix): expose per-version extension packages, catalog, and update tools

brainrake:
Part of [PSQL-1530](https://linear.app/supabase/issue/PSQL-1530/use-profiles-for-multi-version-extensions).

Adds `passthru.perVersion` (`version -> single-version derivation`) to each multi-version extension (18 files, one line each).

Adds flake module `nix/packages/extension-catalog.nix`, exposing `packages` and `legacyPackages`:

- `site-extensions-catalog-<major>` (15/17/orioledb-17)
  - `pg-extensions-catalog.json` mapping `{ ext -> { extversion -> store path } }`
  - `bin/{resolve,update}` that use said json
- `site-extensions-versions-<major>`: single-version ext packages the catalog points at (lean, ~30 MiB closures, no multi-version buildEnv), so CI builds and caches them.
- Tools to update extensions:
  - `site-extensions-resolve` (manifest json -> store paths, reads `PG_EXTENSIONS_CATALOG`)
  - `download-nix-store-paths` (retry/timeout substitution, not extension-specific)
  - `site-extensions-update` (chains both + `nix-env --install --remove-all` onto `PROFILE`, atomic — nix-env only swaps generations on success)

Should be safe since it only adds new outputs.


nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+
+          for f in ${leanPkg}/lib/*; do ln -s "$f" "$out/lib/$(basename "$f")"; done
+          for f in ${leanPkg}/share/postgresql/extension/*; do ln -s "$f" "$ext/$(basename "$f")"; done
+
+          # <name>.control with default_version, from <name>--<version>.control
+          for cf in "$ext"/*--*.control; do
+            b=$(basename "$cf")
+            n="''${b%%--*}"
+            v="''${b#*--}"; v="''${v%.control}"
+            [ -e "$ext/$n.control" ] && continue
```
mmlb: `[[ ]]` is a bash builtin and better than `[` command, we should prefer it since its a bash script anyway.

```suggestion
            [[ -e "$ext/$n.control" ]] && continue
```

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+          ext="$out/share/postgresql/extension"
+
+          for f in ${leanPkg}/lib/*; do ln -s "$f" "$out/lib/$(basename "$f")"; done
+          for f in ${leanPkg}/share/postgresql/extension/*; do ln -s "$f" "$ext/$(basename "$f")"; done
+
+          # <name>.control with default_version, from <name>--<version>.control
+          for cf in "$ext"/*--*.control; do
+            b=$(basename "$cf")
+            n="''${b%%--*}"
+            v="''${b#*--}"; v="''${v%.control}"
```
mmlb: Don't need to use `"` in assignment and is easier to read with the nix escaping in there too.
```suggestion
            n=''${b%%--*}
            v=''${b#*--}; v=''${v%.control}
```

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+
+          # <base>.so -> <base>-<version>.so
+          for f in "$out"/lib/*-[0-9]*; do
+            b=$(basename "$f")
+            suffix=".''${b##*.}"
+            core="''${b%.*}"
+            base="''${core%-[0-9]*}"
+            [ "$base" != "$core" ] || continue
+            gen="$base$suffix"
+            [ -e "$out/lib/$gen" ] || ln -sfn "$b" "$out/lib/$gen"
```
mmlb: ditto re `"` and `[[ ]]` bashism

```suggestion
            suffix=.''${b##*.}
            core=''${b%.*}
            base=''${core%-[0-9]*}
            [[ "$base" != "$core" ]] || continue
            gen=$base$suffix
            [[ -e "$out/lib/$gen" ]] || ln -sfn "$b" "$out/lib/$gen"
```

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+          # modules (e.g. pg-safeupdate), which still ship a library.
+          generic=0
+          for c in "$ext"/*.control; do
+            case "$(basename "$c")" in
+              *--*) ;;
+              *) generic=$((generic + 1)) ;;
+            esac
+          done
+          libs=$(ls -A "$out/lib" 2>/dev/null | wc -l)
+          if [ "$generic" -eq 0 ] && [ "$libs" -eq 0 ]; then
```
mmlb: bash's math tests are nicer than `-eq`, `-ne` ...

```suggestion
          if ((generic == 0)) && ((libs == 0)); then
```

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+          ''
+            set -euo pipefail
+            obj='{}'
+            for path in "''${paths[@]}"; do
+              for ctrl in "$path"/share/postgresql/extension/*.control; do
+                base=$(basename "$ctrl")
+                case "$base" in *--*) continue ;; esac
+                name="''${base%.control}"
+                ver=$(sed -n "s/^default_version = '\(.*\)'.*/\1/p" "$ctrl" | head -1)
+                [ -n "$ver" ] || { echo "no default_version in $ctrl" >&2; exit 1; }
```
mmlb: same thing re bashism and quotes. The `case` could be swapped out for `[[ $base == *--* ]]` instead but meh.

```suggestion
                name=''${base%.control}
                ver=$(sed -n "s/^default_version = '\(.*\)'.*/\1/p" "$ctrl" | head -1)
                [[ -n "$ver" ]] || { echo "no default_version in $ctrl" >&2; exit 1; }

File: nix/packages/extension-catalog.nix [hunk 6]
```diff
@@ -0,0 +1,183 @@
...
+        '';
+
+        # Takes manifest json as argument. Format: {<ext>: <version>}.
+        # Prints nix-store paths of the resolved extensions, one per line.
+        # Does not download or install.
+        site-extensions-resolve = script "site-extensions-resolve" ''
+          : "''${PG_EXTENSIONS_CATALOG:?PG_EXTENSIONS_CATALOG must point at a pg-extensions-catalog.json}"
+          manifest="''${1:?Usage: $0 path-to/pg-extensions.json}"
+          jq -r 'to_entries[] | "\(.key)=\(.value)"' "$manifest" | while IFS='=' read -r name version; do
+            [ -n "$name" ] || continue
```
mmlb: this seems weird, why would we end up with an empty line? If valid, worth a comment for why we might be seeing empty lines

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+      # Catalog json plus bin/site-extensions-{resolve,update} that use it by default.
+      catalogs = lib.mapAttrs' (
+        major: wrappers:
+        lib.nameValuePair "site-extensions-catalog-${major}" (
+          pkgs.runCommand "site-extensions-catalog" { nativeBuildInputs = [ pkgs.makeWrapper ]; } ''
+            mkdir -p "$out/share" "$out/bin"
+            ln -s ${mkCatalogFile wrappers}/share/pg-extensions-catalog.json "$out/share/"
+            makeWrapper ${self'.packages.site-extensions-resolve}/bin/site-extensions-resolve \
+              "$out/bin/site-extensions-resolve" \
+              --set PG_EXTENSIONS_CATALOG "$out/share/pg-extensions-catalog.json"
```
mmlb, line 112: why not --prefix instead? I'd be nice to not have to specify both an envvar and positional arg to the unwrapped scripts, better to both the same way imo.

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+        site-extensions-update = pkgs.writeShellApplication {
+          name = "site-extensions-update";
+          runtimeInputs = [
+            self'.packages.site-extensions-resolve
+            self'.packages.download-nix-store-paths
+          ];
+          text = ''
+            manifest="''${1:-/etc/adminapi/pg-extensions.json}"
+            profile="''${PROFILE:-/nix/var/nix/profiles/site-extensions}"
+            mapfile -t paths < <(site-extensions-resolve "$manifest")
```
mmlb: We have 2 `readarray`s and 0 `mapfile`s in the tree so good to use the same name, not to mention that readarray is more obvious than mapfile imo.

```suggestion
            readarray -t paths < <(site-extensions-resolve "$manifest")
```

nix/packages/extension-catalog.nix
```diff
@@ -0,0 +1,183 @@
...
+        # Takes store paths as args and substitutes from binary cache, with retry.
+        download-nix-store-paths = script "download-nix-store-paths" ''
+          for path in "$@"; do
+            for attempt in 1 2 3; do
+              timeout -k 10s 120s nix-store -r "$path" >/dev/null && continue 2
+              echo "WARNING: attempt $attempt failed for $path" >&2
+            done
+            echo "ERROR: failed to realize $path" >&2
+            exit 1
+          done
```
mmlb: The loop as written makes for serial downloads instead of letting nix download in parallel, if network is flaky this is going to be pretty terrible experience waiting 10s for each path * 3. Maybe makes the happy path a little slower too (but probably not a big deal).

I'd probably do something like

```bash
ok=false
for _ in 1 2 3 4; do
	timeout -k 10s 120s nix-store -r "$@" || continue
	ok=true
	break
done
if ! $ok; then
	ok=true
	for path in "$@"; do
		echo timeout -k 10s 120s nix-store -r "$path" >/dev/null
		echo "ERROR: nix failed to fetch $path after 5 attempts" >&2
		ok=false
	done
fi
if ! $ok;
	exit 1
fi
```

mmlb (changes requested):
A few nits (bashisms and quotes) that would be nice to fix up but the big one to me is the `[[ -n $name ]] || continue` that doesn't smell good to me. Also the serial vs parallel download.

nix/packages/extension-catalog.nix
brainrake: noted, we shouldn't, failing on empty line

nix/packages/extension-catalog.nix
brainrake: `--prefix VAR val` makes VAR equal `val:oldval`, maybe you meant `--add-flags`
Normally you don't have to specify the extension catalog, the script you get in PATH will have it baked in by makeWrapper. So I didn't want to add an optparse for something never actually used form the command line.

nix/packages/extension-catalog.nix
brainrake: wait actually nix has all these options already
`connect-timeout` is 15, leaving as is
`download-attempts` is 5, leaving as is
`stalled-download-timeout` is 300, setting to 120

brainrake: exact failing path is printed on stderr

brainrake:
@mmlb great review, removed a bunch of lines :)

nix/packages/extension-catalog.nix
mmlb: I wish I remembered what I was thinking here, oh well.

mmlb (approved):

PR #2318, opened 2026-07-27
chore: bump pgctld

soedirgo:
Bump pgctld to the Multigres version used by the latest multigres-operator version

### DIFF
--- a/nix/packages/pgctld.nix
+++ b/nix/packages/pgctld.nix
@@ -1,9 +1,10 @@
 {
   lib,
   buildGoModule,
+  go_1_26,
   multigres-src,
 }:
-buildGoModule {
+(buildGoModule.override { go = go_1_26; }) {
   pname = "pgctld";
   version = multigres-src.rev;
   src = multigres-src;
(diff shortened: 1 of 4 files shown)


nix/packages/pgctld.nix
```diff
@@ -1,9 +1,10 @@
 {
   lib,
   buildGoModule,
+  go_1_26,
   multigres-src,
 }:
-buildGoModule {
+(buildGoModule.override { go = go_1_26; }) {
   pname = "pgctld";
```
mmlb:
```suggestion
  buildGo126Module,
  multigres-src,
}:
buildGo126Module {
```

mmlb (changes requested):
Note that in the nixpkgs revision we have pinned go_1_26 is actually rc1 which is not great. That being said it doesn't look like anything in the repo actually uses this nix package (but I very well could be wrong) atm (I would like to change that soon 😄 ) so not a big deal imo.

soedirgo:
Thanks! Applied the suggestion

mmlb (approved):
:shipit:

PR #2319, opened 2026-07-27
ci: Use arm-native-runner for kvm builds on aarch64-linux

mmlb:

Bug fix

NixOS tests require VMs and are being scheduled on blacksmith for linux builds, but that only works for x86_64-linux hosts, aarch64 fails.

## What is the new behavior?

Use arm-native-runners which are currently used for qemu builds and are aarch64 only.

## Additional context

In response to https://github.com/supabase/postgres/actions/runs/30160241092/job/89694722493?pr=2315

Fixes PSQL-1547

### DIFF
--- a/.github/actions/nix-install-ephemeral/action.yml
+++ b/.github/actions/nix-install-ephemeral/action.yml
@@ -1,43 +1,59 @@
-name: 'Install Nix on ephemeral runners'
-description: 'Installs Nix and sets up AWS credentials to push to the Nix binary cache'
+name: Install Nix on ephemeral runners
+description: Installs Nix and sets up AWS credentials to push to the Nix binary cache
 inputs:
   aws-region:
-    description: 'AWS region for the Nix binary cache S3 bucket'
+    description: AWS region for the Nix binary cache S3 bucket
     required: false
-    default: 'us-east-1'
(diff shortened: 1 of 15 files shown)


soedirgo (approved):

utkarash2991:
@mmlb the plan for having the GUC set to WARN i.e. `supautils.restrict_extension_versions = 'warn'` released to general public is for 5th August related changelog: https://github.com/supabase/changelog/pull/128 . Adding just here so we dont accidentally merge and make publicly available the GUC related changes .

mmlb:
good call @utkarash2991 I was working off of your changes so I could trigger the tests and make sure it actually fixed things. Marking draft so I don't merge by mistake...

mmlb:
Ok dropped your commits @utkarash2991 for testing by just gutting most of build-nix.yml and just hard coding pg_tle builds for both aarch64 and x86_64, run can be found in https://github.com/supabase/postgres/actions/runs/30477923240?pr=2319 for https://github.com/supabase/postgres/commit/8d829fcd6e97f8772850b46a66d5fa4b70643435.

This took a bit more refactoring of nix-install-ephemeral + users than I expected but I'm ok with that.

PR #2320, opened 2026-07-27
ci: Fix dockerhub-release-matrix matrix generation

mmlb:

Bug fix

Dockerhub release workflow is failing due to the matrix strategy json being messed up. GITHUB_OUTPUT should be newline deliminated k=v pairs but we are currently writing w/o a newline so it sees one long non-json line.

## What is the new behavior?

The fix is by having newlines separate each output definition.

## Additional context

See failed run: https://github.com/supabase/postgres/actions/runs/30275880815

### DIFF
--- a/.github/workflows/dockerhub-release-matrix.yml
+++ b/.github/workflows/dockerhub-release-matrix.yml
@@ -38,9 +38,9 @@ jobs:
                 $e | insert tag $"($ver)($e.tag_suffix)" | insert base_tag $ver
               })
           let combined = ($base | append $layered)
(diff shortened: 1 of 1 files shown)


brainrake (approved):

mmlb:
I gave up on trying to get this to build but not push in PR branches, its too much of a diff with the current setup using ansible/vars.yml for base images. Will leave that for my upcoming work to use automatic git based versioning instead.

mmlb:
Ok my test build worked to build the images: https://github.com/supabase/postgres/actions/runs/30381275800. Canceling those CI and pushing w/o the DNM commit now.

PR #2326, opened 2026-07-30
Lots of bootstrap script clean ups

mmlb:

Maintenance

So much confusing code here when apt can do fallbacks better, also we don't always use the same apt configs in all apt calls. Non nix'd scripts aren't shellcheck clean. Bunch of dead/unused code hanging around. We run apt-get upgrade all over the ami/qemu image build stages.

## What is the new behavior?

We now let apt handle fallbacks itself. All the scripts use apt with the same settings. I deleted unused files from scripts/ and moved those that were used into ebssurrogate/scripts. I enabled shellcheck in git-hooks and excluded currently failing files.

This is clean ups/refactors in preparation of some other changes in ansible that will come next.


mmlb:
@claude review

samrose (approved):

PR #2327, opened 2026-07-31
ci: resolve package install sha from a flake input

brainrake:
Workflows that install nix packages via `github:supabase/postgres/<sha>#pkg` use the run's own commit sha. That breaks when the exact same workflow file runs in a different repo.

Adds a `resolve-git-sha` composite action: if a `GIT_SHA_FROM_FLAKE_INPUT` repo variable names a flake input, resolves that input's pinned rev from `flake.lock`; otherwise falls back to `github.sha` unchanged. Wired into the four workflows that construct these install refs (`ami-release-nix.yml`, `testinfra-ami-build.yml`, `qemu-image-build.yml`, `test.yml`).

`build-ami/action.yml` gets a second input, `packages_git_sha`, kept separate from `git_sha` (AMI identity tag, used for cache lookups, must stay the triggering commit) so the two don't get conflated.

No effect here: the variable is never set in this repo, so every site falls through to `github.sha` exactly as today.

### DIFF
--- a/.github/actions/build-ami/action.yml
+++ b/.github/actions/build-ami/action.yml
@@ -14,6 +14,9 @@ inputs:
   git_sha:
     description: 'Git SHA for this build'
     required: true
+  packages_git_sha:
+    description: 'Git SHA to install packages from'
+    required: true
   instance_type:
     description: 'EC2 instance type for the build'
     required: true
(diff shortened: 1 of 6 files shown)


imor (approved):

mmlb:
How in the world did this PR break my branch/PR...https://github.com/supabase/postgres/actions/runs/30667526580/job/91279509777?pr=2326

mmlb:
ugghhh its because testinfra workflow has `uses: supabase/postgres/.github/actions/shared-checkout@HEAD` which changes the git workdir from github's synthetic merge commit for PRs back to the HEAD commit of the branch but nix-build workflow does not and there's a disagreement of what workflows/actions are "supposed to exist in .github/" at that point. nix-build is using synthetic merge commit and kicks off a testinfra that has a resolve-git-sha step, but then testinfra runs and checks out my branch's HEAD which does not have the action and boom failure...

PR #2329, opened 2026-07-31
fix(nix): remove pinned Go version in supascan/gatekeeper

brainrake:
## Summary
- `go_1_24` is missing in the nixpkgs 26.05pre bump being pulled in by #2328's `flake.lock` update, breaking `nix-eval` for `checks.*.supascan` on all platforms.
- Rather than repin to another explicit Go version, this drops the override entirely: `pkgs.go` (what `pkgs.buildGoModule` defaults to) already resolves to 1.25.1 on the current nixpkgs and 1.25.4 on the bumped one — both well above supascan's `>= 1.23.2` requirement. This matches how the other Go packages in this flake (`goss`, `pg-startup-profiler`, `packer`, `wal-g`) are already built, and avoids this whole class of breakage on future nixpkgs bumps.
- Same pin existed in `gatekeeper.nix`, fixed the same way.

## Test plan
- [x] `nix eval .#packages.x86_64-linux.supascan.name` / `gatekeeper.name` succeed on develop's current nixpkgs
- [x] `nix build .#packages.aarch64-darwin.supascan` succeeds, `go test ./...` passes, output hash matches the pinned-1.25 build (functionally identical)


imor (approved):

brainrake:
Looks like there is not enough space for test-ami

brainrake:
test-ami-nix (15) fails with out of space error - tracked separately in [PSQL-1570](https://linear.app/supabase/issue/PSQL-1570), unrelated to this fix.

PR #2336, opened 2026-07-31
[validation only] fix(nix): pin ncurses old glibc symver for cfgetospeed

brainrake:
Temporary validation PR targeting #2328's branch to confirm this actually fixes \`GLIBC_2.42' not found\` (from libncursesw.so.6) with real bumped-nixpkgs artifacts on the existing ubuntu-24.04/ubuntu-latest runners (no runner bump needed — the whole point is the bundle now stays compatible with Ubuntu 24.04, matching what the README claims). Will be closed in favor of a develop-based PR once confirmed green.


brainrake:
**Validation result:** the \`dlvsym\`-based ncurses fix works. \`ncurses\` built successfully against the bumped glibc (2.42) with no \`cfgetospeed\`/\`GLIBC_2.42\`/linker errors - confirms the fix resolves the original \`GLIBC_2.42' not found\` failure at the ncurses level.

Overall CI on this PR still fails, but for an unrelated, already-diagnosed reason: this branch is ncurses-only (based on \`develop\`, no cgal fix), so it still hits the separate pre-existing \`boost\`/\`sfcgal\`/postgis cascade that #2332 fixes independently. That's expected here, not a problem with this fix.

Closing as validation-only; the real fix will land as its own PR against \`develop\`.

PR #2337, opened 2026-07-31
[validation only, new-glibc path] ncurses cfgetospeed symver + cgal pin

brainrake:
Combines the ncurses fix with the cgal pin (#2332) so build-native actually succeeds against #2328's bumped nixpkgs, letting test-portability run for real. Validates the new-glibc (>=2.42) path specifically; PR #2336 validates the old-glibc (develop) path. Will close once both confirm green.


brainrake:
**Validation result: inconclusive.** \`Build linux-x64\` and \`Build linux-arm64\` were both **cancelled** after ~6 hours (hit the GitHub Actions job timeout while \`nix build\` was still running) - not a real pass or fail signal. Didn't get a clean end-to-end confirmation of ncurses fix + cgal pin combined against the actual bumped nixpkgs.

The ncurses portion is independently confirmed working (see #2336's validation). The cgal pin is independently confirmed working in #2332. Recommend re-validating the combination with more time budget (or investigating what's making the combined build slower/heavier) rather than trusting this run's result either way.

Closing as validation-only.

PR #2340, opened 2026-08-01
[validation only] Go pin removal on top of AMI volume bump

brainrake:
Temporary validation PR: #2329's Go pin removal rebased onto #2339's 20GB build volume bump, to confirm test-ami-nix (15, amd64/arm64) passes with the larger volume. Will close once confirmed.

### DIFF
--- a/nix/packages/gatekeeper.nix
+++ b/nix/packages/gatekeeper.nix
@@ -1,10 +1,10 @@
 { pkgs, ... }:
 let
 
-  go124 = pkgs.go_1_24;
-  buildGoModule124 = pkgs.buildGoModule.override { go = go124; };
+  go125 = pkgs.go_1_25;
+  buildGoModule125 = pkgs.buildGoModule.override { go = go125; };
 
-  upstream-gatekeeper = buildGoModule124 {
+  upstream-gatekeeper = buildGoModule125 {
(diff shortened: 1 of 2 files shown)


brainrake:
**Validation result: blocked, not about the Go fix itself.** This PR exists to test #2329's Go pin removal against \`test-ami-nix (15, ...)\`, gated on fixing the unrelated \`No space left on device\` AMI build failure (see #2339).

Both disk-space fix attempts in #2339 failed to actually resolve it:
- Bumping \`/dev/xvdc\` (scratch build-vol) 16GB->20GB: no effect - the real ENOSPC happens in ansible's own log flush on the *builder instance's own root disk*, a different volume entirely.
- Adding an explicit \`/dev/sda1\` mapping (theorizing it was that separate root): **caused a regression** - on this HVM AMI, \`/dev/sda1\` aliases the same device as the AMI's actual root (\`ami_root_device\` -> \`/dev/xvda\`), so it grew the *output* AMI's root snapshot, breaking previously-passing PG17/orioledb-17 builds too. Reverted.

Root cause of the disk-space issue is still unidentified. This PR can't produce a real signal on the Go fix until that's sorted - closing as validation-only rather than leaving it open on a broken premise.

PR #2350, opened 2026-08-07
Add ext-explorer: browse pg extension versions by minor release

cmartin616:
## Summary
- Adds `nix run .#ext-explorer`, which generates `pg-extension-explorer.html` — pick a Postgres major (15/17/orioledb-17) and a minor version that actually shipped, and see exactly which extensions and versions were on disk for that build.
- History comes from `nix/ext/version-history.json`, an append-only manifest seeded with the real historical minor-version boundaries mined from `nix/config.nix`'s git log. The generator auto-freezes a new row whenever it detects `nix/config.nix`'s version fields changed since the comparison base, so the timeline grows on its own as future bumps land.
- CI (`.github/workflows/ext-explorer-check.yml`) fails PRs/pushes touching the relevant source files if the committed HTML/history are stale, so this can't silently drift — regenerate locally with `nix run .#ext-explorer` and commit the result.

## Test plan
- [x] `nix run .#ext-explorer -- --check` passes on a clean tree
- [x] Regeneration is idempotent (no diff on a second run)
- [x] Simulated a `nix/config.nix` version bump and confirmed the generator correctly freezes the prior state as a new history row before regenerating for the new one
- [x] Confirmed the "current" period isn't tied to `HEAD`'s commit hash (would otherwise go stale on every unrelated commit)
- [ ] Manual click-through of `pg-extension-explorer.html` in a browser (major/minor dropdowns, search, CLI-only filter)


mmlb:
@cmartin616 can update the PR description and add reasoning for this change? Why are we introducing this? Also for things that need to be up2date it would be better to integrate into pre-commit check in nix/hooks.nix, this way it'll be generated locally on any changes to source or the commit will fail and ci will automatically check it no new workflow needed.

mmlb:
Actually besides all that I'm not really a fan of an ever growing file in git, this seems like something we can generate as a build artifact straight from git history no?

PR #2351, opened 2026-08-08
AMI: Clean up messy apt handling

mmlb:
# What kind of change does this PR introduce?

Maintenance

# What is the current behavior?

So much confusing code here when apt can do fallbacks better.
Bunch of dead/unused code hanging around.
We run apt-get update/upgrade all over the ami/qemu image build stages.

# What is the new behavior?

We now let apt handle fallbacks itself and apt usage is cleaner overall.


Crispy1975 (approved):

mmlb:
Reverting the PR/commit that caused this and converting this to draft. Will re-introduce everything with fixes in here after bugs are ironed out.

mmlb:
I've changed this PR from fix of #2326 to second version of it (since it was reverted). I'm going to push commit by commit so CI runs and we can make sure we don't repeat the issues from #2326 (I've done the same locally so am confident but want the extra feel good of :heavy_check_mark: for each commit)

mmlb:
depends on #2381

mmlb:
Not sure why but when CI runs on the latest commit arm64 machines fail to boot up into stage2 correctly and ssh is never able to connect. Going to push commits one at a time to find when it breaks.

mmlb:
pg15 amd64 has made it into stage2 in https://github.com/supabase/postgres/actions/runs/32775503892/job/97588403063?pr=2351 already so pushing next commit.

hunleyd (changes requested):
1. qemu-bootstrap-nix.sh setup_apt — three stacked defects:
     - [[ $ARCH == x86_64 ]] — ARCH is never set in this script and qemu.pkr.hcl only passes GIT_SHA/POSTGRES_MAJOR_VERSION. Dead branch, always false, no nounset to catch it.
     - Missing the if [[ $ubumirror == "$defmirror" ]]; then return; fi guard the sibling surrogate-bootstrap-nix.sh has. On arm64 (the only arch CI actually builds), defmirror == ubumirror (ports.ubuntu.com both), so this fires today, appending a duplicate URI to the sources file every single qemu build (apt "configured multiple times" warnings, double index fetch).
     - Hardcodes noble instead of deriving $CODENAME like its sibling. rewrite this function to match surrogate-bootstrap-nix.sh's copy exactly (derive ARCH via dpkg --print-architecture with a :? guard, compare amd64, derive $CODENAME from /etc/os-release, add the early return).
2. surrogate-bootstrap-nix.sh:182 — new mirror=$(awk '/^URIs:/{uri=$2} /^Suites:.*\<CODENAME-updates\>/{...}') uses GNU-only \</\> word-boundary syntax. I reproduced this in a real Ubuntu 24.04 container: default /usr/bin/awk there is mawk, and the exact pattern returns empty output; mirror comes out "", fed straight into debootstrap ... "$mirror". Fix: don't use \</\> — anchor on whitespace/field boundaries instead.
3. chroot-bootstrap-nix.sh — deleted the APT_OPTIONS array (Install-Recommends=false, Install-Suggests=false, Acquire::Languages=none) without replacing it. One of the two apt-get install calls in the same file still has --no-install-recommends, the other doesn't which directly contradicts this PR's own stated goal ("all scripts use apt with the same settings"). Also: unquoted fonts-dejavu*/linux-headers* in the cleanup.sh package array (real glob-expansion hazard, currently harmless only because cwd has no matching files).
4. qemu-bootstrap-nix.sh — setup_apt/update_and_upgrade_apt now run before waitfor_boot_finished (the old code and the sibling script both wait first). Cloud-init can still be rewriting ubuntu.sources at that point, silently discarding the sed edit, and/or apt lock-contending with cloud-init's own apt-daily.service and hard-failing under errexit. Didn't fail today's CI run, but it's a timing race, not a guarantee.

mmlb:
> 1. qemu-bootstrap-nix.sh setup_apt — three stacked defects:
>        
>        * [[ $ARCH == x86_64 ]] — ARCH is never set in this script and qemu.pkr.hcl only passes GIT_SHA/POSTGRES_MAJOR_VERSION. Dead branch, always false, no nounset to catch it.
>        * Missing the if [[ $ubumirror == "$defmirror" ]]; then return; fi guard the sibling surrogate-bootstrap-nix.sh has. On arm64 (the only arch CI actually builds), defmirror == ubumirror (ports.ubuntu.com both), so this fires today, appending a duplicate URI to the sources file every single qemu build (apt "configured multiple times" warnings, double index fetch).
>        * Hardcodes noble instead of deriving $CODENAME like its sibling. rewrite this function to match surrogate-bootstrap-nix.sh's copy exactly (derive ARCH via dpkg --print-architecture with a :? guard, compare amd64, derive $CODENAME from /etc/os-release, add the early return).

Fixed all of these, I run qemu builds on my x86_64 machine so need to keep it.

>     2. surrogate-bootstrap-nix.sh:182 — new mirror=$(awk '/^URIs:/{uri=$2} /^Suites:.*<CODENAME-updates>/{...}') uses GNU-only </> word-boundary syntax. I reproduced this in a real Ubuntu 24.04 container: default /usr/bin/awk there is mawk, and the exact pattern returns empty output; mirror comes out "", fed straight into debootstrap ... "$mirror". Fix: don't use </> — anchor on whitespace/field boundaries instead.

Yep looks like this was missed when I switched over to `grep` pipeline instead.

>     3. chroot-bootstrap-nix.sh — deleted the APT_OPTIONS array (Install-Recommends=false, Install-Suggests=false, Acquire::Languages=none) without replacing it. One of the two apt-get install calls in the same file still has --no-install-recommends, the other doesn't which directly contradicts this PR's own stated goal ("all scripts use apt with the same settings"). Also: unquoted fonts-dejavu*/linux-headers* in the cleanup.sh package array (real glob-expansion hazard, currently harmless only because cwd has no matching files).

All scripts use apt with the same settings is no longer a goal for this PR (will be done in a separate one). Will handle the glob expansion note.


>     4. qemu-bootstrap-nix.sh — setup_apt/update_and_upgrade_apt now run before waitfor_boot_finished (the old code and the sibling script both wait first). Cloud-init can still be rewriting ubuntu.sources at that point, silently discarding the sed edit, and/or apt lock-contending with cloud-init's own apt-daily.service and hard-failing under errexit. Didn't fail today's CI run, but it's a timing race, not a guarantee.

Fixed

hunleyd (approved):

PR #2352, opened 2026-08-10
packer: Drop unused files

mmlb:

Maintence

Have a few packer/ansible related files that we no longer need.

## What is the new behavior?

Unused files are not in the repo.


mmlb:
These files no longer exist in repo.

PR #2353, opened 2026-08-10
packer: Use dist-upgrade instead of upgrade

mmlb:

Maintenance

All but one of apt upgrading scripts do a plain `upgrade`, `chroot-bootstrap-nix.sh` is the odd script out. This is not ideal because we are being inconsistent and `upgrade` isn't as good as `dist-upgrade`.

## What is the new behavior?

We use `dist-upgrade` instead of `upgrade` and do so any time we do an apt-get --update (ignoring adding new mirror).


mmlb:
@claude review

samrose (approved):

brainrake (approved):

PR #2356, opened 2026-08-11
fix: harden post-upgrade vacuum analyze in pg_upgrade complete script

Crispy1975:
The post-upgrade `vacuumdb --analyze-in-stages` step failed silently: `start_vacuum_analyze` ended with an `echo`, so it always returned 0 and `retry` never fired. It was also launched straight after `service postgresql restart` with no readiness check.

- propagate vacuumdb's exit code so retries actually happen; drop the dead retry on `run_generated_sql` (deliberately fail-soft)
- wait for `pg_isready` after the step-3 start and step-5 restart before continuing
- analyse partitioned parent tables after vacuumdb — vacuumdb skips them until PG19 and autovacuum never analyses them; incremental on retry, safe against hostile database names
- timeouts for live traffic: `lock_timeout` 10s (vacuumdb and parent analyze), `statement_timeout` 10min on parent analyze; vacuumdb parallelised (`-j` nproc/4) with `vacuum_cost_delay` reset
- fail-soft steps (analyze, patches, generated SQL) log a WARNING instead of failing the completed upgrade, and ship the log to the reporting project so warnings are visible off-box
- prefix step log lines with UTC timestamps (`log` helper in `common.sh`); partitioned-table fixture + stats assertions in the upgrade tests
- bump `postgres_release` patch versions — merge publishes the scripts and cuts new AMIs

https://linear.app/supabase/issue/INDATA-1218/hardening-postgresansiblefilesadmin-api-scriptspg-upgrade-scripts


matthudsonatx (approved):

matthudsonatx (approved):

tgallacher (approved):

cfunkhouser (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -14,6 +14,10 @@ source "$SCRIPT_DIR/common.sh"
 IS_CI=${IS_CI:-}
 LOG_FILE="/var/log/pg-upgrade-complete.log"
 
+function log {
```
cfunkhouser: Moving this to `common.sh` would allow it to be reused across all the upgrade scripts, which will be nice when I start touching `initiate.sh` here momentarily.

cfunkhouser (changes requested):

hunleyd (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Fixed that up. @cfunkhouser

cfunkhouser (approved):
Thanks!

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
hunleyd, line 1:
```
diff --git i/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh w/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
index 84442d43..8f4e0efe 100755
--- i/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
+++ w/ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
@@ -314,14 +314,12 @@ function complete_pg_upgrade {
 	fi
 
 	log "6. Starting vacuum analyze"
-	# A failed analyze is not worth failing the whole upgrade for; the status
-	# file already reads "complete" and the ERR trap would flip it to "failed"
-	retry 3 start_vacuum_analyze || echo "WARNING: vacuum analyze failed after retries"
+	# A failed analyze is not worth failing the whole upgrade for; the status file already reads "complete" and the ERR trap would flip it to "failed"
+	retry 3 start_vacuum_analyze || log "WARNING: vacuum analyze failed after retries"
 
 	log "6.1. Analyzing partitioned tables"
-	# vacuumdb skips partitioned parents (fixed upstream only in PG19) and
-	# autovacuum never analyzes them, so without this they'd have no stats at all
-	retry 3 analyze_partitioned_tables || echo "WARNING: partitioned table analyze failed after retries"
+	# vacuumdb skips partitioned parents (fixed upstream only in PG19) and autovacuum never analyzes them, so without this they'd have no stats at all
+	retry 3 analyze_partitioned_tables || log "WARNING: partitioned table analyze failed after retries"
 
 	log "Upgrade job completed"
 }
@@ -369,14 +367,16 @@ function start_vacuum_analyze {
 }
 
 function analyze_partitioned_tables {
-	local rc=0 db stmts
-	for db in $(psql -X -h localhost -p 5432 -U supabase_admin -d postgres -A -t -c "select datname from pg_database where datallowconn"); do
-		stmts=$(psql -X -h localhost -p 5432 -U supabase_admin -d "$db" -A -t -c "select format('ANALYZE %I.%I;', n.nspname, c.relname) from pg_class c join pg_namespace n on n.oid = c.relnamespace where c.relkind = 'p'") || {
+	local rc=0 db stmts dbs
+	# Capturing the substitution's exit status here (vs. using it directly as the for-loop's word list) stops a failed enumeration from tripping the ERR trap inside the subshell and overwriting /tmp/pg-upgrade-status back to "failed"
+	dbs=$(run_sql -X -p 5432 -A -t -c "select datname from pg_database where datallowconn") || return 1
+	for db in $dbs; do
+		stmts=$(run_sql -X -p 5432 -d "$db" -A -t -c "select format('ANALYZE ONLY %I.%I;', n.nspname, c.relname) from pg_class c join pg_namespace n on n.oid = c.relnamespace where c.relkind = 'p'") || {
 			rc=1
 			continue
 		}
 		if [ -n "$stmts" ]; then
-			echo "$stmts" | psql -X -h localhost -p 5432 -U supabase_admin -d "$db" -v ON_ERROR_STOP=1 || rc=1
+			echo "$stmts" | run_sql -X -p 5432 -d "$db" -v ON_ERROR_STOP=1 || rc=1
 		fi
 	done
 	return $rc
```

hunleyd (dismissed):
things of note:
- ship_logs only fires from the ERR trap. A warn-only failure at step 6/6.1 never leaves the instance. That directly undercuts the Linear issue's own stated motivation ("we could have had an alert on this... react fast").
- No lock_timeout on the new ANALYZE session — by step 6.1 the instance is already serving traffic; a blocking customer lock on a partitioned parent could hang complete_pg_upgrade forever.
- retry 3 wraps the whole multi-database loop — one stuck database forces a full re-do of every database on each retry.
- log helper is only half-adopted — initiate.sh sources the same common.sh but still uses bare echo for its ~13 step lines.
- No test fixture has a partitioned table; analyze_partitioned_tables runs against an empty result set in CI; the behavior it exists to fix is never exercised.
- retry 8 pg_isready gives ~254s of backoff before hard-failing the upgrade. is that enough? too much?

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
cfunkhouser: Don't worry about the `initiate.sh` call sites; I'm handling that along with the Salt disablement as part of INDATA-1220.

NikolayS: great points – indeed, `lock_timeout` makes sense

another point here:

if target version is >= 18, `pg_upgrade` will transfer most of stats – see https://www.postgresql.org/docs/18/pgupgrade.html; and it works when upgrading from older PG to PG18 ([I checked](https://postgres.ai/blog/20260324-pg18-stats-upgrade-across-versions)).

So for upgrades to PG18 and older I'd:
- omit all vacuumdb calls completely (including the case for partitioned tables)
- and included `vacuumdb --all --missing-stats-only` (with `lock_timeout` via `PGOPTIONS` and finite number retries, I think it makes total sense)

On another note, for larger DBs, if/when we run `vacuumdb`, it makes sense to speed up – using two things:
- `PGOPTIONS='-c vacuum_cost_delay=0'` (it is 0 by default but in case it was changed, makes sense to get it back)
- and some`-j` (if we serve traffic already, I'd take 1/4 of vCPU count)

Crispy1975:
@hunleyd 

I have taken most of the suggestions... swapped to `run_sql/log`, took your `dbs` capture, added a partitioned fixture + stats assertion to the upgrade tests, and put a `60s` `lock_timeout` on the analyze session (tested both paths against a live PG15 — a held lock now fails fast into the retry instead of hanging).

One thing I couldn't do... `ANALYZE ONLY` is `PG18` syntax — on `15/17` it's a hard syntax error at or near `ONLY` (verified on 15.x), and those are our only targets. So plain `ANALYZE` is the way; the redundant partition re-sampling is bounded and idempotent.

On the rest:
  - retry re-running all DBs: accepted trade-off — the loop is idempotent and cheap, per-DB retry bookkeeping didn't seem worth it.
  - log in `initiate.sh`: @cfunkhouser picking that up in `INDATA-1220`.
  - `ship_logs` on the warn path: the mechanism exists (`ship_logs` already posts to the reporting project) — leaving it for a follow-up rather than growing this PR too much.
  - retry 8 (~4 min): these restarts follow clean shutdowns so recovery is trivial; if it does trip, the failure is loud and `CLEANUP` restores the original instance, which beats hanging. Happy to bump it if you feel that strongly on it.

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -355,7 +364,26 @@ function start_vacuum_analyze {
...
+		stmts=$(run_sql -X -p 5432 -d "$db" -A -t -c "select format('ANALYZE %I.%I;', n.nspname, c.relname) from pg_class c join pg_namespace n on n.oid = c.relnamespace where c.relkind = 'p'") || {
+			rc=1
+			continue
+		}
+		if [ -n "$stmts" ]; then
+			# The instance is serving traffic by now — fail fast into the retry rather than queue behind a customer lock
+			{
+				echo "set lock_timeout = '60s';"
+				echo "$stmts"
+			} | run_sql -X -p 5432 -d "$db" -v ON_ERROR_STOP=1 || rc=1
```
NikolayS, line 418: If the first ANALYZE hits the 60s `lock_timeout`, every remaining partitioned table in that db gets skipped, and your retry starts from zero -- including re-recursing into leaves already done. Adjusting the select above to exclude tables that that have stats already would make retries more efficient

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: @NikolayS Adopted the `-j` and `PGOPTIONS` suggestions. We're not upgrading to `PG18` yet, only 14/15 > 17. But will note a follow-up for when 18 lands.

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -355,7 +364,26 @@ function start_vacuum_analyze {
...
 	fi
 	vacuumdb --all --analyze-in-stages -U supabase_admin -h localhost -p 5432
-	echo "Upgrade job completed"
+}
+
+function analyze_partitioned_tables {
+	local rc=0 db stmts dbs
+	# Capture the list (vs substituting into the for-list) so a failed enumeration can't trip the ERR trap inside the subshell and overwrite the status file
+	dbs=$(run_sql -X -p 5432 -A -t -c "select datname from pg_database where datallowconn") || return 1
+	for db in $dbs; do
```
NikolayS: I think these two cases are dangerous for this line:


```
test=# create database "*";
CREATE DATABASE
test=# \c *
You are now connected to database "*" as user "nik".
*=#
*=#
*=# create database "my db";
CREATE DATABASE
*=# \c "my db"
You are now connected to database "my db" as user "nik".
my db=#
```

NikolayS: could be fixed with smth like:

```bash
while IFS= read -r db; do
    ...
done <<< "$dbs"
```

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -355,7 +364,26 @@ function start_vacuum_analyze {
...
+		stmts=$(run_sql -X -p 5432 -d "$db" -A -t -c "select format('ANALYZE %I.%I;', n.nspname, c.relname) from pg_class c join pg_namespace n on n.oid = c.relnamespace where c.relkind = 'p'") || {
+			rc=1
+			continue
+		}
+		if [ -n "$stmts" ]; then
+			# The instance is serving traffic by now — fail fast into the retry rather than queue behind a customer lock
+			{
+				echo "set lock_timeout = '60s';"
```
NikolayS: I think 60s is generous for "fail fast."

Under a live workload with autovacuum running (and it will be, aggressively, right after upgrade when everything looks stat-less), SUE-vs-SUE conflicts with autovacuum's own ANALYZE are likely. 

Autovacuum cancels itself for a conflicting waiter, so short timeouts rarely lose; 5–10s with the outer retry would behave better than parking a session for a minute per table. 

Also, if we think about hardening for bigger scale, I'd consider a `statement_timeout` companion if these parents are huge -- recursion into hundreds of leaves under one statement can run long, and you said the instance is serving traffic. Since we fail-soft on ANALYZE, it's better to do it instead of running for ages?

===

in general: thanks for doing this – it's great to see this improvement, many Postgres platforms avoid ANALYZE autmation completely, leaving it on user's shoulders and then we deal with situations like "upgraded on weekend, all was good, then monday came and production is down".

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Fixed up to scan stats.

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Updated taking the suggested iteration style.

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -355,7 +364,26 @@ function start_vacuum_analyze {
 		source "/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh"
 	fi
 	vacuumdb --all --analyze-in-stages -U supabase_admin -h localhost -p 5432
```
NikolayS: since we start using lock_timeout for manual ANALYZE -- it makes sense to have it here too, via PGOPTIONS

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -311,8 +313,15 @@ function complete_pg_upgrade {
 		retry 3 CI_start_postgres
 	fi
 
-	echo "6. Starting vacuum analyze"
-	retry 3 start_vacuum_analyze
+	log "6. Starting vacuum analyze"
+	# A failed analyze is not worth failing the whole upgrade for; the status file already reads "complete" and the ERR trap would flip it to "failed"
+	retry 3 start_vacuum_analyze || log "WARNING: vacuum analyze failed after retries"
```
NikolayS: common.sh already loads REPORTING_ANON_KEY / reporting credentials – maybe this warning could be delivered to become visible?

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Updated

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Updated

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Adding in `ship_logs` which goes to a target project.

cfunkhouser (approved):

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -326,7 +355,10 @@ function run_generated_sql {
 	if [ -d /data/sql ]; then
 		for FILE in /data/sql/*.sql; do
 			if [ -f "$FILE" ]; then
-				run_sql -f "$FILE" || true
+				run_sql -f "$FILE" || {
+					log "WARNING: generated SQL file $FILE failed"
+					warnings=1
```
hunleyd, line 361: this makes the function always ret 0

hunleyd (changes requested):

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
Crispy1975: Yeah, though it was equally true before this PR (the old per-file `|| true` did the same), so the retry has never fired. It's deliberately fail-soft as propagating would trip the `ERR` trap and fail the whole upgrade, and a retry would re-run already-applied `ALTER EXTENSION UPDATE`s, which in turn would then error.

I dropped the dead retry wrapper and left a comment saying why. Whether these failures should block the upgrade properly is going in the follow-up issue. I don't want this PR to grow too much beyond what it is. I will create follow-up tasks here.

hunleyd (changes requested):
- vacuumdb's new lock_timeout=10s: any lock-timeout error aborts the entire staged, all-databases run (not just that table). retry 3 restarts from stage 1 — if attempt 2 dies mid-stage-3, it can leave every table with stage-1's coarse target=1 stats instead of the finer stats an earlier attempt already wrote. Suggested fix: vacuumdb --skip-locked instead of a global lock timeout.
- analyze_partitioned_tables's ON_ERROR_STOP=1 batches every pending ANALYZE into one psql call and the first failing parent strands every later parent in that database, and 3 retries just re-fail on the same one. Suggested fix: one run_sql call per statement instead of one batched call.
- execute_patches/execute_extension_upgrade_patches's new fail-soft wrapping can't catch a failure inside the function. set -e is suppressed for the whole function body because it's the left operand of ||, and the function's exit status is just its last command's. Concretely: if the pg_cron drop/recreate transaction fails but the later grants succeed, execute_patches returns 0, no WARNING logs, nothing ships — silently undermining the PR's own stated goal for exactly the steps most likely to fail. Fix: ( set -e; execute_patches ) || { ... }.
- -d "$db" lets a hostile database name redirect the connection. psql's -d parses its value as a libpq conninfo string when it contains =; a customer-created database named e.g. host=attacker.example.com dbname=postgres redirects a root-launched psql connection off-box. This directly contradicts the PR body's own "safe against hostile database names" claim. refresh_collation.sh in the same directory already has conninfo_for_db() for exactly this hazard (with a comment describing the same danger)

hunleyd (approved):

PR #2362, opened 2026-08-13
feat: Add `ORIOLEDB_ENABLED` into `/etc/environment.d/postgresql.env`

za-arthur:

`database-optimizations.service` ignores `ORIOLEDB_ENABLED` variable, which is defined in `/etc/environment`. This PR adds it into `/etc/environment.d/postgresql.env` similar to other variables and explicitly use it in the `database-optimizations.service` similar to `postgresql.service`. We keep `/etc/environment` unchanged for backward compatibility.

`database-optimizations.service` ignores `ORIOLEDB_ENABLED` variable

## What is the new behavior?

`database-optimizations.service` respects `ORIOLEDB_ENABLED` variable defined in `/etc/environment.d/postgresql.env`

## Additional context

https://linear.app/supabase/issue/PSQL-1621

### DIFF
--- a/ansible/files/database-optimizations.service
+++ b/ansible/files/database-optimizations.service
@@ -3,6 +3,7 @@ Description=Postgresql optimizations
 
 [Service]
 Type=oneshot
+EnvironmentFile=/etc/environment.d/postgresql.env
 # we do not want failures from these commands to cause downstream service startup to fail
 ExecStart=-/opt/supabase-admin-api optimize db --destination-config-file-path /etc/postgresql-custom/generated-optimizations.conf
 ExecStart=-/opt/supabase-admin-api optimize pgbouncer --destination-config-file-path /etc/pgbouncer-custom/generated-optimizations.ini
--- a/ansible/tasks/stage2-setup-postgres.yml
+++ b/ansible/tasks/stage2-setup-postgres.yml
@@ -53,6 +53,11 @@
             line: 'ORIOLEDB_ENABLED=true'
             path: '/etc/environment'
 
+        - name: Add ORIOLEDB_ENABLED environment variable to postgresql.env
+          ansible.builtin.lineinfile:
+            line: 'ORIOLEDB_ENABLED=true'
+            path: '/etc/environment.d/postgresql.env'
+
 - name: Execute things when stage2
   when: stage2
   become: true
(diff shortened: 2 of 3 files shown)


ansible/files/database-optimizations.service
```diff
@@ -3,6 +3,7 @@ Description=Postgresql optimizations
 
 [Service]
 Type=oneshot
+EnvironmentFile=-/etc/environment.d/postgresql.env
```
mmlb: Not sure why we have `-` before this file, its guaranteed to be there from chroot-bootstrap-nix.sh and the task below too. IMO better a hard error than this confusing setup. Not going to block the PR on it though since there's other cases of this in the repo.

mmlb (approved):

ansible/tasks/stage2-setup-postgres.yml
```diff
@@ -242,3 +242,10 @@
     line: 'GRN_PLUGINS_DIR=/var/lib/postgresql/.nix-profile/lib/groonga/plugins'
     path: '/etc/environment.d/postgresql.env'
   become: true
+
+- name: Append ORIOLEDB_ENABLED to /etc/environment.d/postgresql.env
```
mmlb: Actually this would be better in the block above that is for orioledb and already has `become: true` so you could drop that from this bit.

mmlb (changes requested):

ansible/tasks/stage2-setup-postgres.yml
za-arthur: @mmlb I think I put this here because `/etc/environment.d/postgresql.env` is created later, after the orioledb block. Otherwise it required `create: true`. Would that be okay?

mmlb: /etc/environment.d/postgreqsql.env should already exist before stage2 runs since its being written to in chroot-bootstrap-nix.sh which runs from stage1 https://github.com/supabase/postgres/blob/develop/ebssurrogate/scripts/chroot-bootstrap-nix.sh#L288-L300

za-arthur: Makes sense. Fixed in https://github.com/supabase/postgres/pull/2362/changes/8fcbdc39153ebc6dfcf41fbff2db306f9ba26042

ansible/files/database-optimizations.service
za-arthur: Fixed in https://github.com/supabase/postgres/pull/2362/changes/8fcbdc39153ebc6dfcf41fbff2db306f9ba26042

mmlb:
ok lgtm, will wait for the testinfra builds to make sure I didn't read anything wrong :|

mmlb (approved):

utkarash2991 (approved):

PR #2379, opened 2026-08-19
feat: add orioledb debug symbols

za-arthur:

Add separateDebugInfo to `nix/ext/orioledb.nix` so `orioledb.so`'s debug symbols are preserved in a separate output instead of being stripped, and expose it as an opt-in `psql_orioledb-17_debug` package (Linux-only, mirroring the `postgresql_*_debug` convention). Also add it under `legacyPackages` so it gets picked up by the CI cache-push matrix and built into the nix binary cache.

There are no orioledb debug symbols

## What is the new behavior?

orioledb debug symbols are available as an opt-in package

## Additional context

https://linear.app/supabase/issue/ORI-146/add-support-for-orioledb-debug-symbols

### DIFF
--- a/nix/checks.nix
+++ b/nix/checks.nix
@@ -951,6 +951,7 @@
             postgresql_17_debug
             postgresql_17_src
             ;
+          psql_orioledb-17_exts_orioledb_debug = self'.legacyPackages.psql_orioledb-17.exts.orioledb.debug;
         };
     };
 }
--- a/nix/ext/orioledb.nix
+++ b/nix/ext/orioledb.nix
@@ -29,6 +29,7 @@ stdenv.mkDerivation rec {
   buildPhase = ''
     make USE_PGXS=1 ORIOLEDB_PATCHSET_VERSION=20
   '';
+  separateDebugInfo = true;
   installPhase = ''
     runHook preInstall
 


nix/packages/postgres.nix
```diff
@@ -247,9 +247,19 @@
...
     in
     {
-      packages = binPackages;
-      legacyPackages = basePackages // slimPackages // cliPackages;
+      packages = binPackages // orioledbDebugPackages;
+      # Also expose the debug package under legacyPackages: the CI cache-push
+      # matrix (nix run .#github-matrix -- checks legacyPackages, see
+      # .github/workflows/nix-build.yml) only walks checks.* and
+      # legacyPackages.*, so without this it would never get built/cached.
+      legacyPackages = basePackages // slimPackages // cliPackages // orioledbDebugPackages;
```
mmlb: Adding this here will kick off separate builds for orioledb and orioledb-debug. If you want debug symbols/package in the AMI then we can install both there. I am 75% certain that just `separateDebugInfo` is enough for building/splitting/pushing/making-available-for-install.

za-arthur: @mmlb thanks! I reverted this change, I kept it in `packages`.

samrose (approved):

samrose (approved):

nix/packages/postgres.nix
```diff
@@ -247,9 +247,15 @@
...
+
+      # Opt-in package containing the split-out debug symbols for the
+      # orioledb extension's .so (see `separateDebugInfo` in nix/ext/orioledb.nix).
+      orioledbDebugPackages = lib.optionalAttrs pkgs.stdenv.isLinux {
+        psql_orioledb-17_debug = basePackages.psql_orioledb-17.exts.orioledb.debug;
+      };
     in
     {
-      packages = binPackages;
+      packages = binPackages // orioledbDebugPackages;
```
mmlb: you can drop all of this

nix/checks.nix
```diff
@@ -950,6 +950,7 @@
             postgresql_orioledb-17_src
             postgresql_17_debug
             postgresql_17_src
+            psql_orioledb-17_debug
             ;
```
mmlb: If you do this instead

```suggestion
            ;
          psql_orioledb-17_exts_orioledb_debug = self'.legacyPackages.psql_orioledb-17.exts.orioledb.debug;
```

mmlb (changes requested):
The current state is ok but imo less code is better so we don't have to trace back and forth things being defined in different files.

nix/checks.nix
mmlb: This still lets us build/run/install the packages from cli:
```
[12:09:40]-[~/p/g/s/p/src]-[manny@zennix]
nix build .#checks.x86_64-linux.psql_orioledb-17_exts_orioledb_debug
fetching git input 'git+file:///home/manny/projects/github.com/supabase/postgres/src'
Finished at 12:09:41 after 1s
[12:10:00]-[~/p/g/s/p/src]-[manny@zennix]
nix build .#psql_orioledb-17.exts.orioledb.debug
fetching git input 'git+file:///home/manny/projects/github.com/supabase/postgres/src'
Finished at 12:10:01 after 0s
[12:10:05]-[~/p/g/s/p/src]-[manny@zennix]
nix build .#postgresql_orioledb-17.debug
fetching git input 'git+file:///home/manny/projects/github.com/supabase/postgres/src'
Finished at 12:10:05 after 0s
```

nix/packages/postgres.nix
za-arthur: @mmlb do you mean revert both `orioledbDebugPackages` and `packages` changes?

mmlb: Yep those aren't necessary either. Just the `separateDebugInfo = true;` and then we can just refer to `.debug` on the normal package for both installs and tests.

za-arthur: Reverted

mmlb (approved):
nice, thanks!

PR #2380, opened 2026-08-19
Re-organize build scripts, add to shellcheck

mmlb:
## What kind of change does this PR introduce?

Maintenance

## What is the current behavior?

Image build scripts are split between ebssurroage and scripts dir, most of those in scripts/ aren't even used. The scripts aren't being shellchecked in CI.

## What is the new behavior?

Moved useful scripts into ebssurrogate and made them shellcheck clean. I added a shellcheck lint to git-hooks too. Dropped ansible.cfg since I noticed its not used and might as well do so now.

PR #2381, opened 2026-08-20
AMI: Setup swap at stage2 not stage1

mmlb:

Bug fix

For AMI builds we setup 1G swap in stage1 which takes up 10% of the disk during stage2. 

## What is the new behavior?

Swap space is allocated/setup at the end of stage2's ansible run.

## Additional context

The swap space isn't really necessary on these c6 machines with 32G of ram. We do end up running right up to the brink of available disk during builds/installs before cleaning out caches/gc and could put that 1G to better use there.

### DIFF
--- a/ansible/tasks/stage2/setup-swap.yml
+++ b/ansible/tasks/stage2/setup-swap.yml
@@ -1,0 +1,49 @@
+- name: Allocate and format swapfile
+  ansible.builtin.shell: |
+    fallocate -l 1G /swapfile
+    chmod 600 /swapfile
+    mkswap /swapfile
+  args:
+    creates: /swapfile
+
+- name: Add swapfile to fstab
+  ansible.posix.mount:
+    src: /swapfile
+    path: none
+    fstype: swap
+    opts: sw
+    dump: "0"
+    passno: "0"
+    state: present
+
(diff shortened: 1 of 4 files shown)


mmlb:
depends on #2380

ansible/tasks/stage2/setup-swap.yml
```diff
@@ -0,0 +1,49 @@
...
+  ansible.posix.mount:
+    src: /swapfile
+    path: none
+    fstype: swap
+    opts: sw
+    dump: "0"
+    passno: "0"
+    state: present
+
+- name: 'tuned - Decrease the kernel swappiness'  # noqa: name[casing]
```
hunleyd, line 19: the `tuned -` prefix was because this task was in the `setup-tuned` task. you can probably drop it :)

hunleyd: or, if you want to keep the convention, all tasks in this file should be `name: swap - <foo>`

hunleyd (approved):

ansible/tasks/stage2/setup-swap.yml
mmlb: ack, I'll ponder on what I want to do here and fixup in a follow up PR

PR #2384, opened 2026-08-20
Report final AMI disk usage to workflow job

mmlb:

Feature

We don't log/report disk usage of the final AMI.

## What is the new behavior?

We now show disk usage as a nice markdown table in the job summary.

## Additional context

Disk space is important to us but we haven't been reporting it/watching it much, this is an easy first step.


mmlb:
Example output/summary:
image

samrose (approved):

PR #2396, opened 2026-08-24
ci: skip nix cache push when credentials aren't configured

brainrake:
## What

\`push-to-cache: true\` previously hard-failed the whole job if \`nix-signing-key\`/\`role-to-assume\` were empty. Now it resolves once whether push-to-cache is actually possible (requested *and* credentials present), and downgrades to skipping the cache push (with a \`::notice::\`) instead of aborting when they're not.

## Why

Cache reads (\`extra-substituters\`) are public and unauthenticated — only the write-back needs credentials. That write is an optimization, not something that should block an otherwise-successful build. This also means any repo whose flake re-exports this one, and doesn't have (or want) its own AWS/nix-cache credentials configured, gets a fully green CI without needing to provision anything — it just never gets the cache-write optimization.

Behavior when both credentials are present: unchanged.


mmlb (approved):
lgtm

.github/actions/nix-install-ephemeral/action.yml
```diff
@@ -18,20 +18,21 @@ inputs:
...
       shell: bash
       run: |
-        abort() { echo "$@" >&2; exit 1; }
-        [[ -n ${NIX_SIGNING_KEY:+ok} ]] || abort "nix-signing-key is required when push-to-cache is true"
-        [[ -n ${ROLE_TO_ASSUME:+ok} ]] || abort "role-to-assume is required when push-to-cache is true"
-      env:
-        NIX_SIGNING_KEY: ${{ inputs.nix-signing-key }}
-        ROLE_TO_ASSUME: ${{ inputs.role-to-assume }}
+        VALUE=false
+        if [[ "${{ inputs.push-to-cache }}" == true && -n "${{ inputs.nix-signing-key }}" && -n "${{ inputs.role-to-assume }}" ]]; then
```
mmlb: fyi I went through `env` and the `:+ok` as a defense in depth in case github ships a bug and fails to mask the signing-key (previously would have thought not a chance that would happen, but these days...)

brainrake: Oops, restoring that.

PR #2397, opened 2026-08-25
chore: pgbouncer config

staaldraad:
Ensure login always happens against the primary database.


mmlb:
Hey @staaldraad this commit/PR is pretty hard to understand. Can you go into a little more detail in the commit message please, what happens now (random db is chosen?)

AndrewJackson2020:
Looking at the pgbouncer source code [0] it looks like if auth_dbname is not specified then whichever database the client is connecting to is used as the auth database. Pgbouncer will execute the auth query `SELECT * FROM pgbouncer.get_auth($1)` on that auth database. If the get_auth function was not deployed to this database then it will cause the auth to fail.

Have not tested this myself, just the read of the code. Also will say this default behavior is not described in the pgbouncer docs [1].

[0]https://github.com/pgbouncer/pgbouncer/blob/c0d4162c15706e711e8e49a031dad84ed1b1dda2/src/client.c#L43
[1] https://github.com/pgbouncer/pgbouncer/blob/c0d4162c15706e711e8e49a031dad84ed1b1dda2/doc/config.md?plain=1#L546

PR #2398, opened 2026-08-25
feat(multigres): add pg_activity binary to multigres docker image

niconosenzo:
Package pg_activity from nixpkgs as a new `pg-activity` flake output, following the existing `pg-backrest` pattern, and install it alongside pgBackRest in the Multigres nix builder stage.

feature

## Additional context

Include pg_activity on multigres base image.


mmlb:
Seems fine to me, I re-ran failed CI jobs and they are green again. There are better ways to do this but fine for now.

mmlb (approved):

PR #2408, opened 2026-08-29
[DO NOT MERGE] add pg_wait_sampling

za-arthur:

Add pg_wait_sampling into a benchmarking image and bump orioledb version

## What is the new behavior?

## Additional context


ansible/files/pg_wait_sampling_extension.sql
```diff
@@ -0,0 +1,2 @@
+CREATE SCHEMA IF NOT exists extensions;
+CREATE EXTENSION IF NOT EXISTS pg_wait_sampling with schema extensions;
```
matt-brews-sencha, line 2:
```suggestion
CREATE SCHEMA IF NOT EXISTS extensions;
CREATE EXTENSION IF NOT EXISTS pg_wait_sampling WITH SCHEMA extensions;
```

Cosmetic only.

matt-brews-sencha: Do we typically use a shared namespace for all extensions, rather than separate namespaces for each?  Probably the odds of collisions is low, but just curious.

za-arthur: AFAIK we use mostly `extensions` schema with few exceptions, which are installed either to a separate schema or to pg_catalog.

PR #2410, opened 2026-08-31
chore: set search path in migration

staaldraad:
A number of Supabase owned functions are created without a search_path. This breaks from best practice and can lead to unintended side-effects if the current_user search_path changes.

### DIFF
--- a/migrations/db/migrations/20260820000001_rescope_pg_graphql_access_trigger.sql
+++ b/migrations/db/migrations/20260820000001_rescope_pg_graphql_access_trigger.sql
@@ -1,0 +1,48 @@
+-- migrate:up
+
+create or replace function extensions.grant_pg_graphql_access()
+    returns event_trigger
+    language plpgsql
+    set search_path to ''
+as $func$
+begin
+    if not exists (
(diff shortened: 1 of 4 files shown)


migrations/db/migrations/20260820000001_rescope_pg_graphql_access_trigger.sql
```diff
@@ -0,0 +1,239 @@
...
+    drop function if exists graphql_public.graphql;
+    create or replace function graphql_public.graphql(
+        "operationName" text default null,
+        query text default null,
+        variables jsonb default null,
+        extensions jsonb default null
+    )
+        returns jsonb
+        language sql
+        set search_path to ''
```
imor: This will break pg_graphql as explained by @utkarash2991 in this PR: https://github.com/supabase/salt/pull/893. We should remove it and use fully qualified identifiers in the body instead.

imor (approved):
LGTM, `graphql_public.graphql` no longer set search_path to empty.

utkarash2991 (approved):

PR #2412, opened 2026-09-01
AMI: Pin down uid/gid values when creating users/groups

mmlb:

Bug fix

We create users/groups willy nilly without a worry about uid/gids assigned to them.

## What is the new behavior?

We no longer do that. Any user/group created in ansible needs to provide the relevant uid/gid, this is checked by a custom ansible-lint linter. We also check the uid/gid using supascan/goss at ami build time to catch any unexpected change.

## Additional context

When I dropped the nginx user/group the other uid/gids changed and caused an issue with resume from suspend due to file perms.Nothing uses it and its just a drag on the repo, for understanding it
and grepping.


sugatmahanti (approved):

pinaraf:
Note that this enforces "bad" UIDs on these users. UIDs over 1000 are for human users, not for system services. System services should have an UID below 1000 (well, on debian, 500 for RH systems as far as I remember).
I also do not see how this will work if we have any dependency in the future that comes with its own conflicting users.
Finally, since this was never made mandatory on our systems, how would this assertion help in any case? We have varying UIDs depending on the version, this must be considered a fact. The only static UIDs should be the ones "escaping" the rootfs boundary, like Postgres and Wal-g.

pcnc (approved):

mmlb:
> Note that this enforces "bad" UIDs on these users. UIDs over 1000 are for human users, not for system services. System services should have an UID below 1000 (well, on debian, 500 for RH systems as far as I remember).

Yep, not trying to fix the state of things atm just avoid more unexpected heartache.

> I also do not see how this will work if we have any dependency in the future that comes with its own conflicting users.

Not worried about that right now, I expect it will fail at build time so we can address it then.

> Finally, since this was never made mandatory on our systems, how would this assertion help in any case? We have varying UIDs depending on the version, this must be considered a fact.

We don't have any evidence of varying [UG]IDs before [bfe182b4](https://github.com/supabase/postgres/commit/bfe182b4b6afeee6907d0140d8d97582758618e1) and this PR enforces the expectation. The script I've added can be used to fetch AMIs and verify, the lack of issues until recent rollout gives me some small confidence it won't be too bad. That and the fact that both amd64 and arm64 AMIs have matching [UG]IDs.

> The only static UIDs should be the ones "escaping" the rootfs boundary, like Postgres and Wal-g.

This is a stop-gap measure until we come up with a more correct solution, systemd managed ids or encasing in OCI (where file perms can still be an issue when crossing containers).

mmlb:
I've added a script to fetch ami root vol's contents and used it to confirm the pinned uid/gids map to same values we had in previous AMIs and .155 had the wrong ones:
```
[--:--:--]-[~/p/g/s/p/src]-[manny@zennix]
losetup | grep /dev | sort
/dev/loop0         0      0         0  0 /home/manny/projects/github.com/supabase/postgres/src/.working/postgres-17.6.1.155-amd64.img   0     512
/dev/loop1         0      0         0  0 /home/manny/projects/github.com/supabase/postgres/src/.working/postgres-17.6.1.155-arm64.img   0     512
/dev/loop2         0      0         0  0 /home/manny/projects/github.com/supabase/postgres/src/.working/postgres-17.6.1.166-amd64.img   0     512
/dev/loop3         0      0         0  0 /home/manny/projects/github.com/supabase/postgres/src/.working/postgres-17.6.1.166-arm64.img   0     512
[--:--:--]-[~/p/g/s/p/src]-[manny@zennix]
mount | grep mnt | grep /dev | sort
/dev/loop0p2 on /mnt/155-amd64 type ext4 (rw,relatime)
/dev/loop1p2 on /mnt/155-arm64 type ext4 (rw,relatime)
/dev/loop2p2 on /mnt/166-amd64 type ext4 (rw,relatime)
/dev/loop3p2 on /mnt/166-arm64 type ext4 (rw,relatime)
[--:--:--]-[~/p/g/s/p/src]-[manny@zennix]
rg -A2 ansible.builtin.user: ansible |
    grep -e name: -e uid: |
    awk '{print $NF}' |
    xargs -n2 |
    awk '{print $1 ":x:" $2}' |
    sort |
    while read l
        for mnt in {155,166}-{amd,arm}64
            grep -q $l /mnt/$mnt/etc/passwd || echo $mnt $l is wrong
        end
    end
166-amd64 adminapi:x:1006 is wrong
166-arm64 adminapi:x:1006 is wrong
166-amd64 ubuntu:x:1007 is wrong
166-arm64 ubuntu:x:1007 is wrong
[--:--:--]-[~/p/g/s/p/src]-[manny@zennix]
rg -A2 ansible.builtin.group: ansible |
    grep -e name: -e gid: |
    grep -v ssl-cert |
    awk '{print $NF}' |
    xargs -n2 |
    awk '{print $1 ":x:" $2}' |
    sort |
    while read l
        for mnt in {155,166}-{amd,arm}64
            grep -q $l /mnt/$mnt/etc/group || echo $mnt $l is wrong
        end
    end
166-amd64 adminapi:x:1011 is wrong
166-arm64 adminapi:x:1011 is wrong
166-amd64 ubuntu:x:1014 is wrong
166-arm64 ubuntu:x:1014 is wrong
```

PR #2413, opened 2026-09-03
DO NOT MERGE: scratch — unique version strings to test amcheck AMI on green

brainrake:
Scratch/throwaway branch off #2240's head, only changes `ansible/vars.yml` `postgres_release` to append a `-amchecktest` suffix to each version string. Exists purely so `ami-release-nix.yml` can be triggered on a uniquely-tagged build, which is then registered on green (via \`supadev create-project ... --version <this PR URL>\`) to test amcheck surviving a real 15→17 upgrade, without colliding with #2240's own release version tags.

**Do not merge.** Delete once the AMI test is done.


brainrake:
won't work, no way to deploy to green only.

PR #2416, opened 2026-09-04
testinfra: check postgres uid/gid don't drift

brainrake:
Add testinfra check that `postgres` uid/gid (999/1002, pinned in `ansible/tasks/setup-postgres.yml`) don't drift on built AMIs.


mmlb:
This is taken care of recently by #2412

PR #2418, opened 2026-09-04
Update uid/gid pins for qemu

mmlb:

Bug fix and feature!

We aren't building the qemu images because of uid/gid conflicts, but more importantly we don't treat qemu image builds like the AMI builds and don't check the output with supascan.

## What is the new behavior?

Ansible explicitly creates users/groups with same uid/gids as we had previously and supascan verifies the qemu build artifact.

### DIFF
--- a/ansible/playbook.yml
+++ b/ansible/playbook.yml
@@ -183,10 +183,11 @@
         systemctl stop fail2ban.service
 
     - name: Run supascan baseline validation
-      when: stage2 and not qemu
+      when: stage2
       become: yes
       shell: |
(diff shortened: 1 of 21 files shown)


brainrake:
Oof, any way we can avoid diverging ids?

brainrake:
Or was that the previous state? In that case, maybe we can migrate qemu images? What are they used for downstream?

mmlb:
@brainrake yeah oof indeed. This is matching the previous state. I'd like to transition so everything matches but thats not really in the cards right now and we need to get back to the previous values ASAP.

brainrake (approved):

PR #2425, opened 2026-09-07
ci: publish site-env catalogs, extend release path filter

brainrake:
Extends `ami-release-nix.yml`'s existing catalog step (same jq/bucket/role as the live `psql_<major>` catalog) to also publish `site-env-<major>` and `site-update` catalogs. Adds `nix/**` to the release trigger path filter.

DO NOT MERGE - WIP. Depends on #2424 (site-update) existing so `.#site-update.outPath` resolves.

MPG-15


brainrake:
Merged into #2424.

PR #2428, opened 2026-09-07
ami: bake site-update into its own profile

brainrake:
Bakes `site-update` into `/nix/var/nix/profiles/site-update` at AMI build time, symlinked onto PATH. Independent of `site-env-<major>` — salt always re-fetches its own copy of the tool; this just makes a fresh AMI functional/testable before salt's first run.

DO NOT MERGE - WIP. Depends on #2424.

MPG-16


brainrake:
Merged into #2424.

PR #2430, opened 2026-09-07
EXPERIMENT: diff supautils currently deployed vs nix build

brainrake:
One-off check, always fails after printing the diff.


brainrake:
Running this experiment locally instead.

PR #2431, opened 2026-09-07
chore(pgmq): update

brainrake:
DO NOT MERGE until #2433 merges first - this branch is rebased on top of it.

Draft - breaking-change review 1.5.1 -> 1.13.0 to happen in this PR before merge.

- pgmq moved GitHub orgs: `tembo-io/pgmq` -> `pgmq/pgmq`.
- Drop the Supabase-only shim that injected `pgmq._extension_exists` into the 1.5.0->1.5.1 upgrade script - upstream defines it itself (`CREATE OR REPLACE FUNCTION`) from 1.5.1->1.5.2 onward, nothing else in this repo calls it.
- Add `1.13.0` (latest, released 2026-09-07) to `versions.json`. Keeps `1.4.4`/`1.5.1` for pinned-version installs/restores. This makes 1.13.0 the shipped `default_version`.
- Updated `pgmq`/`z_*_ext_interface` expected-output snapshots for 1.13.0's expanded function surface (reviewed in PR comments).

Depends on #2433 (`drop_queue` overload fix) - included here via rebase since testing the fix across all pinned versions needs 1.13.0 to exist. Once #2433 merges to develop, rebase this on top and the diff here will shrink to just this PR's own changes.

MPG-110


brainrake:
Snapshot update review (pgmq 1.5.1 -> 1.13.0, 40 -> 76 functions):

- No `SECURITY DEFINER` in any new function; only `GRANT` is `USAGE`/`SELECT` to `pg_monitor` (pre-existing pattern) - no new privilege escalation.
- `pop` and `create_partitioned` look like breaking signature changes in the diff but aren't - new params (`qty`, `premake`) have safe defaults, old call arity still resolves.
- New notify feature (`enable_notify_insert`) is opt-in per queue, NULL payload, no content leak.
- `pgmq.out` diff verified byte-identical across pg15/17/orioledb-17/multigres-17.
- `z_*_ext_interface.out` diffs carry identical data across platforms (only column-padding whitespace differs); the secondary row-count shift (1168->1162) traced to `a_foo`/`q_foo` test-table columns being offset by new `notify_insert_throttle`/`topic_bindings` table columns.
- All 5 snapshots regenerated from actual CI-produced output, not hand-written.

brainrake:
**`pgmq.set_vt` overload collision (new in 1.13.0)**

1.5.1 had one `set_vt(queue_name, msg_id, vt INTEGER)`. 1.13.0 adds a second with the same param names but `vt TIMESTAMP WITH TIME ZONE`. Same ambiguity shape as `drop_queue` - breaks PostgREST RPC resolution (`Could not choose the best candidate function`) and any caller passing `NULL` for `vt`.

Not fixable like `drop_queue` - we don't own this function, and collapsing it breaks someone either way. Needs to go in the breaking-change writeup, not patched.

(`send`/`send_batch` have the same pattern but unchanged since 1.5.1 - pre-existing, not new. `send_batch_topic` has it too but is brand-new, no prior reliance.)

PR #2434, opened 2026-09-08
feat(nix): add nix-dl, retry+timeout wrapper for nix-store -r

brainrake:
## Summary
- nix-store -r can stall indefinitely on a dropped connection (NixOS/nix#8758, NixOS/nix#15419 — `stalled-download-timeout` exists but is global and its default is often too long).
- Adds `nix-dl`, a shared package wrapping each path realize in `timeout`+retry, matching the pattern already in `ansible/files/admin_api_scripts/pg_upgrade_scripts/initiate.sh`.
- Configurable via `NIX_DL_TIMEOUT` / `NIX_DL_KILL_GRACE` / `NIX_DL_ATTEMPTS` (defaults 120s/10s/3).
- `site-extensions-update` now uses it instead of a bare `nix-store -r` call.

## Test plan
- [x] `nix build .#nix-dl` — builds, shellcheck passes
- [x] `nix build .#site-extensions-update` — builds

### DIFF
--- a/nix/packages/default.nix
+++ b/nix/packages/default.nix
@@ -5,6 +5,7 @@
     ./postgres-env.nix
     ./site-env.nix
     ./extension-catalog.nix
+    ./nix-dl.nix
   ];
   perSystem =
     {
--- a/nix/packages/extension-catalog.nix
+++ b/nix/packages/extension-catalog.nix
@@ -152,13 +152,14 @@
           name = "site-extensions-update";
           runtimeInputs = [
             self'.packages.site-extensions-resolve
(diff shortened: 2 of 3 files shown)


brainrake:
Dropping this: the two issues cited (NixOS/nix#8758, NixOS/nix#15419) describe long bounded stalls with eventual retry, not an indefinite hang — the premise doesn't hold up. Reverted its two call sites (site-extensions-update, site-env-update in #2424) back to a direct `nix-store --realise` call (spelled out instead of `-r` for clarity).

PR #2435, opened 2026-09-08
site-env-update: skip realize+set when profile already current

brainrake:
Split out of #2424. Skips redundant realize+set when the site profile already points at the target path.


brainrake:
Absorbed into #2424 (commit 716aa2f4) with the [[ ]] bashism fix applied per Manny's convention from #2317. Closing in favor of keeping it in the main site-env-update history.

PR #2437, opened 2026-09-08
ci: gate nix linux packages on required glibc <= 2.31

brainrake:
### Why?

Supautils is `dlopen`ed into running postgresql and resolves `glibc` from the one that is already loaded. So it needs to be compatible with old `glibc`. See https://github.com/supabase/postgres/pull/2436

### What

Adds a step to both Linux package matrix jobs that scans every built `legacyPackages.*` artifact for `objdump -T` GLIBC symbol versions and fails if any exceed 2.31 (Ubuntu 20.04 / Debian 11 floor).

## Test plan
- [ ] CI run on this PR exercises the new step across the Linux package matrix

MPG-126

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -73,6 +73,9 @@ jobs:
         uses: ./.github/actions/nix-build-retry
         with:
           attr: .#${{ matrix.attr }}
+      - name: Check glibc floor
+        if: ${{ matrix.attr != '' }}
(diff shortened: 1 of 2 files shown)


mmlb:
There's no explanation for _why_ we want to have this functionality. It also seems to me that we use libc from nix for most binaries so this enforcement doesn't make sense in those cases right?

mmlb:
Also https://github.com/fzakaria/sqlelf is worth a look too ;)

brainrake:
@mmlb updated description to show why we need this. It's a tricky one.

PR #2443, opened 2026-09-10
fix: harden search_path in amcheck after-create.sql

brainrake:
search_path='' still lets pg_temp shadow unqualified pg_depend/pg_proc/pg_extension. Use pg_catalog, pg_temp + qualify refs.

PR #2444, opened 2026-09-10
fix: harden search_path in pgmq after-create.sql

brainrake:
Same issue as amcheck (#2443): unqualified pg_depend/pg_class/pg_extension let pg_temp shadow them, tricking the script into an arbitrary alter ... owner to postgres.


brainrake:
Closing — sticking to salt-only fix for now.

PR #2450, opened 2026-09-14
ci: add package version diff comment on PRs

brainrake:
Reviving #2299 . This can be used instead of Nix Dependency Analysis workflow as a much faster alternative that won't run out of disk space so easily.

- Evaluates every derivation's `drvPath` under `legacyPackages.<system>` on the PR's head and base commits (pure evaluation — no network, no build).
- If nothing differs, reports that directly with zero further work.
- If something differs, only the changed/added/removed attrs are built (substituted from the shared cache) and diffed with `nvd` — the rest of the closure is never touched.
Results are posted as a single collapsed comment per PR, updated in place on subsequent pushes.

Example — no differences: see #2299 ' s first comment.
Example — with a real diff (added package): https://github.com/supabase/postgres/pull/2306


.github/workflows/nix-build.yml
```diff
@@ -278,3 +278,249 @@ jobs:
...
+        uses: ./head/.github/actions/nix-install-ephemeral
+        with:
+          push-to-cache: 'true'
+        env:
+          DEV_AWS_ROLE: ${{ secrets.DEV_AWS_ROLE }}
+          NIX_SIGN_SECRET_KEY: ${{ secrets.NIX_SIGN_SECRET_KEY }}
+
+      - name: Install nix (self-hosted)
+        if: matrix.installer == 'self-hosted'
+        uses: ./head/.github/actions/nix-install-self-hosted
```
mmlb, line 328: Pretty sure nix-install-ephemeral can be used for the arm darwin runners too.

.github/workflows/nix-build.yml
```diff
@@ -278,3 +278,249 @@ jobs:
...
+
+      # nix-eval's packages_matrix only lists *uncached* packages, so it
+      # cannot be used to decide what to diff. Instead, evaluate every
+      # derivation's drvPath under legacyPackages.<system> (recursing into
+      # attrsets like nix-eval-jobs --force-recurse does) on both sides.
+      # drvPath captures any input change (source, patches, deps, build
+      # steps), not just a bumped version string, and evaluating it is pure
+      # -- no substituter queries, no downloads, no builds.
```
mmlb, line 336: Worth uploading the nix-eval-job's output as a build artifact that you can then fetch here? This way don't waste time eval'ing all the things twice.

PR #2455, opened 2026-09-16
ci(ami): nix-store --optimise 

brainrake:
Adds `nix-store --optimise` to hardlink duplicate store paths and reduce AMI disk usage.


brainrake:
it was already there

PR #2456, opened 2026-09-16
nix: add ICU-pinned psql_17_icu73 build variant

brainrake:
WIP pinning ICU per project instead of always tracking the repo-wide `icu75` pin.

- `psql_17_icu73`: new package built against ICU 73.2 (`nix/packages/postgres.nix`)
- `nix/checks.nix`: verifies it actually links `libicuuc.73`
- `ami-release-nix.yml`: publishes it to its own nix-catalog entry
- `initiate.sh`: `ICU_VERSION` stub — when set, fetches the matching `psql_${PGVERSION}_icu${ICU_VERSION}` catalog instead of the default. Detection of the right value per project is not implemented yet.
- `generic.nix`'s `icu75` arg renamed to a generic `icu`; the pin now lives explicitly at the `default.nix` callsite instead of being implied by the argument's name.

No behavior change for existing builds (verified: `psql_17`, `psql_15`, `psql_orioledb-17`, `psql_17_slim`, `psql_17_cli` all resolve to the same store paths as before).


brainrake:
Approach abandoined.

PR #2457, opened 2026-09-16
ci: Configure AMI build EBS settings for max perf/faster builds

mmlb:
Our AMI build time is bottle necked on the EBS snapshot times for both stage1 and final stage2 volumes. A recent experiment indicated that we can drop the snapshot times from 9m -> 1m and similar perf for writes during build (though not as big a deal). This will cost us a few cents more in AWS but saves on engineer time which means we should be able to treat the AMI builds as PR gating jobs w/o extensive tweaking of the workflow.

feature

The AMI builds are bottlenecked on EBS snapshot times for the AMIs, which we have 2 of (stage1 and stage2/final AMI) 2x10m. 

## What is the new behavior?

Faster snapshots means faster CI builds the largest benefits are 20m -> 2m, likely smaller gains from the writes that are going on as part of the AMI build itself.


brainrake (approved):

mmlb:
1h11m -> 51m, I'll take it.

PR #2459, opened 2026-09-17
feat: Enable coredump collection and processing for OrioleDB

pgnickb:

feature: collect and process coredumps for orioledb instances.

coredumpts arent' collected

## What is the new behavior?

upon crash a coredump is written, then it is processed into a bundle including:
- full backtrace including more details specific to orioledb
- part of postgres logs from the time of the crash
- build related info
- coredump metadata

## Additional context

this work is critical for orioledb team as we're about to start beta and debugging any future crashes without coredump is going to be very problematic.

One important caveat: currently only `postgres` debug symbols are shipped. Shipping `orioledb` debug symbols is outside of the scope of this PR.


ansible/files/coredump/process-orioledb-coredumps.sh
```diff
@@ -0,0 +1,281 @@
...
+	awk '$1 == "csvlog" {p = $2} $1 == "stderr" && !p {p = $2} END {print p}' "$PGDATA_CURRENT_LOGFILES"
+}
+
+enforce_retention() {
+	find "$OUTPUT_DIR" -maxdepth 1 -type f -mtime "+${MAX_AGE_DAYS}" -delete 2>/dev/null || true
+
+	while true; do
+		total=$(du -sb "$OUTPUT_DIR" 2>/dev/null | cut -f1)
+		[ -z "$total" ] && break
+		[ "$total" -le "$MAX_TOTAL_BYTES" ] && break
```
mmlb:
```suggestion
		(( total <= MAX_TOTAL_BYTES )) && break
```

testinfra/test_ami_nix.py
```diff
@@ -1407,3 +1407,183 @@ def test_apparmor_denies_access_to_sensitive_paths(host):
...
+    core files containing customer data."""
+    result = run_ssh_command(
+        host["ssh"], "stat -c '%a %U:%G' /var/lib/systemd/coredump"
+    )
+    assert result["succeeded"], f"stat failed: {result['stderr']}"
+    mode, owner = result["stdout"].strip().split()
+    assert mode in ("700", "750"), (
+        f"Expected /var/lib/systemd/coredump to be root-only, got mode {mode}"
+    )
+    assert owner.startswith("root:"), (
```
mmlb: Why not root:root? One of the allowed modes is 750 and if this ends up with some other group that can be escaped to it can be a problem?

testinfra/test_ami_nix.py
```diff
@@ -1407,3 +1407,183 @@ def test_apparmor_denies_access_to_sensitive_paths(host):
...
+
+def test_postgres_prestart_does_not_reset_core_limit(host):
+    """Regression guard: postgres_prestart.sh must never touch 'ulimit -c',
+    or it would silently defeat the postgresql.service LimitCORE=infinity
+    coredump drop-in."""
+    result = run_ssh_command(
+        host["ssh"], "cat /usr/local/bin/postgres_prestart.sh"
+    )
+    assert result["succeeded"], f"Could not read prestart script: {result['stderr']}"
+    assert "ulimit -c" not in result["stdout"], (
```
mmlb: This worries me slightly, I'd feel slightly better with a regex for handling spaces but even then thats pretty light. Why don't we block on `ulimit` completely as an extra conservative check? We can handle dealing with w/o ulimit is being added when we get there...

testinfra/test_ami_nix.py
```diff
@@ -1407,3 +1407,183 @@ def test_apparmor_denies_access_to_sensitive_paths(host):
...
+
+    # Postgres backend: grab a real backend pid and crash it the same way.
+    backend_pid = _crash_a_backend_and_wait_for_recovery(host)
+
+    after = run_ssh_command(
+        host["ssh"], "sudo coredumpctl list --no-legend 2>/dev/null || true"
+    )
+    new_lines = [
+        line for line in after["stdout"].splitlines() if line not in before_lines
+    ]
```
mmlb:
```suggestion
    new_lines = set(after["stdout"].splitlines()) - before
```

ansible/files/coredump/orioledb-coredump.service
```diff
@@ -0,0 +1,14 @@
...
+[Service]
+Type=oneshot
+ExecStart=/usr/local/sbin/process-orioledb-coredumps.sh
+User=root
+Group=root
+NoNewPrivileges=yes
+PrivateTmp=yes
+ProtectSystem=strict
+ProtectHome=yes
+ReadWritePaths=/var/lib/systemd/coredump /var/lib/orioledb-coredumps /run
```
mmlb:
```suggestion
ReadOnlyPaths=/var/lib/systemd/coredump
ReadWritePaths=/var/lib/orioledb-coredumps /run
```

ansible/files/coredump/orioledb-coredump.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Periodic fallback sweep for PostgreSQL/OrioleDB core dumps
+
+[Timer]
+OnBootSec=10min
+OnUnitActiveSec=10min
```
mmlb: We already have issues with salt every 10m and this should be very rare situations, can we bump to 1/h or something instead? Also consider adding a `RandomizedDelaySec=15m` or something too?

ansible/tasks/setup-coredump-processing.yml
```diff
@@ -0,0 +1,75 @@
+---
+# gdb and binutils (readelf) are only used by the processor script below,
+# which is only installed for OrioleDB images due to is_psql_oriole
+- name: ensure gdb and binutils are installed
+  ansible.builtin.apt:
+    name:
+      - gdb
+      - binutils
+  become: true
```
mmlb: unnecessary, I'm going to drop them all at some point and would prefer to no introduce new ones :D, ditto down the file
```suggestion
```

testinfra/test_ami_nix.py
mmlb: previous was fine but since you already setup before as a `set()` my mind did not like that you didn't use set here too :D

ansible/files/coredump/orioledb-coredump.timer
pgnickb: Yeah it's valid. The only risk is a crash loop scenario and `coredump-ctl` imposing its own retention, but in that scenario every individual coredump isn't that valuable. Will do.

ansible/files/coredump/orioledb-coredump.service
pgnickb: That's the coredump path. We delete from that path to enforce our own retention.
I.e. processed cores are deleted, unprocessable cores are deleted after 3 attemps, etc. See `ansible/files/coredump/process-orioledb-coredumps.sh`

testinfra/test_ami_nix.py
pgnickb: But will be caught by the check below. And afaiu it isn't a problem since group can only read the filenames which isn't critical.

testinfra/test_ami_nix.py
pgnickb: Totally agree on the regex. Good call.

Not sure we want to be as tight on other things ulimit though. Ultimately it's your call of course. Do you think this is the right place to do that? I'd at least rename the test then.

mmlb: yeah I'd say lets just do `ulimit` itself, its not there now so we can cross that bridge when we get there (probably good to have to discuss it then anyway)

ansible/files/coredump/orioledb-coredump.service
mmlb: ah yeah that makes sense, seems fine then.

ansible/files/postgresql_config/postgresql.service
```diff
@@ -11,12 +11,14 @@ Type=notify
 User=postgres
 ExecStart=/usr/lib/postgresql/bin/postgres -D /etc/postgresql
 ExecStartPre=+/usr/local/bin/postgres_prestart.sh
+{% if is_psql_oriole is defined and is_psql_oriole %}
```
mmlb: is_psql_oriole is always defined

ansible/tasks/setup-postgres.yml
```diff
@@ -292,89 +292,92 @@
...
-    state: 'directory'
-  become: true
+    # Required for this (plus CoredumpFilter= in postgresql.service and the
+    # storage limits below) to do anything at all: without systemd-coredump,
+    # core_pattern stays at the kernel's plain "core" default and none of this
+    # gets exercised (confirmed on a local test VM).
+    - name: ensure systemd-coredump is installed
+      ansible.builtin.apt:
+        name: systemd-coredump
+      become: true
```
mmlb: these snuck in again

ansible/tasks/setup-postgres.yml
```diff
@@ -1,3 +1,10 @@
+# is_psql_oriole is also (re)computed in stage2-setup-postgres.yml, but this
+# file's coredump block below runs in both stage1 and stage2, so the fact
+# needs to exist before stage2-only tasks have had a chance to set it.
+- name: Determine postgres flavor from psql_version
+  ansible.builtin.set_fact:
+    is_psql_oriole: "{{ psql_version is defined and psql_version == 'psql_orioledb-17' }}"
+
```
mmlb: move to early in playbook.yml (along with the similar bools)

PR #2448, opened 2026-09-14
unblock ci: work around crates.io User-Agent blocklist

brainrake:
## Summary
- crates.io started 403'ing `python-requests`' default UA and rate-limiting its API endpoint ([NixOS/nixpkgs#512735](https://github.com/NixOS/nixpkgs/pull/512735)), breaking every rust-based build here — `fetchCargoVendor`, `importCargoLock`, and `fetchCrate` alike.
- `nix/cargo-pgrx/fix-cargo.nix` redirects each to `static.crates.io` / swaps in the fixed util script, scoped to cargo-pgrx's `rustPlatform` construction.
- Verified with `nix build .#cargo-pgrx_0_14_3` and against Raminder's pg_graphql 1.6.2 bump (`rs/bump-pg-graphql`) — builds clean end-to-end.

## Test plan
- [x] `nix build .#cargo-pgrx_0_14_3` succeeds
- [x] Raminder's pg_graphql 1.6.2 bump builds clean with this fix applied
- [ ] CI green on darwin-arm64 build job

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2463, opened 2026-09-18
unblock ci: update cargo-pgrx 0.11.3 vendor hash

brainrake:
Hash mismatch fix, may be related to supabase/postgres#2448. Also checked all other pgrx hashes, only this one changed.


mmlb:
I was pretty curious about this only not being caught by CI and it looks like we can drop all these explicit pgrx packages. I'm going to push a PR doing so that should replace this one.

mmlb (changes requested):
this is actually dead code, better off being deleted see https://github.com/supabase/postgres/pull/2475

brainrake:
superseded by #2475

PR #2471, opened 2026-09-21
checks: drop legacyPkgName, resolve extensions from pgpkg.version

brainrake:
- checks: simplify
- checks: fix supautils loading to match prod (`session_preload_libraries` instead of `shared_preload_libraries`)


nix/checks.nix
```diff
@@ -399,8 +362,12 @@
                 substitute ${./tests/postgresql.conf.in} "$PGTAP_CLUSTER"/postgresql.conf \
                   --subst-var-by PGSODIUM_GETKEY_SCRIPT "${getkey-script}/bin/pgsodium-getkey" \
                   --subst-var-by PRELOAD_LIBRARIES "$PRELOAD_LIBRARIES"
-                echo "listen_addresses = '127.0.0.1'" >> "$PGTAP_CLUSTER"/postgresql.conf
-                echo "port = ${pgPort}" >> "$PGTAP_CLUSTER"/postgresql.conf
+                {
+                  echo "listen_addresses = '127.0.0.1'"
+                  echo "port = ${pgPort}"
+                  echo "session_preload_libraries = 'supautils'"
+                  echo "dynamic_library_path = '${supautils}/lib:\$libdir'"
```
mmlb, line 369: Everything looks great except for this bit which seems like it snuck in? There's no mention of this in the commit message.

mmlb (changes requested):
commit message isn't super great would have preferred just the why and not so much the how.

nix/checks.nix
brainrake: First line is to match production more closely. Prod uses `session_preload_libraries` for supautils, while this test used `shared_preload_libraries`.
Second line is needed because supautils is also removed from `WANTED_EXTS` (to not end up in `shared_preload_libraries`) so it is not in `$libdir`.

brainrake: It was mentioned in the PR description. Sorry about the commit slop, forgot to review that. Fixed.

mmlb: ah, my eyes definitely crossed when looking at the old description :D

mmlb (approved):
lgtm nice simplification and making things look like prod is the way to be!

PR #2472, opened 2026-09-21
fix: disable Kexec Handover to free up ~128MiB of RAM

lowjoel:

fix: disable Kexec Handover to free up ~128MiB of RAM

Required for KUBE-1370.

On Linux 7.0 kernels kho is enabled by Ubuntu:

```
ubuntu@db-jrskfeoczfnalrxisytn:~$ sudo dmesg | grep -i 'cma-'
[  +0.000101] Memory: 545236K/819200K available (27136K kernel code, 6266K rwdata, 20296K rodata, 15232K init, 1301K bss, 233864K reserved, 32768K cma-reserved)
```

This means that we're reserving even more memory than on 6.8.0:

```
ubuntu@db-nfkaciskftvcorhjdwao:~$  cat /proc/cmdline
BOOT_IMAGE=/vmlinuz-6.8.0-106-generic root=PARTUUID=6555485d-34e0-4488-9101-ad42900d9ace ro console=tty1 console=ttyAMA0
ubuntu@db-nfkaciskftvcorhjdwao:~$ sudo dmesg | grep -i 'cma-'
[    0.000000] Memory: 701468K/819200K available (23488K kernel code, 5208K rwdata, 16440K rodata, 12288K init, 1233K bss, 84964K reserved, 32768K cma-reserved)
```

## What is the new behavior?

```
ubuntu@db-jrskfeoczfnalrxisytn:~$  cat /proc/cmdline
BOOT_IMAGE=/vmlinuz-7.0.0-30-generic root=PARTUUID=f257fc73-df70-4995-8eb6-953e15761105 ro console=tty1 console=ttyAMA0 kho=off panic=-1
ubuntu@db-jrskfeoczfnalrxisytn:~$  sudo dmesg | grep 'cma-'
[    0.003953] Memory: 682576K/819200K available (27136K kernel code, 6266K rwdata, 20296K rodata, 15232K init, 1301K bss, 96588K reserved, 32768K cma-reserved)
```

## Additional context

kho reserves memory to be handed over to the kexec'ed kernel; on memory
constrained instances this is a significant amount of memory. Disable
kho since we don't use kexec to reboot anyway.


ansible/tasks/setup-system.yml
```diff
@@ -109,6 +109,18 @@
...
+      ansible.builtin.copy:
+        content:
+          GRUB_CMDLINE_LINUX_DEFAULT="${GRUB_CMDLINE_LINUX_DEFAULT} kho=off"
+        dest: /etc/default/grub.d/51-disable-kho.cfg
+        owner: root
+        group: root
+      notify:
+        - Regenerate grub configuration
+      when:
+        - qemu
```
mmlb:
```suggestion
      when: qemu
```
but actually why not for all builds?

lowjoel: I _don't_ think we use kexec in v2; but I wanted to keep this for v3 only. If we are ok with this in v2 I can remove the restriction. Let me test that it does as it says on v3 first.

lowjoel:
OK I've tested this with `17.6.1.175-lowjoel`, kho=off is set in cmdline and reserved memory is back to ~96MiB. I've removed the qemu restriction; @mmlb could you have a review please?

mmlb (approved):
Nice find!

PR #2473, opened 2026-09-21
[DO NOT MERGE] Test run for OrioleDB AMI release process

pgnickb:
Please ignore


mmlb:
You might want to rebase on top of #2155

za-arthur:
> You might want to rebase on top of https://github.com/supabase/postgres/pull/2155

@mmlb can you give more context?

mmlb:
#2155 upgrades to 17.11 and as such already handles wal2json an similar stuff. That PR is putting in new code to skip tests/configs for orioledb because orioledb in tree is <17.11. Some of those things you are doing in this PR so imo it would make sense to rebase so you target that PR and can integrate into that setup of configs/tests/... instead of doing similar here.

#2155 is almost ready to be merged so make sense for you to rebase on top of it now vs later.

za-arthur:
> https://github.com/supabase/postgres/pull/2155 upgrades to 17.11 and as such already handles wal2json an similar stuff. That PR is putting in new code to skip tests/configs for orioledb because orioledb in tree is <17.11. Some of those things you are doing in this PR so imo it would make sense to rebase so you target that PR and can integrate into that setup of configs/tests/... instead of doing similar here.

But that PR upgrades only tests, it doesn't upgrade postgres to 17.11.

mmlb:
ah sorry about that I'm pushing the commits individually so CI runs and I see the expected failures for the regression tests before the upgrade "magically" fixes them :D. I'll push the last 2 commits that end up with upgraded PG and then avoid doing this again.

mmlb:
Note the last couple of comments I made in the PR

za-arthur:
> Note the last couple of comments I made in the PR

Hm, I think we didn't have failures of hstore tests in this PR.

za-arthur:
I only had to add `wal2json` into `output_plugin_libraries` in few places

PR #2475, opened 2026-09-21
Drop explicit pgrx outputs/packages

mmlb:

Maintenance

We define pgrx versions statically in cargo-pgrx and "export" as outputs/packages/devshell.

## What is the new behavior?

These statically defined/explicit pgrx packages are dropped.

## Additional context

mkPgrxExtension calls mkCargoPgrx using hashes from versions.json in the directory.


mmlb:
PTAL @brainrake 
closes #2463

brainrake (approved):

PR #2476, opened 2026-09-21
ci/test: Clean up unnecessary/dead steps

mmlb:

CI Maintenance

We have pretty bad names for jobs/steps since they share long redundant prefixes thus causing GHA UI to truncate the more useful information. For example `nix-build-packages-aarch...`. There's no real useful information from `nix-build-` prefix.

Also the `test` workflow is a bit of a misnomer, its only testing the post migration dump schema matches what is in git. So I renamed it to a better one.

Ditto the `docker-image-test` workflow step, its more building container images.

## What is the new behavior?

Better names:
`nix-build-packages-aarch64-linux` becomes
`builds-aarch64-linux`

`nix-build-checks-aarch64-linux` becomes
`checks-aarch64-linux`

So any truncation will be minimal.


.github/workflows/nix-build.yml
```diff
@@ -223,55 +211,51 @@ jobs:
...
 
-  nix-build-checks:
+  checks:
     name: Nix Checks ✅
     runs-on: blacksmith-2vcpu-ubuntu-2404
     needs:
-      - nix-build-checks-aarch64-darwin
-      - nix-build-checks-aarch64-linux
-      - nix-build-checks-x86_64-linux
+      - checks-aarch64-darwin
```
tgallacher, line 218: Is Darwin run; Thought it was excluded in the previous PR? 🤔

tgallacher: darwin is also excluded from the `amis` list below

tgallacher (approved):

.github/workflows/nix-build.yml
mmlb: checks-aarch64-darwin is for the nix checks, which we do run for all 3 flavors. This is stuff like linters, formatters, nixosVMTests etc. If you are referring to https://github.com/supabase/postgres/pull/2477, whats being skipped there is nixos-tests for aarch64-darwin which are the qemu based nixos vm tests. No point doing those I guess.

Darwin being excluded from amis is correct too, we don't build darwin AMIs. The needs was dropped because there's no reason to depend on darwin binaries if they aren't going to be used in the AMIs.

mmlb:
Will merge once develop is green

PR #2477, opened 2026-09-21
ci: Run NixOS tests on x86_64-linux too

mmlb:

CI feature

NixOS VM tests are only run for aarch64-linux systems.

## What is the new behavior?

x86_64-linux also runs NixOS VM tests since its a load bearing production system.


mmlb:
I was surprised that CI wasn't trying to build/run the x86 vm tests but turns out its because they already ran from a separate PR's run, see https://github.com/supabase/postgres/actions/runs/35249521381

tgallacher (approved):

PR #858, opened 2024-01-17
[GEN-7585] chore: patch pg_cron perms post-pg_upgrade

pcnc:
* Blocked by user-defined triggers/fkeys
  * Context: https://supabase.slack.com/archives/C01D6TWFFFW/p1705482136318409?thread_ts=1702959923.630909&cid=C01D6TWFFFW

PR #329, opened 2022-10-18
chore: Add migrations script to postgres AMI

Lakshmipathi:
Signed-off-by: Lakshmipathi <lakshmi@supabase.io>

PR #2478, opened 2026-09-21
fix: rescope pg_graphql and pg_cron event triggers during pg_upgrade

utkarash2991:
Old projects still have `issue_pg_graphql_access` scoped to `CREATE FUNCTION` (created before 17.6.1.111 / 15.14.1.111) and `issue_pg_cron_access` scoped to `CREATE SCHEMA` (created before 15.1.0.130), while the grant functions they invoke were since updated fleet-wide and only act on `CREATE EXTENSION` events. With that mismatch, recreating the extensions during an upgrade never re-applies their wiring: `graphql_public.graphql` is left as the "not enabled" placeholder and the postgres role's cron grants are skipped.

How the mismatch crept in:
- #329 seeded `issue_pg_cron_access` on `CREATE SCHEMA` (2022); the pg_graphql trigger was likewise seeded on `CREATE FUNCTION`
- #780 (pg_cron, first shipped in 15.1.0.130) and the April pg_graphql migration `20260421000001` (first shipped in 17.6.1.111 / 15.14.1.111) rescoped trigger + function together — but migrations only run at image build, so existing projects kept the old triggers
- supabase/salt#807 and supabase/salt#847 then replaced only the grant functions fleet-wide with the `CREATE EXTENSION`-scoped bodies, leaving old projects with a trigger that fires on events the new functions ignore
- #858 added the pg_cron drop/recreate to `complete.sh`, which (like the pg_graphql disable/re-enable that was always part of pg_upgrade) now silently loses the wiring on those projects

This adds a `rescope_extension_event_triggers` helper to `common.sh` and calls it from two places:

- `initiate.sh`, before the extensions are dropped — the rescoped triggers are carried into the new cluster by pg_upgrade, and the failure-path re-enable on the source instance is covered as well
- `complete.sh`, before `run_generated_sql` recreates the extensions — safety net in case the source ran an older script bundle

Linear: [PSQLEXT-1672](https://linear.app/supabase/issue/PSQLEXT-1672/prevent-pg-graphql-from-disabling-after-project-upgrades)


ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
```diff
@@ -332,6 +332,17 @@ function complete_pg_upgrade {
...
+	# execute_patches (which recreates pg_cron): the triggers only re-apply the
+	# extension wiring if they are already scoped to CREATE EXTENSION when the
+	# recreate happens. Normally initiate.sh has already rescoped on the source
+	# cluster; this is a safety net in case it ran from an older script bundle.
+	log "3.2. Rescoping extension event triggers"
+	rescope_extension_event_triggers || {
+		log "WARNING: failed to rescope extension event triggers"
+		warnings=1
+	}
+
```
mmlb: This shouldn't be necessary since we never run from an older script bundle, I would prefer to avoid unnecessary cruft.

mmlb (changes requested):

ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
```diff
@@ -69,6 +69,58 @@ function run_sql {
...
+    ) then
+        execute 'drop event trigger if exists issue_pg_cron_access';
+        execute $q$
+            create event trigger issue_pg_cron_access
+                on ddl_command_end
+                when tag in ('CREATE EXTENSION')
+                execute procedure extensions.grant_pg_cron_access()
+        $q$;
+    end if;
+end $rescope$;
```
imor: @utkarash2991 is pg_cron also affected with the same issue?

utkarash2991: @imor I did find the same discrepancy in pg_cron too , so fixing this one in same PR.

ansible/files/admin_api_scripts/pg_upgrade_scripts/complete.sh
utkarash2991: Cool Taken care in latest update

ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
```diff
@@ -69,6 +69,61 @@ function run_sql {
...
+# migrations 20260421000001_rescope_pg_graphql_access_trigger.sql and
+# 20231020085357_revoke_writes_on_cron_job_from_postgres.sql; no-op on
+# projects that already have the CREATE EXTENSION scope.
+# Drop this once no supported upgrade source predates 17.6.1.111 /
+# 15.14.1.111 — from then on every project already has the CREATE
+# EXTENSION triggers and this is dead code.
+function rescope_extension_event_triggers {
+	local sql
+	sql=$(
+		cat <<'EOF'
```
mmlb:
```suggestion
		cat <<-'EOF'
```
and `shfmt` for make great good

ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
```diff
@@ -69,6 +69,61 @@ function run_sql {
...
+# act on CREATE EXTENSION events. With that mismatch, recreating the extension
+# never re-applies its wiring (graphql_public.graphql wrapper, cron grants).
+# Rescope the triggers so the CREATE EXTENSION statements that re-enable the
+# extensions after pg_upgrade fire them. Trigger definitions kept identical to
+# migrations 20260421000001_rescope_pg_graphql_access_trigger.sql and
+# 20231020085357_revoke_writes_on_cron_job_from_postgres.sql; no-op on
+# projects that already have the CREATE EXTENSION scope.
+# Drop this once no supported upgrade source predates 17.6.1.111 /
+# 15.14.1.111 — from then on every project already has the CREATE
+# EXTENSION triggers and this is dead code.
```
mmlb: Moved the drop message to the top of the comment so it doesn't get lost and made it a TODO since some editors highlight those (and are grepable). Some small rewordings otherwise.

```suggestion
# TODO: Drop once all projects have been upgraded past 17.6.1.111, 15.14.1.111
#
# Old projects still have grant-access event triggers scoped to the tags
# their original seed used while the grant functions themselves (updated
# fleet-wide) only act on CREATE EXTENSION events.
#
# * issue_pg_graphql_access on CREATE FUNCTION before build 17.6.1.111 /
#   15.14.1.111
# * issue_pg_cron_access on CREATE SCHEMA before 15.1.0.130
#
# With this mismatch, recreating the extension never re-applies its wiring
# (graphql_public.graphql wrapper, cron grants). Rescope the triggers so
# the CREATE EXTENSION statements that re-enable the extensions after
# pg_upgrade fire them.
#
# Trigger definitions kept identical to migrations
# 20260421000001_rescope_pg_graphql_access_trigger.sql and
# 20231020085357_revoke_writes_on_cron_job_from_postgres.sql. This is a
# no-op on projects that already have the CREATE EXTENSION scope.
```

mmlb (changes requested):

ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
utkarash2991: Done

ansible/files/admin_api_scripts/pg_upgrade_scripts/common.sh
utkarash2991: Done

mmlb (approved):
Thanks! LGTM.

